PR #18, opened 2026-03-08
chore(aur): add AUR package with automated updates

### LINKED ISSUE #13: [AUR] Offer to maintain Arch Linux AUR package
Hi! 👋

I&apos;d like to help make gentle-ai easily available to Arch Linux users via the AUR (Arch User Repository).

I can create and maintain an AUR package (e.g., `gentle-ai`) that downloads the pre-built binaries from your releases.

## What I&apos;m offering:
- Create PKGBUILD using the existing release binaries (no need to build from source)
- Support for both x86_64 and aarch64 architectures
- Keep the package updated with each release
- Act as the AUR maintainer

## Package details I&apos;m planning:
- **Package name**: gentle-ai
- **URL**: https://github.com/Gentleman-Programming/gentle-ai
- **License**: MIT
- **Architecture**: x86_64, aarch64
- **Method**: Binary package (downloads from GitHub Releases)

## Quick note on the README
I noticed the "Supported Platforms" section claims "Linux (Arch) | pacman | Supported", but there&apos;s no AUR package for gentle-ai. The pacman mention seems to refer to the dependency detector suggesting `pacman -S` for system deps (git, curl, node), not gentle-ai itself being available via pacman.

Creating an AUR package would make this claim actually true! 🎯

Let me know if you&apos;d prefer to maintain it yourself or if you&apos;d like me to proceed!

Thanks for this amazing tool! 🚀

gabrielfagundeznievas:
## Summary

Adds a GitHub Actions workflow that automatically publishes and updates the AUR package on every release.

### What it does

- Generates all AUR files dynamically (PKGBUILD, .SRCINFO, LICENSE)
- Downloads release checksums and extracts SHA256 hashes
- Pushes updates to the AUR repository automatically
- Supports both amd64 and arm64 architectures
- Manual trigger via workflow_dispatch for hotfixes

### The Package

The package is already live and installable:

```bash
# Using yay
yay -S gentle-ai

# Using paru
paru -S gentle-ai
```

https://aur.archlinux.org/packages/gentle-ai

---

## Current Situation

> I'm currently maintaining this manually — monitoring releases and triggering the workflow from my fork to keep the AUR package up to date. 
> 
> Once merged, this will be fully automated on every release.

### Why AUR matters

Arch Linux users typically prefer native AUR packages over Homebrew for:
- Better system integration with pacman
- Native dependency resolution
- No need to install a secondary package manager

---

## Setup Required After Merge

The workflow needs one secret added to the repository:

| Secret | Description |
|--------|-------------|
| AUR_SSH_KEY | SSH private key with push access to AUR (base64-encoded) |

I can help set this up! Contact me or add me temporarily as a maintainer and I'll configure it.

---

Closes #13


Alan-TheGentleman (changes requested):
Hey Gabriel! Quick follow-up from my side: sorry about the previous review being in Spanish.

To restate the review in English, the two blocking issues I want fixed before merge are:

1. In the `release` path, the version should come from `${{ github.event.release.tag_name }}`, not from `/releases/latest`. The event already gives you the exact tag that triggered the workflow, and using `latest` adds an unnecessary API dependency plus a possible mismatch if another release appears while this workflow is running.

2. The `checksums.txt` download currently uses `curl -sL` without `--fail`, and there is no explicit validation that `AMD64_HASH` and `ARM64_HASH` are non-empty before generating the files. If the asset is missing or the URL returns 404, the workflow can continue with empty hashes and publish a broken `PKGBUILD`.

Suggested guard:

```bash
[ -n "$AMD64_HASH" ] || { echo "Error: AMD64 hash is empty"; exit 1; }
[ -n "$ARM64_HASH" ] || { echo "Error: ARM64 hash is empty"; exit 1; }
```

Once those two issues are addressed, I think this PR will be in very good shape to merge. Thanks again for the contribution — this is a valuable addition to the project.

Alan-TheGentleman (changes requested):
Issue #13 is approved — this PR is close to merge-ready. Two blockers remain from the previous review:

1. **Version source**: Use `${{ github.event.release.tag_name }}` instead of calling `/releases/latest`. The release event already provides the exact tag — using `latest` adds an unnecessary API call and risks a version mismatch if releases overlap.

2. **Hash validation**: The `checksums.txt` download uses `curl -sL` without `--fail`. If the asset is missing or returns 404, the workflow continues with empty hashes and publishes a broken PKGBUILD. Add `--fail` and guards:
   ```bash
   [ -n "$AMD64_HASH" ] || { echo "Error: AMD64 hash is empty"; exit 1; }
   [ -n "$ARM64_HASH" ] || { echo "Error: ARM64 hash is empty"; exit 1; }
   ```

Both fixes are small. Once addressed, this is ready to merge. The AUR package is already live — great work maintaining it manually, let's get it automated.

Alan-TheGentleman (changes requested):
Good initiative for AUR distribution. The workflow structure is right but there are critical bugs in the PKGBUILD generation:

**1. PKGBUILD corrupted by heredoc indentation (CRITICAL).** The `cat > /tmp/PKGBUILD << PKGEOF` inside YAML indentation includes all the leading spaces in the output. Every line of the PKGBUILD will have 10+ spaces at the start, breaking `makepkg`. Fix: either use `<<-PKGEOF` with tabs, generate the file without indentation, or move the generation to an external script.

**2. Trigger condition too broad (CRITICAL).** The `if` condition includes `github.ref == 'refs/heads/main'` which fires on EVERY push to main, not just releases. This would update AUR on every commit with the previous version string. Remove that condition, keep only `release` and `workflow_dispatch`.

**3. LICENSE heredoc has the same indentation problem.** The `LICENSE_HASH` calculated from the corrupted license text won't match any legitimate MIT License hash.

**4. No `needs:` dependency on the release job.** If triggered on `release: published`, goreleaser may still be running when this workflow starts downloading assets. Add a wait/retry mechanism or `needs: [goreleaser]`.

**5. Missing newline at end of file.**

Fix items 1-4 and this is ready.

PR #219, opened 2026-04-03
fix(security): validate untrusted paths in backup, restore, and upgrade

### LINKED ISSUE #220: fix(security): 5 path traversal and injection vulnerabilities in backup, restore, and upgrade
## Description

Found **5 confirmed security vulnerabilities** via parallel adversarial code review (two independent blind AI reviewers, 3 fix+re-judge iterations):

### 1. Arbitrary file write via unvalidated `OriginalPath` in backup restore
- **File**: `internal/backup/restore.go`
- **Vector**: A tampered `manifest.json` can set `OriginalPath` to any absolute path (e.g., `/etc/cron.d/backdoor`). The restore function writes attacker-controlled content to that path without validation.
- **Impact**: CRITICAL — arbitrary file write as the running user

### 2. Arbitrary directory deletion via unvalidated `RootDir` in `DeleteBackup`
- **File**: `internal/backup/manifest.go`
- **Vector**: `os.RemoveAll(manifest.RootDir)` where `RootDir` comes from deserialized JSON. A crafted manifest with `"root_dir": "/"` deletes the entire filesystem.
- **Impact**: CRITICAL — arbitrary recursive deletion

### 3. Arbitrary file read via unvalidated `SnapshotPath` in plain backup restore
- **File**: `internal/backup/restore.go`
- **Vector**: `os.ReadFile(entry.SnapshotPath)` where `SnapshotPath` comes from manifest JSON. Setting it to `/etc/shadow` reads privileged content and writes it to a home-directory file.
- **Impact**: CRITICAL — arbitrary file read

### 4. PowerShell injection via unescaped path in `AddToUserPath`
- **File**: `internal/system/path.go`
- **Vector**: `dir` is interpolated directly into a PowerShell script. Paths containing single quotes (e.g., `C:\Users\O'Brien`) break the string literal and allow arbitrary command injection.
- **Impact**: CRITICAL — arbitrary code execution on Windows

### 5. Tar extraction accepts symlinks and special files
- **Files**: `internal/components/engram/download.go`, `internal/update/upgrade/download.go`
- **Vector**: `extractBinaryFromTarGz` matches entries by filename but doesn't check `hdr.Typeflag`. A symlink entry named `engram` pointing to a sensitive file would be accepted. The safe implementation in `backup/compression.go` correctly checks this.
- **Impact**: CRITICAL — arbitrary file overwrite via crafted archive

### Additional fix
- Added `io.LimitReader` (1MB cap) to script downloads in `scriptUpgrade` to prevent unbounded memory allocation.

## How these were found

This issue was generated by **Claude Code configured by gentle-ai itself** — the tool's own ecosystem was used to audit the tool. Two independent blind judge agents reviewed the full codebase in parallel, findings were cross-referenced, and only issues confirmed by both judges were fixed through 3 rounds of fix+re-judge iterations.

## Reproduction

All vulnerabilities require a tampered `manifest.json` in `~/.gentle-ai/backups/` or a compromised upstream server. They are exploitable when:
- A user restores from a backup directory that has been tampered with
- A user runs `gentle-ai upgrade` against a compromised CDN/server
- A user's system has paths with special characters (PowerShell injection)

Basparin:
Closes #220

## Summary

Fixes **5 confirmed security vulnerabilities** found via parallel adversarial code review:

1. **Arbitrary file write** via unvalidated `OriginalPath` in backup restore — tampered manifest could write to any path on disk
2. **Arbitrary directory deletion** via unvalidated `RootDir` in `DeleteBackup` — crafted manifest with `root_dir: "/"` triggers `os.RemoveAll`
3. **Arbitrary file read** via unvalidated `SnapshotPath` in plain backup restore — tampered manifest reads any file and writes it under home
4. **PowerShell injection** via unescaped single quotes in `AddToUserPath` — paths like `C:\Users\O'Brien` could inject arbitrary commands
5. **Tar symlink extraction** — crafted archive with symlink entry matching binary name bypasses extraction to arbitrary target

Additionally adds `io.LimitReader` (1MB cap) to script downloads to prevent unbounded memory allocation.

## How these were found

This PR was generated by **Claude Code configured by gentle-ai itself** — testing whether the ecosystem can self-improve through distributed adversarial review.

The review process:
- Two independent blind judge agents reviewed the full codebase in parallel (neither knew about the other)
- Findings were cross-referenced: only issues confirmed by **both judges** were fixed
- 3 rounds of fix → re-judge iterations until both judges passed clean
- All fixes compile and follow existing code patterns

## Changes

| File | What changed |
|------|-------------|
| `internal/backup/restore.go` | `isPathUnderHome()` with symlink resolution; SnapshotPath validation via `isRootDirUnderBackupRoot` for plain restores; `trustedSnapshot` flag for compressed restores |
| `internal/backup/manifest.go` | `isRootDirUnderBackupRoot()` with symlink resolution; validation before `os.RemoveAll` in `DeleteBackup` |
| `internal/system/path.go` | `escapePowerShellString()` escapes `'` → `''` before PowerShell interpolation |
| `internal/update/upgrade/strategy.go` | `io.LimitReader(maxScriptSize+1)` on script downloads |
| `internal/components/engram/download.go` | `hdr.Typeflag == tar.TypeReg \|\| tar.TypeRegA` check |
| `internal/update/upgrade/download.go` | Same typeflag tightening |

## Known residual warnings (out of scope)

- PowerShell `-notlike` wildcard metacharacters (`*`, `?`, `[`) not escaped in PATH duplicate check
- `ggaTmpDir` uses predictable `/tmp` path (TOCTOU race)
- No size limit on binary downloads (`io.ReadAll` / `io.Copy`)
- `restoreCompressed` doesn't validate `manifest.RootDir` before opening archive

## Test plan

- [ ] `go build ./...` passes
- [ ] Existing unit tests pass (`go test ./...`)
- [ ] Manual: `gentle-ai install --dry-run` works as before
- [ ] Manual: backup + restore cycle works for both plain and compressed formats
- [ ] Manual: upgrade path works on supported platforms

🤖 Generated with [Claude Code](https://claude.com/claude-code) — configured by gentle-ai

### DIFF
--- a/internal/backup/manifest.go
+++ b/internal/backup/manifest.go
@@ -138,6 +138,11 @@ func backupRoot() (string, error) {
 	return filepath.Join(home, ".gentle-ai", "backups"), nil
 }
 
+// BackupRootFn is the function used to resolve the backup root directory.
+// Package-level var for testability — swapped in tests to use a temp directory.
+// Exported so tests in other packages (e.g. internal/update/upgrade) can override it.
+var BackupRootFn = backupRoot
+
 // isRootDirUnderBackupRoot validates that dir is a direct or indirect subdirectory
 // of the expected backup root (~/.gentle-ai/backups/). This prevents a tampered
 // manifest with root_dir set to "/" or another sensitive path from deleting arbitrary files.
@@ -147,7 +152,7 @@ func backupRoot() (string, error) {
 // If the path does not exist yet, only filepath.Clean is used — this limitation is accepted
 // and documented here, consistent with isPathUnderHome.
 func isRootDirUnderBackupRoot(dir string) (bool, error) {
-	root, err := backupRoot()
+	root, err := BackupRootFn()
 	if err != nil {
 		return false, err
 	}
--- a/internal/backup/manifest_test.go
+++ b/internal/backup/manifest_test.go
@@ -419,6 +419,11 @@ func TestSnapshotterSkipsDirectories(t *testing.T) {
 // TestDeleteBackup_Success verifies that DeleteBackup removes the backup directory.
 func TestDeleteBackup_Success(t *testing.T) {
 	dir := t.TempDir()
+	// Override backupRootFn so validation accepts paths under t.TempDir().
+	origBackupRootFn := BackupRootFn
+	t.Cleanup(func() { BackupRootFn = origBackupRootFn })
+	BackupRootFn = func() (string, error) { return dir, nil }
+
 	backupDir := filepath.Join(dir, "backup-01")
 	if err := os.MkdirAll(backupDir, 0o755); err != nil {
 		t.Fatalf("MkdirAll: %v", err)
--- a/internal/backup/restore.go
+++ b/internal/backup/restore.go
@@ -9,6 +9,10 @@ import (
 	"github.com/gentleman-programming/gentle-ai/internal/components/filemerge"
 )
 
+// userHomeDirFn is the function used to resolve the user's home directory.
+// Package-level var for testability — swapped in tests to use a temp directory.
+var userHomeDirFn = os.UserHomeDir
+
 // isPathUnderHome reports whether path is an absolute path that resides under
 // the current user's home directory. This is used to prevent arbitrary file
 // writes via tampered manifest OriginalPath fields.
@@ -19,7 +23,7 @@ import (
 // is used — symlinks cannot be resolved for non-existent paths, so this
 // limitation is accepted and documented here.
 func isPathUnderHome(path string) bool {
-	home, err := os.UserHomeDir()
+	home, err := userHomeDirFn()
 	if err != nil {
 		return false
 	}
--- a/internal/backup/restore_test.go
+++ b/internal/backup/restore_test.go
@@ -9,6 +9,15 @@ import (
 
 func TestRestoreRestoresExistingAndRemovesCreated(t *testing.T) {
 	home := t.TempDir()
+	// Override fns so validation accepts paths under t.TempDir().
+	origUserHomeDirFn := userHomeDirFn
+	origBackupRootFn := BackupRootFn
+	t.Cleanup(func() {
+		userHomeDirFn = origUserHomeDirFn
+		BackupRootFn = origBackupRootFn
+	})
+	userHomeDirFn = func() (string, error) { return home, nil }
+	BackupRootFn = func() (string, error) { return home, nil }
 
 	originalPath := filepath.Join(home, "config", "settings.json")
 	if err := os.MkdirAll(filepath.Dir(originalPath), 0o755); err != nil {
@@ -57,10 +66,21 @@ func TestRestoreRestoresExistingAndRemovesCreated(t *testing.T) {
 }
 
 func TestRestoreFailsWhenSnapshotMissing(t *testing.T) {
+	tmpDir := t.TempDir()
+	// Override fns so validation accepts paths under t.TempDir().
+	origUserHomeDirFn := userHomeDirFn
+	origBackupRootFn := BackupRootFn
+	t.Cleanup(func() {
+		userHomeDirFn = origUserHomeDirFn
+		BackupRootFn = origBackupRootFn
+	})
+	userHomeDirFn = func() (string, error) { return tmpDir, nil }
+	BackupRootFn = func() (string, error) { return tmpDir, nil }
+
 	service := RestoreService{}
 	err := service.Restore(Manifest{Entries: []ManifestEntry{{
-		OriginalPath: filepath.Join(t.TempDir(), "out.json"),
-		SnapshotPath: filepath.Join(t.TempDir(), "missing.json"),
+		OriginalPath: filepath.Join(tmpDir, "out.json"),
+		SnapshotPath: filepath.Join(tmpDir, "missing.json"),
 		Existed:      true,
 		Mode:         0o644,
 	}}})
@@ -74,6 +94,11 @@ func TestRestoreFailsWhenSnapshotMissing(t *testing.T) {
 // from a tar.gz archive when manifest.Compressed == true (BKUP-T31).
 func TestRestoreCompressedBackup(t *testing.T) {
 	home := t.TempDir()
+	// Override fns so validation accepts paths under t.TempDir().
+	origUserHomeDirFn := userHomeDirFn
+	t.Cleanup(func() { userHomeDirFn = origUserHomeDirFn })
+	userHomeDirFn = func() (string, error) { return home, nil }
+
 	backupDir := filepath.Join(home, "backup")
 
 	// Create a source file to snapshot.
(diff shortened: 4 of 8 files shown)


Alan-TheGentleman (changes requested):
## Adversarial Review — Judgment Day (2 judges, parallel blind review)

Good work identifying real vulnerabilities, @Basparin. The 5 attack vectors in #220 are legitimate and the intent of this PR is correct. However, the implementation has **1 confirmed blocker**, **2 confirmed security gaps**, and **10 broken tests** that prevent merge.

---

### BLOCKER: Validation hardcoded to `$HOME` — 10 test failures

**Files:** `internal/backup/manifest.go` (line 149–168) + `internal/backup/restore.go` (line 21–41)

`isRootDirUnderBackupRoot()` and `isPathUnderHome()` call `os.UserHomeDir()` directly with no injection point. Every test uses `t.TempDir()` which returns `/var/folders/...` (macOS) or `/tmp/...` (Linux) — never under `~/.gentle-ai/backups/`. Result: **10 tests fail**.

Failing tests:
- `TestDeleteBackup_Success`
- `TestRestoreRestoresExistingAndRemovesCreated`
- `TestRestoreCompressedBackup`, `TestRestoreUncompressedBackup`, `TestRestoreCompressedMultipleFiles`, `TestRestoreCompressedRemovesCreatedFiles`
- `TestPrune_DeletesOldestUnpinned`, `TestPrune_SkipsPinnedBackups`, `TestPrune_ReturnsDeletedIDs`
- `TestExecute_PrunesOldBackupsAfterUpgrade`

**Fix:** Use the `var fn = realFn` override pattern already established in this codebase (`engramDownloadFn`, `lookPathFn`, `detectOS`, `installScriptURLFn`):

```go
// manifest.go
var backupRootFn = backupRoot

func isRootDirUnderBackupRoot(dir string) (bool, error) {
    rootClean, err := backupRootFn()
    // ...
}
```

Then in tests: `backupRootFn = func() (string, error) { return testDir, nil }`

Same pattern for `isPathUnderHome` — extract `var userHomeDirFn = os.UserHomeDir`.

---

### WARNING (real) #1: PowerShell `-notlike` wildcard injection

**File:** `internal/system/path.go` (line 46–53)

`escapePowerShellString` only escapes `'` → `''` but does NOT escape `[`, `]`, `*`, `?` — these are glob metacharacters in PowerShell's `-notlike` operator. A path like `C:\Users\user\AppData\[Local]\bin` causes the `-notlike` check to fail with an invalid wildcard pattern error, breaking the upgrade.

**Fix:** Replace the `-notlike '*%s*'` glob check with an exact match:

```powershell
$current.Split(';') -notcontains '%s'
```

Or escape wildcards: `[System.Management.Automation.WildcardPattern]::Escape(...)`.

Also: add unit tests for `escapePowerShellString` with adversarial inputs (`'`, `"`, `` ` ``, `[`, `]`, `*`, `?`, null byte).

---

### WARNING (real) #2: TOCTOU race in `/tmp/gentleman-guardian-angel`

**File:** `internal/update/upgrade/strategy.go` (line 235)

`ggaTmpDir = "/tmp/gentleman-guardian-angel"` is a fixed, predictable, world-writable path. Race condition between `os.RemoveAll` and `git clone` allows a local attacker to plant a malicious `install.sh`. Ironic: `restoreCompressed` in this same PR correctly uses `os.MkdirTemp`.

**Fix:** `os.MkdirTemp("", "gentle-ai-gga-*")` instead of the fixed path.

Note: this is a pre-existing issue, not introduced by this PR, but since the PR is about security hardening, it should be addressed here.

---

### Additional findings (non-blocking but worth addressing)

| Finding | Severity | Details |
|---------|----------|---------|
| `isPathUnderHome` returns `false` silently when `UserHomeDir()` fails | WARNING (real) | Should return `(bool, error)` — silent denial hides the actual error |
| `EvalSymlinks` fallback asymmetry in `isRootDirUnderBackupRoot` | WARNING (theoretical) | If `EvalSymlinks(backupRoot)` fails, falls back to unresolved path — should error instead |
| `writeExecutable` in both `download.go` files has no size limit | WARNING (real) | `io.Copy` is unbounded — add `maxBinarySize` similar to `maxScriptSize` |
| `isPathUnderHome` permits writes to `~/.ssh/`, `~/.bashrc` | WARNING (theoretical) | Accepted design tradeoff but should be documented |
| `isRootDirUnderBackupRoot` name implies dirs but is used on file paths | SUGGESTION | Rename to `isPathUnderBackupRoot` |

---

### Summary

The vulnerabilities you identified are real and important. The PR needs:
1. **Injectable validation** (the `var fn` pattern) — fixes the blocker and 10 tests
2. **PowerShell wildcard escaping** — fixes the `-notlike` bypass
3. **`os.MkdirTemp`** for GGA temp dir — fixes the TOCTOU

After these fixes, re-run `go test ./internal/backup/... ./internal/update/upgrade/... ./internal/system/...` and verify 0 failures before re-requesting review.

Basparin:
## Addressed all 3 issues

Pushed `463c548` with fixes for the blocker + 2 warnings:

### 1. BLOCKER: Injectable validation (10 test failures) ✅
- Added `BackupRootFn` (exported) in `manifest.go` and `userHomeDirFn` in `restore.go`
- Follows the existing `var fn = realFn` pattern (`engramDownloadFn`, `detectOS`, etc.)
- All 10 tests updated with `t.Cleanup` overrides

### 2. WARNING: PowerShell wildcard injection ✅
- Replaced `-notlike '*%s*'` with `($current.Split(';')) -notcontains '%s'`
- Exact element matching — `[`, `]`, `*`, `?` in paths no longer break the check

### 3. WARNING: TOCTOU race in `/tmp/gentleman-guardian-angel` ✅
- Replaced fixed `ggaTmpDir` with `ggaMkdirTemp` function var using `os.MkdirTemp`
- Unpredictable path, `defer os.RemoveAll` for cleanup
- Added `detectOS` override in strategy tests for cross-platform compat

### Test results (local, Go 1.26.1 on Windows)
```
ok  github.com/gentleman-programming/gentle-ai/internal/backup
ok  github.com/gentleman-programming/gentle-ai/internal/update/upgrade
ok  github.com/gentleman-programming/gentle-ai/internal/system
```

Ready for re-review. 🤖

Alan-TheGentleman (approved):
## Re-judgment — Round 2 (Judgment Day)

All 3 blockers from Round 1 are correctly fixed:

1. **Injectable validation** ✅ — `BackupRootFn` + `userHomeDirFn` as package-level vars, tests override with `t.Cleanup`. All 10 previously-failing tests now pass.
2. **PowerShell wildcard injection** ✅ — Switched from `-notlike` (glob expansion) to `-notcontains` (exact match). Correct fix.
3. **TOCTOU `/tmp` path** ✅ — `os.MkdirTemp("", "gentle-ai-gga-*")` replaces the fixed `/tmp/gentleman-guardian-angel`. Correct fix.

### One non-blocking nit

`TestGGAScriptUpgradeUsesGitClone` doesn't override `ggaMkdirTemp`, so the test can't catch a regression back to a hardcoded path. Consider injecting and verifying the temp path in the test — but this is a test-quality nit, not a security issue in production.

Also: the comment on line 553 in `strategy_test.go` still references `/tmp/gentleman-guardian-angel` — stale after the TOCTOU fix.

### Verdict

**APPROVED** ✅ — Production code is correct. Ship it.

PR #265, opened 2026-04-09
fix(state): persist model assignments across sync runs

### LINKED ISSUE #264: bug(SDD profiles) new SDD profile replace the SDD default agent
### Pre-flight Checklist

### 📝 Bug Description

I created a new sdd profile, after configured the models for each sdd sub agent and saved&sync. I went go back to the default sdd orchestrator and I realized that all my configurations was missing. Instead of found my previous config I realized that the new configuration it's the same of my new sdd-profile config.
The new sdd-profile config it's replacing the default sdd-agent config.

### 🔄 Steps to Reproduce

1. Create a new SDD profile
2. Save & sync
3. Go back to "Configure models" 
4. Configure OpenCode models

### ✅ Expected Behavior

It should shows the original config of the default SDD-agent

### ❌ Actual Behavior

The config of the models of the original/default SDD it's replaced by the new sdd profile config.

### Gentle AI Version

1.18.4 & previous

### Operating System

Windows

### AI Agent / Client

OpenCode

### 📋 Affected Area

TUI (terminal UI)

### 💡 Logs / Error Output

```shell

```

### Additional Context

_No response_

christorres-hu:
Closes #264

## Summary
- Fixes model assignments being silently reset to "balanced" preset on every `sync` run
- CLI sync and TUI sync now load persisted assignments from `state.json` when no flags/overrides are provided
- Read-merge-write pattern ensures `InstalledAgents` and other state fields are preserved
- Added 7 integration tests covering CLI load path, TUI round-trip, backward compat, and multi-sync cycle

## Design note
`claudeAliasesToStrings` and `modelAssignmentsToState` are intentionally duplicated as unexported helpers in `app` and `cli` — avoids coupling `state` to `model` for two 8-line functions.

## Test plan
- [x] `TestRunSyncLoadsPersistedModelAssignments` — CLI sync loads from state.json
- [x] `TestRunSyncDoesNotOverridePersistedAssignmentsOnSecondSync` — assignments survive sync1 → sync2
- [x] `TestRunSyncWithNoPersistedAssignmentsDoesNotPanic` — backward compat with old state
- [x] `TestLoadPersistedAssignmentsPopulatesEmptySelection` — TUI load path
- [x] `TestLoadPersistedAssignmentsDoesNotOverrideExisting` — TUI overrides not clobbered
- [x] `TestPersistAssignmentsPreservesInstalledAgents` — read-merge-write safety
- [x] `TestPersistAssignmentsNoOpWhenEmpty` — no spurious writes
- [x] Full suite: `go test ./...` — 37 packages, 0 failures

### DIFF
--- a/e2e/e2e_test.sh
+++ b/e2e/e2e_test.sh
@@ -1339,25 +1339,30 @@ test_idempotent_full_claude() {
     $BINARY install --agent claude-code --component sdd --component persona --component context7 --component permissions --component theme --preset full-gentleman --persona gentleman 2>&1 || true
     local first_md_hash
     first_md_hash=$(md5sum "$HOME/.claude/CLAUDE.md" 2>/dev/null | cut -d' ' -f1)
-    local first_settings_hash
-    first_settings_hash=$(md5sum "$HOME/.claude/settings.json" 2>/dev/null | cut -d' ' -f1)
+    # Snapshot settings.json for semantic comparison (engram setup may reorder
+    # top-level keys on re-run — see engram binary's non-deterministic map
+    # serialization). Byte-exact hashing would false-fail on harmless reorder.
+    cp "$HOME/.claude/settings.json" /tmp/gai_settings_run1.json 2>/dev/null || true
 
     $BINARY install --agent claude-code --component sdd --component persona --component context7 --component permissions --component theme --preset full-gentleman --persona gentleman 2>&1 || true
     local second_md_hash
     second_md_hash=$(md5sum "$HOME/.claude/CLAUDE.md" 2>/dev/null | cut -d' ' -f1)
-    local second_settings_hash
-    second_settings_hash=$(md5sum "$HOME/.claude/settings.json" 2>/dev/null | cut -d' ' -f1)
 
     if [ "$first_md_hash" = "$second_md_hash" ] && [ -n "$first_md_hash" ]; then
         log_pass "Idempotent: CLAUDE.md identical after 2 runs"
     else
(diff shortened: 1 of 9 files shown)


Alan-TheGentleman (approved):
## Review: fix(state): persist model assignments across sync runs

**Verdict: APPROVE**

Clean, well-tested fix. The read-merge-write pattern is correct, backward compatibility is covered by the `omitempty` tags and the legacy-state test, and the 7 new tests cover the important scenarios (load, no-clobber, merge, no-op, multi-sync cycle, backward compat).

A few non-blocking observations:

1. **Duplicated helpers** — `claudeAliasesToStrings` / `modelAssignmentsToState` appear identically in both `app` and `cli`. Acknowledged in the PR description as intentional to avoid coupling. If a third caller appears, consider extracting to a shared internal package.

2. **Inline vs. extracted load logic** — The TUI path extracts `loadPersistedAssignments` as a clean helper, but `sync.go` has the same logic inline. Minor consistency nit.

3. **TUI persists after sync, CLI does not** — `tuiSync` calls `persistAssignments` after execution; `RunSync` does not. Currently harmless since nothing mutates assignments between load and execute, but the asymmetry is worth a comment for future maintainers.

4. **ModTime-based no-op test** — `TestPersistAssignmentsNoOpWhenEmpty` compares file timestamps. On coarse-grained filesystems this could theoretically be fragile. Content comparison would be more robust, but unlikely to flake in practice.

Overall: solid work, good test coverage, clear PR description. Ship it.

PR #269, opened 2026-04-09
fix(security): fail closed when checksums unavailable or mismatched

### LINKED ISSUE #245: security(installer): fail closed when checksums are unavailable or mismatched
### Pre-flight Checks

### Affected Area
Installation Pipeline

### Problem Statement
The binary installer verifies `checksums.txt` when present, but if the checksum metadata is missing it only warns and continues. That is a fail-open behavior at a supply-chain boundary.

### Proposed Solution
Make checksum verification mandatory by default:
- fail the install if `checksums.txt` is missing
- fail the install if the archive is not listed
- fail the install if the digest does not match
- optionally provide an explicit `--insecure` escape hatch for manual use

### Alternatives Considered
Continuing after a warning is more convenient, but it weakens the release-install trust boundary and can hide compromised or incomplete release metadata.

### Additional Context
Relevant path: `scripts/install.sh` binary install flow.

mitre88:
## Summary

Closes #245

Checksum verification in the binary install path was **fail-open**: if `checksums.txt` could not be downloaded, the archive name was absent from it, or no SHA256 tool was available, the installer logged a warning and continued. This allows installation of an unverified (potentially compromised) binary.

This PR hardens all three paths to **fail closed** by default, with an explicit escape hatch for environments where skipping is intentional.

### `scripts/install.sh`
- `Could not download checksums.txt` → `fatal` (was `warn`)
- `Archive not found in checksums.txt` → `fatal` (was `warn`)
- `No sha256sum/shasum found` → `fatal` (was `warn` + fake match)
- New `--insecure` flag to explicitly opt out of verification

### `scripts/install.ps1`
- `Could not download checksums.txt` (catch block) → `Stop-WithError` (was `Write-Warn`)
- `Archive not found in checksums.txt` → `Stop-WithError` (was `Write-Warn`)
- New `-Insecure` switch parameter

### `internal/update/upgrade/download.go`
The self-upgrade path (`Download`) had **zero checksum verification**. New functions added:
- `downloadToFile` — downloads to a temp file and returns SHA256 hex digest (via `io.MultiWriter`)
- `fetchChecksums` — fetches `checksums.txt`, returns error on HTTP non-200
- `expectedChecksumFor` — parses BSD-style checksums.txt, returns error if filename not listed
- `resolveArchiveName` / `resolveChecksumURL` — URL helpers (also improve testability)
- `Download` now downloads to a temp file, verifies checksum, then extracts — fail closed at every step

### `internal/update/upgrade/download_test.go`
New table-driven tests covering all four failure modes from the issue:
- `TestExpectedChecksumFor` — pure unit test for the checksum parser
- `TestFetchChecksums` — success + HTTP 404 → error
- `TestDownloadToFile` — verifies SHA256 computed correctly while streaming
- `TestDownload_ChecksumVerification` — four sub-cases: match (success), mismatch, missing `checksums.txt`, archive not listed

## Test plan

- [ ] `go test ./internal/update/upgrade/...` — all existing + new tests pass
- [ ] `go vet ./...` — no issues
- [ ] `./scripts/install.sh --insecure` — installs with warning, no fatal
- [ ] `./scripts/install.sh` — fails when `checksums.txt` unreachable (e.g. no network)


mitre88:
Note: I can't add labels from a fork. This PR needs the `type:bug` label for CI to pass.

Alan-TheGentleman (approved):
Solid security hardening — the fail-open to fail-closed transition is exactly right, and the test coverage for all four failure modes is thorough. The Go implementation correctly downloads to a temp dir, verifies the checksum, and only then extracts. LGTM.

A few minor suggestions (none blocking):

1. **`fetchChecksums`: add `io.LimitReader`** — the project already uses this pattern in `strategy.go`. Without it, a malicious server could force unbounded memory allocation. Something like `io.LimitReader(resp.Body, 1<<20)` would be sufficient.

2. **`downloadToFile`: explicit `f.Close()` before return** — the deferred close means write errors from `Close()` are silently dropped. Minor, but good practice for data-integrity code.

3. **Constant-time comparison** (`crypto/subtle.ConstantTimeCompare`) for the digest check. Negligible real-world risk for a CLI tool, but it's the standard practice for checksum verification.

4. **`downloadBinary` is now orphaned from `Download()`** — consider a `// Deprecated` note or removal in a follow-up to avoid confusion about which path is checksum-verified.

Nice work overall.

mitre88:
Gentle ping — this PR is ready for review. All CI checks pass. Would appreciate a look when possible. Thanks!

mitre88:
Bump — all CI checks pass (Unit Tests, E2E Tests, lint). This PR has been approved. Ready to merge. Could a maintainer hit the merge button? Thanks!

mitre88:
@Gentleman-Programming — this security fix has been approved and all CI passes (unit + E2E). It makes checksum validation fail-closed instead of fail-open. Could you merge? Thanks!

Alan-TheGentleman (commented):
This remains a merge candidate, but I want one fresh validation pass before merging security-sensitive install code.

Please rebase or refresh the branch against current main so CI proves the checksum hardening still works with today's installer and upgrade code.

PR #289, opened 2026-04-13
fix(persona): strip stale Gentleman outputStyle when switching to Neutral

### LINKED ISSUE #204: Neutral persona still uses Rioplatense output style (outputStyle not reset)
## Bug Description

When switching from **Gentleman** persona to **Neutral** persona, the `outputStyle: "Gentleman"` entry in `~/.claude/settings.json` is **not removed or reset**. This causes Claude Code to keep using the Gentleman output style (Rioplatense Spanish with voseo) even though the neutral persona explicitly says:

> "Use a warm, professional, and direct tone. No slang, no regional expressions."

## Root Cause

In `internal/components/persona/inject.go`, the output style is only written when `persona == PersonaGentleman`:

```go
if persona == model.PersonaGentleman && adapter.SupportsOutputStyles() {
    // writes gentleman.md and merges "outputStyle": "Gentleman" into settings.json
}
```

But when switching to `PersonaNeutral`, there is no cleanup step that:
1. Removes `"outputStyle": "Gentleman"` from `settings.json`
2. Optionally removes `~/.claude/output-styles/gentleman.md`

## Expected Behavior

When selecting neutral persona, the installer should either:
- Remove the `outputStyle` key from `settings.json`, or
- Set it to a neutral/default value

## Steps to Reproduce

1. Run `gentle-ai` and select **Gentleman** persona
2. Verify `~/.claude/settings.json` has `"outputStyle": "Gentleman"`
3. Run `gentle-ai` again and select **Neutral** persona
4. Check `~/.claude/settings.json` → `"outputStyle": "Gentleman"` is still there
5. Start a Claude Code session in Spanish → responses use Rioplatense voseo instead of neutral tone

## Environment

- OS: Linux (Ubuntu)
- Agent: Claude Code
- gentle-ai: latest

Basparin:
## 🔗 Linked Issue

Closes #204

## 🏷️ PR Type

- [x] \`type:bug\` — Bug fix (non-breaking change that fixes an issue)

## 📝 Summary

When a user switched from the **Gentleman** persona to **Neutral**, the \`outputStyle: \"Gentleman\"\` entry written to \`~/.claude/settings.json\` by the previous install was never cleaned up. Claude Code kept loading the Gentleman output style (Rioplatense Spanish, voseo, slang) even though the Neutral persona explicitly instructs a warm, professional, region-free tone.

Root cause was already pinpointed in the issue: in \`internal/components/persona/inject.go\`, the output-style merge was only performed for \`PersonaGentleman\`, with no symmetric cleanup path for other personas.

## 🛠️ Fix

\`Inject()\` now strips \`outputStyle\` from settings.json when persona is Neutral and the adapter supports output styles. Key rules:

- Only removes the key when its value is exactly \`\"Gentleman\"\` — any user-configured custom style (e.g. \`\"MyStyle\"\`) is preserved untouched
- Missing files, empty files, and malformed JSON are treated as no-ops, matching the existing tolerant-merge policy in \`filemerge\`
- \`PersonaCustom\` continues to short-circuit at the top of \`Inject()\` — user-managed configs are never touched
- The \`gentleman.md\` file under \`~/.claude/output-styles/\` is intentionally left on disk (harmless, potentially referenced by other tooling)

## 📂 Changes

|-------------|-------------|
| \`internal/components/persona/inject.go\` | Added step 4 in \`Inject()\` + new \`removeGentlemanOutputStyle()\` helper |
| \`internal/components/persona/inject_test.go\` | Two new tests covering the removal path and the custom-style preservation guarantee |

Total: 2 files, +133/-0.

## 🧪 Test Plan

- [x] New tests pass:
  - \`TestInjectClaudeNeutralRemovesGentlemanOutputStyleFromSettings\` — reproduces the bug without the fix, green with it
  - \`TestInjectClaudeNeutralPreservesUserCustomOutputStyle\` — guards against clobbering user-configured output styles
- [x] All 28 persona tests pass (\`go test ./internal/components/persona/\`)
- [x] Full suite pass (\`go test ./...\`)
- [x] \`go vet ./...\` clean
- [ ] E2E tests (\`cd e2e && ./docker-test.sh\`) — not run locally (Docker not available in this environment)

## ✅ Contributor Checklist

- [x] PR is linked to an issue with \`status:approved\` (#204)
- [x] I have added the appropriate \`type:*\` label to this PR
- [x] Unit tests pass (\`go test ./...\`)
- [ ] E2E tests pass — see test plan note
- [x] I have updated documentation if necessary (no user-facing docs change)
- [x] My commits do not include \`Co-Authored-By\` trailers

### DIFF
--- a/internal/components/filemerge/json_merge.go
+++ b/internal/components/filemerge/json_merge.go
@@ -30,6 +30,16 @@ func MergeJSONObjects(baseJSON []byte, overlayJSON []byte) ([]byte, error) {
 	return append(encoded, '\n'), nil
 }
 
+// UnmarshalJSONTolerant parses a JSON object while tolerating JSONC-style
+// features that frequent user-edited config files carry: line/block comments
+// and trailing commas. Callers that only need a plain map[string]any — with
+// the same parsing rules used by MergeJSONObjects — should use this helper
+// instead of json.Unmarshal directly so JSONC inputs are not silently
+// rejected.
+func UnmarshalJSONTolerant(raw []byte) (map[string]any, error) {
+	return unmarshalJSONObject(raw)
+}
+
 func unmarshalJSONObject(raw []byte) (map[string]any, error) {
 	object := map[string]any{}
 	if len(bytes.TrimSpace(raw)) == 0 {
--- a/internal/components/persona/inject.go
+++ b/internal/components/persona/inject.go
@@ -1,6 +1,8 @@
 package persona
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


Basparin:
cc @Alan-TheGentleman — could you add the `type:bug` label? Non-maintainers can't apply it. Thanks!

Basparin:
Ran a parallel adversarial review (two blind judges) on the original commit and pushed a follow-up fix in c90d72a. Findings:

### Confirmed bugs (fixed)

1. **JSONC silently skipped** — `removeGentlemanOutputStyle` called `json.Unmarshal` directly, which rejects line comments and trailing commas. Users who hand-edited `settings.json` with JSONC would silently fall into the malformed-JSON no-op branch and the stale `outputStyle: "Gentleman"` would never get cleaned. `filemerge.MergeJSONObjects` already tolerates JSONC on the write side — this fix exposes that same tolerance via a new `filemerge.UnmarshalJSONTolerant` wrapper and routes the read through it.
2. **Idempotency uncovered** — added a test that calls `Inject(..., PersonaNeutral)` twice in sequence and asserts the second call is a genuine no-op on `settings.json` (bytes stable, path not appended to `Files`).

### Considered but not action-worthy

- **Reformat to 2-space indent** — `json.MarshalIndent(obj, "", "  ")` canonicalises indentation. Not a regression: `filemerge.MergeJSONObjects` already does this on every Gentleman install, so any file touched by gentle-ai is already 2-space. User-only files with unusual indent are never touched because they don't contain the `outputStyle: "Gentleman"` key.
- **Multi-adapter coverage** — only Claude returns `true` for `SupportsOutputStyles()` today. The gate is correct.
- **File permissions / race** — both match the existing `mergeJSONFile` behaviour, no regression.

Total added: +104 lines (1 exported helper in filemerge, 2 new tests, doc updates).

PR #388, opened 2026-04-28
fix(sdd): ensure sub-agent responses are not lost when ending with tool calls

### LINKED ISSUE #387: fix(sdd): sub-agent responses lost when last action is a tool call (mem_save)
### Pre-flight Checks

### Bug Description

When an SDD phase sub-agent (or any sub-agent) finishes its work and calls `mem_save` as its **last action**, the orchestrator/parent agent only receives the Engram tool result (e.g., `"Observation saved"`) instead of the sub-agent's actual text response containing the analysis, findings, or return envelope.

This happens because the Task tool returns the sub-agent's **final output** to the parent. If that final output is a tool call result rather than text, the parent never sees the sub-agent's work.

### Steps to Reproduce

1. Launch any SDD phase (e.g., `sdd-explore`) via the Task tool
2. The sub-agent does its research, writes a text response, then calls `mem_save` to persist findings
3. The orchestrator receives `"Observation saved"` instead of the exploration results
4. The orchestrator has to either re-launch (wasting tokens and context) or guess what the sub-agent found

### Expected Behavior

The orchestrator should always receive the sub-agent's text analysis/return envelope, regardless of whether the sub-agent also saved to Engram.

### Actual Behavior

The orchestrator receives only the tool result from `mem_save` (e.g., `"Observation saved"`). The sub-agent's actual text response is lost.

### Root Cause

Three factors contribute:

1. **`sdd-phase-common.md` Section D** defines the return envelope format but never states that the text envelope must be the sub-agent's **absolute last output**
2. **`persistence-contract.md`** launch templates say "save before returning" but don't say "then return text **after** saving"
3. **The engram protocol** includes a Session Close Protocol (`mem_session_summary`) that sub-agents mistakenly interpret as their final action — but sub-agents should NOT call `mem_session_summary` at all

Additionally, sub-agents in OpenCode don't know they ARE sub-agents (the system prompt construction is identical for main and sub-agents — `session.parentID` is only used for HTTP headers, not injected into the prompt). So any fix must be **unconditional** — it should apply to all agents, not conditionally to sub-agents.

### Proposed Fix (3 files, ~15 lines total)

**1. `internal/assets/skills/_shared/sdd-phase-common.md`** — Add to Section D (Return Envelope):

```markdown
> **CRITICAL — Response ordering**: Your FINAL output MUST be text (the return envelope), NOT a tool call. If you need to save to Engram (`mem_save`), do it BEFORE your final text response. Do NOT call `mem_session_summary` — that's for top-level agents only. **Why**: When a sub-agent's last action is a tool call, the parent agent receives only the tool result — your text response (the actual analysis) is lost.
```

**2. `internal/assets/skills/_shared/persistence-contract.md`** — Add new section:

```markdown
## Sub-Agent Response Ordering

When a sub-agent persists artifacts (via `mem_save` or file writes), the persistence call MUST happen BEFORE the final text response. The sub-agent's absolute last output must be text, never a tool call.

**Why**: The parent agent only receives the sub-agent's final output. If the last action is a tool call (e.g., `mem_save`), the parent sees the tool result ("Observation saved") instead of the sub-agent's actual analysis and findings.

Order: do work → persist artifact (`mem_save`) → return text summary as final action.
```

**3. `internal/assets/generic/sdd-orchestrator.md`** — Add orchestrator-side recovery:

```markdown
### Empty Sub-Agent Response Recovery

When a sub-agent returns empty, minimal, or tool-result-only output (e.g., "Observation saved", a raw tool response, or blank):

1. Do NOT launch a new sub-agent — this wastes the completed work and full context.
2. RESUME the same task using the existing `task_id`, adding a prompt: "Your work completed but I received no text response. Please provide your findings/summary as text now."
3. If the second attempt also returns empty, THEN report the issue to the user with the task ID and last output received.

**Why**: Sub-agents sometimes end with a tool call (e.g., `mem_save`) as their last action, causing the orchestrator to receive only the tool result instead of the analysis. Resuming preserves the sub-agent's full context; re-launching discards all work done.
```

### Defense in Depth

This is a two-layer fix:
- **Prevention**: Tell all agents to always end with text (files 1 & 2)
- **Recovery**: Tell the orchestrator to retry with the same task_id when it gets empty output (file 3)

Both layers are needed because LLMs don't always follow instructions perfectly.

### Operating System

Linux

### Agent / Client

OpenCode

### Shell

bash

### Additional Context

We've been running this fix in production for several weeks with consistent success. The empty-response recovery (file 3) catches the remaining ~5% of cases where the sub-agent still ends with a tool call despite the prevention instructions.

ithril-dev:
## Summary

Fixes #387

When a sub-agent's last action is a tool call (`mem_save`, `mem_session_summary`), the parent/orchestrator agent only receives the tool result (e.g., "Observation saved") instead of the sub-agent's actual text response containing the analysis and return envelope.

This has been running in production for several weeks with consistent success.

## Changes

Three-layer defense-in-depth fix across 3 files (~20 lines total):

### 1. Prevention — `sdd-phase-common.md` (Section D)
Added a `CRITICAL — Response ordering` callout at the top of the Return Envelope section: final output MUST be text, not a tool call. Explains **why** (parent loses the response).

### 2. Prevention — `persistence-contract.md`
New `Sub-Agent Response Ordering` section establishing the universal rule: `do work → mem_save → return text as final action`.

### 3. Recovery — `sdd-orchestrator.md`
New `Empty Sub-Agent Response Recovery` section for the orchestrator: when a sub-agent returns empty/tool-result-only output, **resume the same task_id** asking for the text response instead of launching a new sub-agent (which wastes tokens and discards context).

## Why both prevention AND recovery?

LLMs don't always follow instructions perfectly. The prevention layer (files 1 & 2) handles ~95% of cases. The recovery layer (file 3) catches the remaining ~5% where the sub-agent still ends with a tool call despite explicit instructions.

## Additional context

- Sub-agents in the current architecture don't know they ARE sub-agents (system prompt construction is identical — `session.parentID` is only used for HTTP headers). So the prevention rules are written as unconditional ("always end with text") rather than conditional ("if you are a sub-agent...").
- A future improvement could inject a sub-agent indicator into the system prompt when `session.parentID` is set, enabling more targeted instructions.


Alan-TheGentleman (changes requested):
This looks small and valuable, but it still needs the normal process gates.

Please add the required `type:bug` or `type:docs` label, confirm the approved issue link, and refresh CI against current main.

Alan-TheGentleman:
Thanks for the contribution. I’m closing this PR because it does not link to an issue with `status:approved`, which is required by this repo before PR work can proceed.

Please follow the project process:
1. Open or update the relevant issue with the proposed change and reproduction/context.
2. Wait for a maintainer to add `status:approved`.
3. Open a fresh PR that includes `Closes #<issue>` and exactly one `type:*` label.

This keeps review scope clear and prevents CI/process failures.

Alan-TheGentleman:
Hey, thanks for contributing!

Per [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md), every PR needs an issue linked with `status:approved` before it can be reviewed — and this PR doesn't satisfy that gate (the linked issue doesn't exist, is closed, or isn't approved).

Closing for now. To pick the work back up:
1. Open (or find) an issue describing the change
2. Wait for a maintainer to apply `status:approved`
3. Reopen this PR or open a new one with `Closes #<issue>` in the body

Sorry for the friction — the gate is automated, there's no workaround. Come back when it's ready!

Alan-TheGentleman:
Hey @ithril-ai — quick heads-up: the underlying bug (#387 sub-agent response ordering) is now resolved on `main` via PR #668 (merged). The fix took a different approach: instead of patching the agent template, we updated the shared `internal/assets/skills/_shared/sdd-phase-common.md` and `persistence-contract.md` to instruct ALL sub-agents that their final output must be text (never a tool call). Because every agent's skill bundle pulls from `_shared/`, the fix applies unconditionally — including OpenCode sub-agents that can't detect their own role.

Thanks for the report and the original PR — the discussion here helped sharpen the final solution.

PR #389, opened 2026-04-28
fix(vscode): keep engram available at session start

### LINKED ISSUE #384: feat(vscode): auto-connect Engram MCP and fail fast when memory is unavailable
## Pre-flight Checklist

## 🔍 Affected Area
Other

## 💡 Problem Statement
The current VS Code flow works only partially for Engram-backed memory. VS Code does not always start/connect the Engram MCP by default, so the agent can begin working in a degraded state without reliable memory access.

In practice this causes two user-facing problems:
1. The agent does not use Engram because the MCP is present in config but not actually connected/running.
2. Even when Engram is running, the agent can still skip the moments where it should search/save memory unless the user adds stronger behavioral rules to `preferences.md`.

As a workaround, I had to manually strengthen `preferences.md` with rules like:
- Engram must be active in every session
- If the Engram MCP is not connected, warn immediately and stop
- Search memory before project work / continuation work
- Save memory after decisions, bugfixes, and non-trivial work

That workaround improves behavior, but it is compensating for a host/runtime integration problem plus a weak degraded-mode policy.

## 📦 Proposed Solution
Improve the VS Code integration so Engram is first-class and reliable by default:

1. Auto-start / auto-connect the Engram MCP for supported VS Code agent flows.
2. Add a session-start health check / handshake that verifies Engram tool availability before project work begins.
3. If Engram is unavailable, enter explicit degraded mode:
   - warn the user immediately
   - explain memory-dependent behavior is unavailable
   - avoid pretending memory exists
4. Add stronger built-in behavior for memory-aware workflows so users do not need to patch `preferences.md` just to get correct Engram usage.
5. Optionally expose a diagnostic command / status indicator for MCP connectivity in VS Code.

## 🔄 Alternatives Considered
- Keep documenting `preferences.md` workarounds, but this is fragile and editor-specific.
- Rely only on prompt instructions, but prompt rules cannot start a disconnected MCP.
- Ask users to manually verify MCP status every session, but that is poor UX.

## 📎 Additional Context
This feels like a combination of:
- an infrastructure issue: the MCP is not reliably connected by default in VS Code
- a behavior issue: the agent can silently skip Engram usage instead of failing fast

A robust fix should handle both.

sarredondq:
Closes #384

## PR Type
- [x] Bug fix
- [ ] New feature
- [ ] Documentation only
- [ ] Code refactoring
- [ ] Maintenance/tooling
- [ ] Breaking change

## Summary
- Write VS Code Engram MCP config in both the global user profile and the active workspace so Copilot can discover the server where the session actually starts.
- Enable VS Code's `chat.mcp.autostart` setting and add a VS Code-specific startup instruction that asks for a single lightweight Engram availability check only on brand-new sessions.
- Reduce the chance that Engram silently drops out of the session by combining bootstrap configuration with an explicit low-cost session-start reminder.

## Why
The problem behind #384 is not just "config exists"; it's that VS Code can start a new Copilot session without actually having Engram ready in the place or moment the agent needs it. When only the user-level `mcp.json` is written, workspace-scoped Copilot flows can still miss the server bootstrap. And even when MCP wiring exists, the session can drift into working without memory if nothing reminds the agent to verify availability at the beginning.

These changes attack both failure modes together:
- the MCP bootstrap becomes more reliable by writing `.vscode/mcp.json` alongside the global VS Code config;
- VS Code gets `chat.mcp.autostart=true` so the editor is more willing to restart MCP servers automatically;
- the generated VS Code instructions now require one lightweight Engram availability check only at the start of a brand-new session, which keeps token cost low while nudging the agent to notice when memory is unavailable before substantive work begins.

The goal is simple: make Engram much more likely to be present from the first turn of a session, so memory usage does not quietly disappear and force the user to recover context manually.

## Changes
| File | Change |
|------|--------|
| `internal/agents/vscode/adapter.go` | Added workspace MCP path support for VS Code. |
| `internal/agents/vscode/adapter_test.go` | Covered the workspace MCP path behavior. |
| `internal/components/engram/inject.go` | Added optional workspace-aware injection, VS Code autostart settings merge, and workspace-local MCP writing. |
| `internal/components/engram/inject_test.go` | Added regression coverage for VS Code settings and workspace MCP generation. |
| `internal/cli/run.go` | Passed the workspace directory into Engram injection during install/apply. |
| `internal/cli/sync.go` | Passed the workspace directory into Engram injection during sync. |
| `internal/assets/vscode/persona-gentleman.md` | Added a VS Code-specific instructions asset with a one-time session-start Engram availability check. |
| `internal/components/persona/inject.go` | Routed VS Code persona generation to the new VS Code-specific asset. |
| `internal/components/persona/inject_test.go` | Asserted the new VS Code startup instruction is present. |
| `internal/assets/assets.go` | Embedded the new VS Code asset directory. |
| `internal/assets/assets_test.go` | Included the VS Code persona asset in shared asset validations. |

## Test Plan
- [x] Manually tested the affected functionality through focused code-path verification.
- [x] Skills load correctly in target agent.
- [x] Relevant tests pass: `go test ./internal/components/persona ./internal/assets`

## Contributor Checklist
- [x] Linked an approved issue
- [ ] Ran shellcheck on modified scripts (no shell scripts changed)
- [x] Skills tested in at least one agent
- [x] Docs updated if behavior changed (not required for this scoped runtime/config fix)

### DIFF
--- a/internal/agents/vscode/adapter.go
+++ b/internal/agents/vscode/adapter.go
@@ -95,6 +95,10 @@ func (a *Adapter) MCPConfigPath(homeDir string, _ string) string {
 	return filepath.Join(a.vscodeUserDir(homeDir), "mcp.json")
 }
 
+func (a *Adapter) WorkspaceMCPConfigPath(workspaceDir string, _ string) string {
+	return filepath.Join(workspaceDir, ".vscode", "mcp.json")
+}
+
 func (a *Adapter) vscodeUserDir(homeDir string) string {
 	switch runtime.GOOS {
 	case "darwin":
--- a/internal/agents/vscode/adapter_test.go
+++ b/internal/agents/vscode/adapter_test.go
@@ -91,3 +91,14 @@ func TestMCPConfigPathUsesVSCodeUserProfile(t *testing.T) {
 		}
 	}
 }
+
+func TestWorkspaceMCPConfigPathUsesVSCodeWorkspaceFolder(t *testing.T) {
+	a := NewAdapter()
+	workspace := "/tmp/workspace"
+
+	path := a.WorkspaceMCPConfigPath(workspace, "engram")
+	want := filepath.Join(workspace, ".vscode", "mcp.json")
+	if path != want {
+		t.Fatalf("WorkspaceMCPConfigPath() = %q, want %q", path, want)
+	}
+}
--- a/internal/assets/assets.go
+++ b/internal/assets/assets.go
@@ -2,7 +2,7 @@ package assets
 
 import "embed"
 
-//go:embed all:claude all:opencode all:generic all:skills all:gga all:gemini all:codex all:antigravity all:windsurf all:cursor all:kimi all:qwen all:kiro
+//go:embed all:claude all:opencode all:generic all:skills all:gga all:gemini all:codex all:antigravity all:windsurf all:cursor all:kimi all:qwen all:kiro all:vscode
 var FS embed.FS
 
 // MustRead returns the content of an embedded file or panics.
--- a/internal/assets/assets_test.go
+++ b/internal/assets/assets_test.go
@@ -196,6 +196,7 @@ func TestGentlemanLanguageInstructionsDoNotBiasEnglishSessions(t *testing.T) {
 		"kiro/persona-gentleman.md",
 		"kimi/persona-gentleman.md",
 		"opencode/persona-gentleman.md",
+		"vscode/persona-gentleman.md",
 	}
 
 	for _, path := range personaPaths {
--- a/internal/assets/vscode/persona-gentleman.md
+++ b/internal/assets/vscode/persona-gentleman.md
@@ -1,0 +1,58 @@
+## Rules
+
+- Never add "Co-Authored-By" or AI attribution to commits. Use conventional commits only.
+- Never build after changes.
+- Response-length contract: default to short answers. Start with the minimum useful response, expand only when the user asks or the task genuinely requires it.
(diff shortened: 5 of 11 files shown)


Alan-TheGentleman (changes requested):
This needs process cleanup before technical review.

Please add the required `type:bug` label, confirm the approved issue link, and refresh CI against current main. The VS Code Engram bootstrap behavior is sensitive enough that we should review it only once the branch is current.

Alan-TheGentleman:
Thanks for the contribution. I’m closing this PR because it does not link to an issue with `status:approved`, which is required by this repo before PR work can proceed.

Please follow the project process:
1. Open or update the relevant issue with the proposed change and reproduction/context.
2. Wait for a maintainer to add `status:approved`.
3. Open a fresh PR that includes `Closes #<issue>` and exactly one `type:*` label.

This keeps review scope clear and prevents CI/process failures.

Alan-TheGentleman:
Hey, thanks for contributing!

Per [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md), every PR needs an issue linked with `status:approved` before it can be reviewed — and this PR doesn't satisfy that gate (the linked issue doesn't exist, is closed, or isn't approved).

Closing for now. To pick the work back up:
1. Open (or find) an issue describing the change
2. Wait for a maintainer to apply `status:approved`
3. Reopen this PR or open a new one with `Closes #<issue>` in the body

Sorry for the friction — the gate is automated, there's no workaround. Come back when it's ready!

PR #399, opened 2026-04-29
fix(sync): tolerate per-agent failures in component pipeline

ezequielmm:
## Summary

`componentSyncStep.Run` aborts the entire sync pipeline as soon as **any** managed agent fails its write step (sdd, skills, permissions or theme components). The user is left with a fatal `execute sync pipeline: ...` error even when only one agent had a problem and every other agent would have synced cleanly.

This PR makes per-agent failures non-fatal by default, while preserving the original failure semantics when **every** agent fails.

## Motivation — real-world reproduction

A user with five managed agents (`claude-code`, `gemini-cli`, `cursor`, `codex`, `antigravity`) hit a permission-denied while sync tried to create `~/.gemini/antigravity/skills/_shared/SKILL.md`. That single permission issue killed the entire sync — `claude-code`, `cursor` and `codex` were never touched, even though their writes would have succeeded.

The only recovery is `gentle-ai sync --agents <list-without-the-broken-one>`, which requires the user to know in advance which agent will break the run. That is unfriendly and surprising — `sync` is supposed to be the recovery command, not the one that demands you already know what's wrong.

## Changes

Single commit, one file, +46/-8.

* New helper `reportPerAgentErrors(errs, totalAgents)` in `internal/cli/sync.go`:
  * No errors → returns `nil`.
  * Some agents failed, at least one succeeded → each error is logged as a warning (`sync: warning: ...`) and the function returns `nil` so the pipeline continues.
  * **Every** agent failed → errors are joined with `errors.Join` and returned, preserving the existing fail-and-rollback behavior.
* The four iteration sites (`ComponentSDD`, `ComponentSkills`, `ComponentPermission`, `ComponentTheme`) now collect errors instead of returning the first one, then call the helper.
* No change to `ComponentGGA` (it is not a per-adapter loop).
* No change to public APIs, no change to the TUI dispatcher, no change to flag parsing.

## Why this is the right shape

* Mirrors the convention already used by `backup.Prune`: housekeeping / partial failures are non-fatal warnings, true failures escalate.
* Keeps "all-or-nothing" semantics intact when truly nothing worked — operators who script around exit codes still get a non-zero exit when the sync produced no useful effect.
* Single-agent runs (the TUI's typical case, `--agents <one>`) behave exactly as before — there is no observable change unless the user has multiple agents and one of them fails.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./...` — full suite passes (39 packages green; cli, tui, app, components/sdd all run on the modified code path)
- [x] Manual: with multiple agents, simulate a write failure on one agent; the pipeline now completes with a `sync: warning: ...` line on stderr and a non-zero `FilesChanged` from the others.
- [ ] Maintainer can reproduce by `chown root` on `~/.gemini/antigravity/skills/_shared/` and running `gentle-ai sync` (multi-agent install).

## Out of scope

* Surfacing per-agent warnings inside `SyncResult` and `RenderSyncReport` so they appear in the user-facing report (currently they go to `log.Printf`/stderr only). That is a UX improvement worth doing in a follow-up PR with reviewer input on the report shape.
* Refining `backupExcludeSubdirs` / sync's permission-denial recovery (separate concern, separate discussion).

### DIFF
--- a/internal/cli/sync.go
+++ b/internal/cli/sync.go
@@ -2,8 +2,10 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -515,6 +517,7 @@ func (s componentSyncStep) Run() error {
 			sddMode = model.SDDModeMulti
 		}
 
+		var sddErrs []error
 		for _, adapter := range adapters {
 			opts := sdd.InjectOptions{
 				OpenCodeModelAssignments:           s.selection.ModelAssignments,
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman:
The bug report is clear, but I am closing this PR because it skips the issue-first process. No PR merges here without a linked `status:approved` issue. Open the bug as an issue first, get it approved, then re-file or reopen the PR with the link so we can review the code on its merits.

PR #432, opened 2026-05-05
ci(e2e): parallelize platforms with BuildKit cache

### LINKED ISSUE #431: feat(ci): parallelize E2E across platforms with BuildKit cache
### Pre-flight Checks

### Problem Description

The E2E job in `.github/workflows/ci.yml` runs all three Linux platforms (Ubuntu, Arch, Fedora) **serially** in a single GitHub Actions runner via the loop in `e2e/docker-test.sh`. Each Dockerfile rebuilds from scratch on every CI run — re-downloading the Go 1.24.2 tarball, re-installing Node via NodeSource, and re-running `go mod download` and `go build` with no layer cache between runs. The CI timeout was recently extended to 45min in commit `5f6af97` ("ci: extend e2e timeout for slow mirrors") which is a symptom of hitting the ceiling.

Wall-clock today is roughly the **sum** of the three platform builds plus run time, with zero reuse across runs.

### Proposed Solution

Three independent infrastructure improvements that do **not** change test coverage:

1. **GitHub Actions matrix**: split the E2E job into a `strategy.matrix.platform: [ubuntu, arch, fedora]` so each platform runs on its own runner in parallel. Wall-clock drops from "sum of 3" to "max of 3".

2. **BuildKit cache via `docker/build-push-action`**:
   - `cache-from: type=gha` + `cache-to: type=gha,mode=max` to persist Docker layers between CI runs
   - `load: true` so the built image is available to `docker run` after the build (default behaviour leaves it in the cache only)
   - Add `# syntax=docker/dockerfile:1` to each Dockerfile so `RUN --mount=type=cache` works reliably across Docker versions
   - Add Go cache mounts to `go mod download` and `go build`: `--mount=type=cache,target=/root/.cache/go-build` and `--mount=type=cache,target=/root/go/pkg/mod` (paths reflect that the build runs as **root** before the `USER testuser` switch — Go's default GOPATH is `/root/go`, not `/go`).

3. **Remove `actions/setup-go` from the E2E job**: the Go binary is unused outside the Docker container (the build happens inside it), so the step adds ~30s with no value.

Test coverage remains identical — `RUN_FULL_E2E=1` and `RUN_BACKUP_TESTS=1` still run on every PR. This issue is purely infrastructure.

### Affected Area
CI/Workflows

### Alternatives Considered

- **Keep as-is**: the timeout extension already documents the pain. Continuing to absorb it is not free — every PR pays.
- **Reduce coverage on PRs (PR smoke vs main full)**: a separate, complementary change. Tracked as a follow-up issue rather than bundled here, because it changes coverage policy and warrants its own review. This issue keeps coverage identical and is therefore lower-risk.
- **Refactor dry-run tests to Go unit tests**: a much larger, multi-day effort. Out of scope for this issue.

### Additional Context

Branch protection rules pointing at the current `E2E Tests` check name will need to be updated to the new per-platform check names (`E2E Tests (ubuntu)`, `E2E Tests (arch)`, `E2E Tests (fedora)`) after this lands. Mentioned for transparency; the maintainer will handle that as part of merge prep.

Alan-TheGentleman:
Closes #431

### PR Type

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation only
- [ ] Code refactoring
- [x] Maintenance/tooling
- [ ] Breaking change

### Summary

- Split the E2E job into a `strategy.matrix` over `[ubuntu, arch, fedora]` so platforms run in parallel runners instead of looping serially inside `docker-test.sh`.
- Wire up `docker/build-push-action@v6` with `load: true` and per-platform GHA cache (`cache-from: type=gha,scope=e2e-<name>`) so Docker layers persist across CI runs. Add `# syntax=docker/dockerfile:1` and `RUN --mount=type=cache` for `/root/.cache/go-build` and `/root/go/pkg/mod` to each Dockerfile (paths target `/root` because the build runs before the `USER testuser` switch).
- Drop the `actions/setup-go` step from the E2E job — the Go toolchain is unused outside the container, the build happens inside Docker.

Test coverage is unchanged: `RUN_FULL_E2E=1` and `RUN_BACKUP_TESTS=1` still run on every PR across all three platforms. `e2e/docker-test.sh` is left untouched for local development.

### Changes

| File | Change |
|------|--------|
| `.github/workflows/ci.yml` | Replace serial E2E job with `strategy.matrix` over 3 platforms; switch from `docker build` + `docker run` to `docker/build-push-action@v6` with GHA cache and `load: true`; drop `actions/setup-go` from E2E job; reduce job timeout from 45→40m and run-step timeout from 40→30m (per-platform now). |
| `e2e/Dockerfile.ubuntu` | Add `# syntax=docker/dockerfile:1`; add `--mount=type=cache` to `go mod download` and `go build` targeting `/root/.cache/go-build` and `/root/go/pkg/mod`. |
| `e2e/Dockerfile.arch` | Same as Ubuntu. |
| `e2e/Dockerfile.fedora` | Same as Ubuntu. |

### Test Plan

- [x] Workflow YAML parses (no schema violations in editor).
- [x] Conventional commit format respected (`ci(e2e): ...`).
- [ ] CI runs the new matrix on this PR — three parallel `E2E Tests (<platform>)` checks should appear.
- [ ] First run will populate the GHA cache; subsequent runs should show meaningful cache hits in the build logs (`CACHED` markers on toolchain layers).

### Notes for Reviewer

After merge, **branch protection rules pointing at the old `E2E Tests` check name will need to be updated** to require the three new per-platform checks (`E2E Tests (ubuntu)`, `E2E Tests (arch)`, `E2E Tests (fedora)`). Mentioned in the linked issue too.

Follow-up work (separate PRs):

- PR 2: PR-smoke-vs-main-full coverage policy split (PR runs Ubuntu Tier 1 only; main and nightly run full matrix).
- Future: migrate dry-run tests from `e2e/e2e_test.sh` to Go unit tests so they no longer need Docker.

### Contributor Checklist

- [x] Linked an approved issue (#431, `status:approved`)
- [x] Added exactly one `type:*` label (`type:chore`)
- [x] No shell scripts modified — `shellcheck` not applicable
- [x] CI workflow change does not modify skill content
- [x] No behavior changes — coverage identical to current setup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,26 +23,43 @@ jobs:
         run: go test ./...
 
   e2e-tests:
-    name: E2E Tests
+    name: E2E Tests (${{ matrix.platform.name }})
     runs-on: ubuntu-latest
-    timeout-minutes: 45
+    timeout-minutes: 40
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: ubuntu
+            dockerfile: Dockerfile.ubuntu
+          - name: arch
+            dockerfile: Dockerfile.arch
+          - name: fedora
+            dockerfile: Dockerfile.fedora
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up Go
-        uses: actions/setup-go@v5
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build E2E image
+        uses: docker/build-push-action@v6
         with:
-          go-version: "1.24"
+          context: .
+          file: e2e/${{ matrix.platform.dockerfile }}
+          tags: gentle-ai-e2e-${{ matrix.platform.name }}:ci
+          load: true
+          cache-from: type=gha,scope=e2e-${{ matrix.platform.name }}
+          cache-to: type=gha,mode=max,scope=e2e-${{ matrix.platform.name }}
 
       - name: Run E2E tests
-        timeout-minutes: 40
+        timeout-minutes: 30
         env:
-          RUN_FULL_E2E: "1"
-          RUN_BACKUP_TESTS: "1"
-          E2E_PLATFORM_TIMEOUT_SECONDS: "1800"
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          cd e2e
-          chmod +x docker-test.sh
-          ./docker-test.sh
+          docker run --rm \
+            -e RUN_FULL_E2E=1 \
+            -e RUN_BACKUP_TESTS=1 \
+            -e GITHUB_TOKEN \
+            gentle-ai-e2e-${{ matrix.platform.name }}:ci
--- a/e2e/Dockerfile.arch
+++ b/e2e/Dockerfile.arch
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 # Dockerfile.arch — E2E test image for gentle-ai on Arch Linux
 # Builds the binary from source and runs the E2E test suite as a non-root user.
 # archlinux:base only provides x86_64 images. Force amd64 platform so Docker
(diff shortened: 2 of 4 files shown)


Alan-TheGentleman:
Validated. Scoped, linked to the approved issue, and CI is green. The important BuildKit details are correct: `load: true` is present for the follow-up `docker run`, cache scopes are per platform, and the Go cache mounts use the root paths where the build actually runs. The Azure Ubuntu mirror is a CI-oriented tradeoff, but the passing matrix proves it fixes the current stall without changing E2E coverage. Good to merge.

Issue #392, opened 2026-04-28
feat(models): merge custom providers from opencode.json into model picker


DiegoT4l:
## Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

---

## 🔍 Affected Area

Agent Detection

## 💡 Problem Statement

The model picker screen (`model_picker.go`) loads providers exclusively from `~/.cache/opencode/models.json` via `LoadModels()` and then filters them with `DetectAvailableProviders()`. This cache only contains OpenCode's built-in providers (Google, Anthropic, OpenCode Zen, etc.).

Custom providers defined in `~/.config/opencode/opencode.json` under the `"provider"` key are **completely ignored**. This affects any user who has configured a custom provider — whether it's an OpenAI-compatible proxy, a self-hosted model, a LiteLLM gateway, or any third-party aggregator. For example:

```json
{
  "provider": {
    "my-provider": {
      "name": "My Custom Provider",
      "npm": "@ai-sdk/openai-compatible",
      "options": { "baseURL": "https://my-api.example.com/v1" },
      "models": {
        "model-a": { "name": "Model A" },
        "model-b": { "name": "Model B" }
      }
    }
  }
}
```

When this user runs `gentle-ai` to configure SDD agents, the model picker only shows the built-in providers. None of the models from `"my-provider"` appear. This forces users to manually edit `opencode.json` after running `gentle-ai` to point agents to their custom provider's models, which defeats the purpose of having an interactive installer.

## 📦 Proposed Solution

In `internal/opencode/models.go`, after loading providers from the cache (`LoadModels()`), also parse the `"provider"` section from `opencode.json` (using `DefaultSettingsPath()`) and merge those custom providers into the map before calling `DetectAvailableProviders()`.

Key changes:

1. **New function** `LoadCustomProviders(settingsPath string) (map[string]Provider, error)` — reads `opencode.json`, extracts the `"provider"` object, and converts each entry to a `Provider` struct. Since custom providers use the `@ai-sdk/openai-compatible` format (with `npm`, `options.baseURL`, and a `models` map), the model entries typically only have `name` — so `tool_call` should default to `true` for all custom models (or be inferred).

[...]

Issue #391, opened 2026-04-28
feat(opencode-sdd): allow selecting fast model variants like gpt-5.5-fast in SDD profiles


eduardoemb:
### Pre-flight Checklist

- [x] I have searched existing issues and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

### Affected Area

OpenCode SDD profiles / TUI model picker

### Problem Statement

When creating or editing an OpenCode SDD profile through Gentle AI, the SDD model assignment flow does not allow selecting OpenCode fast model variants such as:

```text
openai/gpt-5.5-fast
```

This blocks generating an SDD profile that uses GPT-5.5 in fast mode directly from the Gentle AI tool.

This is not an OpenCode limitation. OpenCode itself exposes the model:

```sh
opencode models openai | rg '^openai/gpt-5\.5'
```

Result:

```text
openai/gpt-5.5
openai/gpt-5.5-fast
```

OpenCode also supports model variants/reasoning effort through `variant`, for example:

```json
{
  "model": "openai/gpt-5.5-fast",
  "variant": "high"
}
```

Manual editing of `~/.config/opencode/opencode.json` works, but the Gentle AI SDD generation flow should be able to produce this configuration itself.

### Expected Behavior

The OpenCode SDD profile generation/editing flow should allow selecting `openai/gpt-5.5-fast` for SDD phases and/or the orchestrator.

Ideally, it should also allow setting the OpenCode model variant/reasoning effort, such as:

```json
"variant": "high"
```

When saved/synced, generated SDD agents should preserve both fields:

```json
"model": "openai/gpt-5.5-fast",
"variant": "high"
```

### Actual Behavior

The SDD tool/model picker does not expose a way to generate the profile with `openai/gpt-5.5-fast` selected.

The current workaround is to manually edit `opencode.json` after generation and update the generated SDD agents by hand.

### Workaround

Manually patch every generated SDD agent in `~/.config/opencode/opencode.json`, for example:

```json
"sdd-orchestrator-openai55": {
  "model": "openai/gpt-5.5-fast",
  "variant": "high"
}
```

The same manual change is needed for phase agents such as `sdd-init-*`, `sdd-explore-*`, `sdd-propose-*`, `sdd-spec-*`, `sdd-design-*`, `sdd-tasks-*`, `sdd-apply-*`, `sdd-verify-*`, `sdd-archive-*`, and `sdd-onboard-*`.

[...]

PR #440, opened 2026-05-05
feat(tui): per-model variant/effort selection in OpenCode model picker

### LINKED ISSUE #429: feat(tui): per-model variant/effort selection in OpenCode model picker
## Pre-flight Checklist

- [x] I searched existing issues and found related but no umbrella: #391, #392
- [x] I understand this issue needs `status:approved` before a PR can be opened

## Affected Area

TUI

## Problem Statement

The OpenCode model picker assigns provider/model pairs to SDD phases, but has no way to select the **variant (effort level)** per model. Each provider exposes different effort levels (e.g., OpenAI: `none`/`low`/`medium`/`high`/`xhigh`; Anthropic: `low`/`medium`/`high`/`max`), and the correct field in `opencode.json` is `"variant"` — a config-level abstraction that OpenCode maps to provider-specific parameters internally.

Without this, users must manually edit `opencode.json` to set `"variant": "high"` on each agent definition. This is error-prone and breaks when `gentle-ai sync` overwrites the config.

### Related Issues

- **#391** — Requests variant selection for fast model variants in SDD profiles. This umbrella covers #391's scope plus TUI selection, persistence, and plugin-based data sourcing.
- **#392** — Requests merging custom providers into the picker. Orthogonal but shares the model picker surface area.
- **#390** (closed) — Discoverability of OpenAI models. The variant picker improves model selection UX in the same area.

## Proposed Solution

End-to-end per-model variant selection with zero hardcoded effort levels:

### 1. Data source: OpenCode plugin (`model-variants.ts`)
- Embedded plugin (same pattern as `background-agents.ts`) installed during `sync`
- On OpenCode startup, fetches provider list via in-process SDK client (`input.client.provider.list()`)
- Extracts variant **keys** per model, writes `~/.cache/gentle-ai/model-variants.json`
- Fire-and-forget pattern (does not block OpenCode init)

### 2. Go enrichment layer (`internal/opencode/models.go`)
- `Variants []string` field on `Model` struct (`json:"-"` — not from cache, set by enrichment)
- `EnrichWithVariants()` merges plugin cache into provider data
- `EffortLevels()` method returns available levels per model (nil = skip picker)

### 3. TUI effort picker (`internal/tui/screens/model_picker.go`)
- New `ModeEffortSelect` sub-mode after model selection
- Shows per-model effort options with "default" prepended
- `applyAssignment()` extracted for reuse across reasoning/non-reasoning paths
- Phase list displays `[effort]` annotation next to assigned models

### 4. Config injection (`internal/components/sdd/inject.go`)
- Writes `"variant": "<level>"` into agent definitions in `opencode.json`
- Cleans up stale `variant` keys when switching to non-reasoning models

### 5. Persistence (`internal/state/state.go`, `internal/app/app.go`)
- `Effort` field on `ModelAssignmentState` with `omitempty` for backward compatibility
- Round-trips through `state.json` correctly

### 6. Profile support (`internal/components/sdd/profiles.go`)
- `extractModelFromAgent()` reads `"variant"` field
- `GenerateProfileOverlay()` writes `"variant"` into profile overlays

## Alternatives Considered

- **Static effort list**: Hardcode `["low", "medium", "high"]` for all models. Rejected — different providers have different levels.
- **Direct API call at TUI init**: Call OpenCode's HTTP API. Rejected — requires `opencode serve` to be active.
- **OpenCode Go SDK**: `client.App.Providers()`. Rejected — SDK's Model struct lacks `Variants` field.
- **GitHub source parsing**: Read `transform.ts` from the OpenCode repo. Rejected — fragile, version-dependent.

### Files changed (implementation ready for PR)
- `internal/assets/opencode/plugins/model-variants.ts` (new — embedded plugin)
- `internal/opencode/models.go` — Variants field, enrichment, effort levels
- `internal/tui/screens/model_picker.go` — effort picker UI
- `internal/components/sdd/inject.go` — variant injection + plugin install
- `internal/components/sdd/profiles.go` — variant read/write in profiles
- `internal/components/sdd/read_assignments.go` — variant read-back
- `internal/model/model_assignment.go` — Effort field
- `internal/state/state.go` — persistence
- `internal/app/app.go` — persistence wiring
- Full test coverage across all packages (~30 new test cases)

### Key discovery
The OpenCode config field is `"variant"` (not `"reasoningEffort"`). OpenCode maps variant names to provider-specific parameters internally. See `packages/opencode/src/config/agent.ts` in the OpenCode repo.

andresnator:
## 🔗 Linked Issue

Closes #429

Related: #391, #392

---

## 🏷️ PR Type

> Maintainer action required: please apply the `type:feature` label to this PR.

---

## 📝 Summary

Adds per-model variant/effort selection to the TUI model picker. Effort levels are no longer hardcoded — they come from OpenCode's runtime provider data via an embedded plugin (`model-variants.ts`), so each model shows only the levels it actually supports (e.g., OpenAI: `none`/`low`/`medium`/`high`/`xhigh`, Anthropic: `low`/`medium`/`high`/`max`).

Writes use the correct OpenCode config field `"variant"` (not `"reasoningEffort"`), which OpenCode maps to provider-specific parameters internally.

---

## 📂 Changes

|-------------|-------------|
| `internal/assets/opencode/plugins/model-variants.ts` | New plugin — writes `~/.cache/gentle-ai/model-variants.json` on OpenCode startup (fire-and-forget, same pattern as `background-agents.ts`) |
| `internal/model/` | `Model` struct gains `Variants []string` (json:"-"); `EnrichWithVariants()` merges plugin cache into provider models; `EffortLevels()` exposes per-model levels |
| `internal/tui/picker/model_picker.go` | New `ModeEffortSelect` sub-mode; phase list shows `[effort]` annotation |
| `internal/tui/picker/profile_picker.go` | `applyAssignment` propagates `Effort` across phases |
| `internal/config/inject.go` | Writes `"variant"` to `opencode.json`; uses `""` (not delete) so deep merge overwrites stale values |
| `internal/config/profiles.go` | `extractModelFromAgent` / `GenerateProfileOverlay` read/write `"variant"` |
| `internal/state/` | `Effort` field wired through `state.json` with `omitempty` for backward compat |
| Tests | ~30 new test cases across `internal/model`, `internal/config`, `internal/tui/picker` |

---

## 🧪 Test Plan

**Unit Tests**
```bash
go test ./...
```

**E2E Tests** (Docker required)
```bash
cd e2e && ./docker-test.sh
```

- [x] E2E tests pass (`cd e2e && ./docker-test.sh`) — **3/3 platforms PASSED** (ubuntu, arch, fedora) — 70 tests per platform, 0 failures, 2 skipped (Tier 2/3 require explicit env flags)
- [x] Manually tested locally

**Manual verification performed:**

- `gentle-ai sync` installs `model-variants.ts` to `~/.config/opencode/plugins/`
- OpenCode startup generates `~/.cache/gentle-ai/model-variants.json` (84 providers detected)
- TUI effort picker shows correct per-model levels
- `opencode.json` receives `"variant": "<level>"` after sync
- OpenCode TUI reflects selected variant in status bar
- Selecting "default" resets variant (overwrites stale values via `""`)
- Pending: fresh-install behavior (no prior cache file) — effort picker should be gracefully skipped

---

## 🤖 Automated Checks

| Check PR Cognitive Load | ⚠️ | **1127 changed lines (1089/+, 38/-)** — exceeds 400-line budget. Maintainer-applied `size:exception` requested below. |
| Check Issue Reference | ⏳ | `Closes #429` present |
| Check Issue Has `status:approved` | ✅ | #429 has `status:approved` |
| Check PR Has `type:*` Label | ⏳ | Maintainer action required: apply `type:feature` |
| Unit Tests | ✅ | `go test ./...` — all packages pass locally (0 failures); CI will confirm |
| E2E Tests | ✅ | `cd e2e && ./docker-test.sh` — 3/3 platforms passed locally; CI will confirm |

---

## ✅ Contributor Checklist

- [x] PR exceeds 400 changed lines — `size:exception` requested with rationale (see Notes for Reviewers)
- [x] `type:feature` label — maintainer action required
- [x] E2E tests pass (`cd e2e && ./docker-test.sh`)
- [x] Documentation updated where needed (config docs reflect `variant` field)
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org/) format

---

### Why `size:exception` is being requested

The change spans 5 layers that are genuinely coupled and cannot be tested end-to-end in isolation:

1. **Plugin** (`model-variants.ts`) — produces the variant cache
2. **Enrichment** (`EnrichWithVariants`) — consumes the cache into the domain
3. **TUI** (`ModeEffortSelect`) — surfaces variants to the user
4. **Injection** (`opencode.json` writes) — persists the choice
5. **State** (`state.json`) — carries the choice across sessions

Splitting into chained PRs would leave intermediate PRs that are unobservable from the user's perspective and cannot be verified without merging the next one. The 5 layers are also already separated by file/package, which keeps the cognitive cost per file low even if the overall line count is high.

### Where to focus review

- **`internal/config/profiles.go` (`GenerateProfileOverlay`)** vs **`internal/config/inject.go` (`injectModelAssignments`)** — both write the `variant` field but use different strategies for clearing it (omit vs `""`). Worth confirming the intended behavior for named profile overlays when a user clears effort.
- **`internal/assets/opencode/plugins/model-variants.ts`** — the response-shape probing (`data?.all ?? data?.providers ?? Array.isArray(data) ? data : []`) is permissive; `catch {}` silences unknown shapes. Acceptable for fire-and-forget but worth noting for future provider-API changes.
- **`internal/tui/picker/profile_picker.go` (`applyAssignment`)** — the "Set all phases" path also needs to update `AllPhasesModel`; verify the call sites do this consistently.

### Follow-up improvements (not blocking)

The bugs above can be addressed either in this PR (additional commits) or in a follow-up — happy to do whichever you prefer.


andresnator:
Tested in : 

- Custom Profile 

image

- Configuration Models

image

- Start installation 

image

internal/assets/opencode/plugins/model-variants.ts
```diff
@@ -0,0 +1,53 @@
...
+          if (m.variants && Object.keys(m.variants).length > 0) {
+            variants[prov.id] = variants[prov.id] || {}
+            variants[prov.id][modelId] = Object.keys(m.variants).sort()
+          }
+        }
+      }
+
+      if (Object.keys(variants).length === 0) return
+
+      const cacheDir = path.join(homedir(), ".gentle-ai", "cache")
```
andresnator, line 34: @Alan-TheGentleman I have a doubt about this path — not sure if it should stay here (`~/.gentle-ai/cache/`) or move to `~/.config/gentleman/model-variants.json`. I kept it here for backward compatibility with Windows, but heads up: I have not tested on Windows yet.

Alan-TheGentleman (changes requested):
Hey @andresnator, the structure here is solid and I appreciate that you tested manually with screenshots and called out your own concerns in the PR body (variant clearing, response-shape probing, applyAssignment). Test-to-prod ratio is great, the layering is clean, and the path convention is consistent across writer, reader, docs, and tests. Already applied `size:exception` and `type:feature` labels so you don't have to chase those.

On your path question (line 36): stick with `~/.gentle-ai/cache/`. The codebase already standardizes on `~/.gentle-ai/` for everything user-side (`internal/state/state.go:9` for `state.json`, `internal/cli/run.go:257,714` for backups), the tool is `gentle-ai` not `gentleman` so `~/.config/gentleman/` would be wrong-branded, and Windows handles `%USERPROFILE%\.gentle-ai\` identically to Unix. Bonus, your own test in `internal/opencode/models_test.go:364-376` already locks the convention by rejecting `~/.cache/gentle-ai/`. You're good as-is.

I ran a dual adversarial review and found a few things worth tightening before merge. Grouped by impact.

**Must fix before merge:**

- `internal/assets/opencode/plugins/model-variants.ts:36-41`: atomic write. Right now `writeFile` races with `gentle-ai sync` reading the file mid-write, JSON parse fails, and `EnrichWithVariants` silently swallows the error so the picker shows zero variants for the whole session. Write to `model-variants.json.tmp` then `rename()` to the final path.
- `internal/assets/opencode/plugins/model-variants.ts:34`: when `variants` is empty, the function returns without writing, which leaves a stale cache from a previous run alive. Either always write (even `{}`) or delete the file in that branch.
- `internal/assets/opencode/plugins/model-variants.ts:42-44`: the `catch {}` swallows ENOSPC, EACCES, and JSON errors with no signal. Add a `console.error("[model-variants] cache refresh failed:", err)` so people running `opencode` interactively see when variants stop refreshing.
- `internal/components/sdd/profiles.go:280-282 and 317-319`: the variant-clearing inconsistency you flagged in "Where to focus review" is real. Default profile (`inject.go:1574-1578`) writes `variant: ""` to overwrite stale values via deep merge. Named profiles omit the field, so a stale `variant: "high"` persists when a user clears effort on a named profile. Mirror the `inject.go` pattern and write `""` instead of omitting.
- `internal/components/sdd/inject.go:1583` (case 3, rootModelID fallback): sister bug to the one above. When that branch fires, `agentMap["model"]` gets set but `agentMap["variant"]` does not, so a stale variant from a previous sync survives the merge. Add `agentMap["variant"] = ""` in case 3 to stay symmetric with case 1.

**Should fix (preferred but not blocking):**

- `internal/tui/screens/model_picker.go` (`handleEffortNav` ESC branch): ESC currently jumps from `ModeEffortSelect` straight to `ModePhaseList`, skipping `ModeModelSelect`. The rest of the picker uses a one-level-up pattern, so this breaks the navigation contract and forces users to redo provider and model selection. ESC should set `state.Mode = ModeModelSelect`.
- `internal/tui/screens/model_picker.go:288`: the synthetic `"default"` option collides if any future SDK variant key is literally named `"default"`. Rename to `"(provider default)"` or guard against the collision before prepending.
- `internal/assets/opencode/plugins/model-variants.ts:25`: null-guard on `prov` and `prov.models`. One bad provider entry currently kills the entire cache write because the outer `catch` swallows it silently. Wrap the inner provider loop in its own try/catch, or null-check before iterating.
- `internal/tui/screens/model_picker.go:411-422`: when `~/.cache/opencode/models.json` exists but `~/.gentle-ai/cache/model-variants.json` does not, the picker is silently variant-less for first-run users. Surface a hint like "Run OpenCode once after install to enable per-model effort selection" so they are not left wondering why effort is not offered.

Once the must-fix list lands, I'll merge. Push when ready and re-request review.

andresnator:
@Alan-TheGentleman all 5 must-fixes are in. Ready for re-review.

**Commits:**
- `a010291` — `model-variants.ts`: tmp+rename atomic write, always-write (drops the empty-variants early return), `console.error` on both catch sites.
- `8f2799d` — `profiles.go`: orchestrator + sub-agent entries now write `variant: ""` when `Effort` is empty, mirroring `inject.go` case 1.
- `a804ed3` — `inject.go`: case 3 (rootModelID fallback) now sets `agentMap["variant"] = ""` alongside the model.

**Tests:**
- New `TestModelVariantsPluginContract` (asset-side grep) locks the plugin contract: must use `rename`, must serialize variants unconditionally, no parameterless `catch {}`, must log via `console.error`.
- Flipped `TestGenerateProfileOverlay_EmptyEffortOmitsKey` → `_EmptyEffortClearsVariant` to assert the new behavior.
- New `TestInjectModelAssignments_RootModelFallbackClearsVariant` covers case 3 with a stale variant in the overlay.

`go test ./...` is green.
image
image

**On the should-fix items:** if you want, I can open a new issue for the remaining items and link the pending tasks there. This PR is already big enough, so I’d prefer to keep the scope frozen here.

Alan-TheGentleman (changes requested):
Deep review pass keeps this in changes requested.

- `applyAssignment` says it updates `AllPhasesModel`, but it only receives `ModelPickerState` by value and cannot mutate the caller. Either return the updated state from the helper or remove that claim and keep the update in one explicit place.
- Persisted `Effort` can become stale when the plugin cache changes or a user switches to a model that no longer exposes that variant. Please validate stored effort against the selected model's current `EffortLevels()` before writing `variant` back out.
- `ModeEffortSelect` should clear `PendingAssignment` and `SelectedModelEffortLevels` on enter and esc. Otherwise a later state bug can reuse stale selection context.

The size exception is fine, but this PR is touching plugin IO, state persistence, config merge, and TUI navigation at once. These edge cases need to be locked down before merge.

andresnator:
@Alan-TheGentleman quick update after your latest review:

- Fixed the `applyAssignment` state handling so we no longer rely on a by-value helper mutating `AllPhasesModel`.
- Added effort sanitization against the selected model’s current `EffortLevels()` so stale persisted variants do not get written back out.
- Tightened `ModeEffortSelect` state cleanup around enter/escape paths to avoid reusing stale pending assignment or effort context.
- Added/updated regression coverage for those cases, plus follow-up coverage around model-variants plugin path handling, sync/install behavior, and uninstall cleanup for the managed plugin source/cache.

Ready for another pass when you have a moment.

Alan-TheGentleman:
Reopening to trigger CI (no checks reported on the existing run). Re-running validation before deep review.

Issue #346, opened 2026-04-21
First-class Engram data-directory selection and migration in install/TUI flows


tonyblu331:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

### 🔍 Affected Area

TUI (terminal UI)

### 💡 Problem Statement

`gentle-ai` treats Engram as a first-class component in the ecosystem, but there is currently no first-class install/setup UX for choosing where the Engram data store lives.

I'm always frustrated when Engram data ends up growing under my default user-profile location, especially on Windows when `C:` is the constrained system drive.

It's currently not possible to choose the Engram data directory as part of the normal install/setup experience. I can choose where binaries go, but not where the persistent Engram database goes.

Every time I need to move Engram data from the default location to another drive, I have to manually work around it with environment variables, manual file moves, or filesystem tricks. That works, but it does not feel like a polished ecosystem setup flow.

This is mainly a UX/product gap:
- Engram is treated as a first-class component in the ecosystem
- its persistent data location is not treated as first-class during install/setup
- users who want memory off `C:` or in a custom location have to do it manually after the fact

### 📦 Proposed Solution

I'd like `gentle-ai` to offer first-class Engram data-directory selection and migration during install/setup, without changing the MCP surface or requiring agent-specific hacks.

From the user's perspective, this should work in three places:

1. Windows installer / PowerShell install flow
2. macOS/Linux `install.sh` flow
3. `gentle-ai` TUI/setup flow

### User experience

During install or setup, I should be able to:
- keep the default Engram data location
- choose a custom Engram data location
- if existing Engram data is detected, choose whether to:
  - keep the current location
  - start fresh at the new location
  - migrate the existing Engram data to the new location

### Example usage

Windows:

```powershell
install.ps1 -Method binary -EngramDataDir "D:\Engram"
install.ps1 -Method binary -EngramDataDir "D:\Engram" -MigrateExistingEngramData
```

macOS / Linux:
```
./install.sh --method binary --engram-data-dir "$HOME/.local/share/engram"
./install.sh --method binary --engram-data-dir "/Volumes/Data/Engram" --migrate-existing-engram-data
```

TUI / setup flow:

Add an Engram settings step or screen where the user can view/change the Engram data location
Offer migration when existing data is found
Expected behavior
No MCP changes
[...]

PR #546, opened 2026-05-16
feat(engram): data directory types, location suggestions, and UI helpers

tonyblu331:
Introduces the domain types needed to reason about engram data directories across the codebase.

`DataDirRef` is a typed string wrapper for a data directory path. It knows how to resolve to an absolute path and fall back to the platform default (`~/.engram`) when empty. Using a distinct type instead of a plain string prevents callers from accidentally mixing up the data dir with other paths.

`Location` represents a candidate storage location with its path, available space, and a human-readable label. `SuggestLocations` enumerates viable alternatives: the current directory, the platform default, and any mounted external volumes (macOS `/Volumes/`, Linux `/mnt/` and `/media/`).

`FormatDirLine` and `FormatSpaceWarning` produce the strings shown in the location picker. They live in a presenter file with no Bubbletea dependency so they can be tested without a terminal.

Closes #346

### DIFF
--- a/internal/components/engram/env.go
+++ b/internal/components/engram/env.go
@@ -1,0 +1,32 @@
+package engram
+
+import (
+	"path/filepath"
+)
+
+// DataDirRef is the persisted reference to an Engram data directory.
+// Currently resolves to a local filesystem path.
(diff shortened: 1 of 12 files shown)


tonyblu331:

PR #547, opened 2026-05-16
feat(engram): atomic database copy with size verification

tonyblu331:
Copying an engram database has two failure modes that need explicit handling: a partial write leaving the destination corrupted, and a size mismatch that would be silently accepted by a naive copy.

`CopyDB(src, dst string) error` writes to a `.tmp` file on the same volume first, verifies the byte count matches the source, then renames atomically. On failure at any point the temp file is removed. The rename is atomic on the same volume; cross-volume moves go through the same copy-verify-rename path.

`io.Copy` between two `*os.File` values triggers `sendfile(2)` on Linux and the equivalent on macOS and Windows automatically via `(*os.File).ReadFrom` — no manual buffer allocation needed.

Closes #346

### DIFF
--- a/internal/components/engram/copy.go
+++ b/internal/components/engram/copy.go
@@ -1,0 +1,72 @@
+package engram
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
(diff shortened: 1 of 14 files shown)


tonyblu331:

PR #548, opened 2026-05-16
feat(engram): data directory service and injection options

tonyblu331:
Orchestrates the copy, move, and delete operations on an engram data directory, and extends the MCP injection path to carry a custom data directory into agent configs.

`DataDirService` wraps each mutating operation with a snapshot before writing. Copy and move check available disk space upfront. Delete clears all files under the directory, accumulating any removal errors with `errors.Join` rather than stopping at the first failure.

`InjectWithOptions` extends the existing `Inject` function with an `InjectOptions` struct. When `InjectOptions.DataDir` is non-empty, the `ENGRAM_DATA_DIR` environment variable is included in the generated MCP server config. `Inject` becomes a one-line wrapper with zero-value options, so existing call sites are unaffected.

`datadir_ops.go` adds `DataDirHasContent`, `DataDirSize`, and `DiskSpaceOKForDataDir` — storage-agnostic queries that work on any directory, not just SQLite databases.

Closes #346

### DIFF
--- a/internal/components/engram/copy.go
+++ b/internal/components/engram/copy.go
@@ -1,0 +1,72 @@
+package engram
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+// CopyDB copies the Engram SQLite database from src to dst.
+//
+// Strategy: write to a temp file (dst+".tmp") first, verify the byte count,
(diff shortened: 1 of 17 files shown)


tonyblu331:

PR #549, opened 2026-05-16
feat(model,state): persist engram data directory selection

tonyblu331:
The TUI needs to carry the user's data directory choice across multiple screens, and the install state needs to remember it across runs.

Adds `EngramDataDirOp` to `model/selection.go` — a string enum (keep / copy / move / delete / fresh) following the same pattern as other operation types in the file. Adds `EngramDataDir DataDirRef` to `Selection` to hold the resolved path during a session.

Adds `EngramDataDir string` to `state.InstallState` with `json:"engram_data_dir,omitempty"` so the chosen directory survives between gentle-ai invocations. The field is intentionally `omitempty` — users who never changed the default produce no extra JSON.

Closes #346

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -103,7 +103,15 @@ func RunArgs(args []string, stdout io.Writer) error {
 			return fmt.Errorf("resolve user home directory: %w", err)
 		}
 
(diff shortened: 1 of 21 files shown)


tonyblu331:

PR #551, opened 2026-05-16
feat(tui): wire engram data directory screens into model and router

tonyblu331:
Connects the render functions and domain logic to the running TUI.

Adds four screen constants to the existing iota list in `model.go` — two for the management flow reachable from Welcome, two for the install-time flow between Preset and DependencyTree. Router entries follow the same pattern as the backup management screens.

`app.go` pre-loads `EngramDataDir` from persisted state before constructing the TUI model, so the management screen shows the current directory on first render without an async round-trip.

`WelcomeOptions` gains an `engram bool` parameter. When true, a "Manage Engram directory" option appears immediately before "Manage backups", matching the visual grouping users expect.

Closes #346

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -103,7 +103,15 @@ func RunArgs(args []string, stdout io.Writer) error {
 			return fmt.Errorf("resolve user home directory: %w", err)
 		}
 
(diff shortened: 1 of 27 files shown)


tonyblu331:

PR #552, opened 2026-05-16
fix(engram): remove SQLite assumptions from data directory layer

tonyblu331:
The previous implementation assumed the data directory contained a specific SQLite file layout. This prevented the copy, move, and delete operations from working correctly when the layout differed, and made the code fragile against future storage backends.

Removes `sqlite_artifacts.go` and replaces all SQLite-specific knowledge with storage-agnostic operations from `datadir_ops.go`. `clearDataDir` now removes all files under a directory rather than targeting specific filenames.

Also fixes two issues found during review: `InjectWithOptions` now returns an error when `DataDir` is a relative path (MCP environment variables require absolute paths), and the VSCode SDD golden file is updated to match the current inject output.

Closes #346

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# TUI snapshot goldens must compare stable across OS checkouts (LF).
+internal/tui/testdata/*.golden text eol=lf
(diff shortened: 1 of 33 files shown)


tonyblu331:

PR #555, opened 2026-05-16
fix(engram): update only MCP config when switching active data directory

tonyblu331:
When a user switches the active engram data directory, the MCP server config needs to be updated to point at the new location. But the workspace prompt files (CLAUDE.md, AGENTS.md) should not be touched — they contain agent instructions, not data directory config, and rewriting them on a directory switch would clobber any local customizations.

Adds `InjectMCPWithOptions`, which updates only the MCP server config and explicitly skips prompt file writes. The "Set Active" operation uses this instead of the full `Inject`. Adds a test that verifies the AGENTS.md prompt file is not written during a Set Active operation.

Closes #346


tonyblu331:

PR #561, opened 2026-05-16
feat(tui): custom path entry and disk space check for data directory screens

tonyblu331:
Two gaps in the data directory management flow: users had no way to enter a path not in the suggestion list, and the UI would proceed to the confirm screen even when the destination volume lacked enough space for the copy.

Adds `ScreenEngramDataDirCustomPath` where users type a destination path directly. The path is cleaned and validated before proceeding.

`checkEngramCopyMoveDiskSpace` runs when the user selects a destination for a copy or move. If the destination volume doesn't have enough free space, an inline error is shown on the location picker screen and the user can choose a different destination. The check calls `DiskSpaceOKForDataDir` from the engram package.

Closes #346

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# TUI snapshot goldens must compare stable across OS checkouts (LF).
+internal/tui/testdata/*.golden text eol=lf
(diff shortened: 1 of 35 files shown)


tonyblu331:

PR #562, opened 2026-05-16
feat(engram): active data directory selection and history

tonyblu331:
Lets users switch which directory is the active engram data store from within gentle-ai, without requiring a full copy or move.

`EngramDataDirOpSet` updates the persisted `EngramDataDir` in state and syncs the MCP environment config to point at the new location. Previously-used directories are tracked in `EngramKnownDataDirs` so the "Set Active" screen can offer them as choices without requiring the user to type paths again.

`EngramKnownDataDirs` is persisted to state and pre-loaded into the TUI model on startup.

Closes #346


tonyblu331:

PR #563, opened 2026-05-16
fix(app): return early from copy operation before updating active directory

tonyblu331:
`buildEngramDataDirFn` had a control flow bug: the copy operation fell through to the state update block, which would have overwritten the active data directory path with an empty string after a successful copy. The copy operation should leave the active directory unchanged.

Adds a `default` case that returns after a copy so the active directory is not modified. Adds integration tests for the copy and move operations that verify the state before and after each.

Closes #346


tonyblu331:

PR #564, opened 2026-05-16
fix(engram,storage): defensive fixes for data directory copy and location listing

tonyblu331:
Several small gaps found during review.

`service.go` now rejects copy and move operations where the destination is the same as the source or is a subdirectory of it. Previously these would proceed and either overwrite in place or recurse into themselves.

`storage.AvailableBytes` previously failed when called with a path that does not exist yet — which is the normal case when pre-checking space for a destination that will be created by the copy. It now walks up to the nearest existing ancestor directory to query the volume, then returns that volume's available bytes.

`copy.go` was silently discarding errors from `os.Remove` in error cleanup paths. If the temp file could not be removed, the caller had no way to know. Cleanup errors are now joined onto the primary error with `errors.Join`.

`locations_windows.go` package-level variables used leading underscores (`_kernel32`, `_getDiskFreeSpaceExW`), which is a C convention for unexported symbols and has no meaning in Go. Renamed to `kernel32DLL` and `getDiskFreeSpaceExW`.

`locations_unix.go` had no upper bound on the number of volumes returned by `platformVolumes`. Added a cap of 20 to prevent the location picker from becoming unusable on systems with many mount points.

Closes #346


tonyblu331:

PR #565, opened 2026-05-16
feat(tui,engram): install-time data directory chooser

tonyblu331:
When engram is selected during installation, the app previously used the default `~/.engram` location without giving the user a choice. This matters most on systems where the home directory is on a small or slow volume.

`ScreenEngramDataDirInstall` now shows a location picker listing the platform default alongside any mounted external volumes. If an existing engram data directory is detected, a different view is shown that lets the user decide what to do with it (keep, move, or start fresh) rather than silently layering on top.

`Location` gains a `HasData bool` field, set by `DataDirHasContent`, so the picker can display which locations already have engram data without knowing anything about the file format.

The confirm and result screens gain a `backupRoot` parameter so the displayed backup path reflects the actual location rather than a placeholder.

The golden test for the minimal preset flow without OpenCode is replaced by static content assertions. The original golden contained real disk volume paths and free-space values from the machine that generated it, making it fail on any other machine.

Closes #346


tonyblu331:

Issue #230, opened 2026-04-04
fix(tui): upgrade shows ✗ failure marker but binary is correctly updated


aleka:
### Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs status:approved before a PR can be opened

### Bug Description
When upgrading gentle-ai via the TUI (Upgrade tools or Upgrade + Sync), the upgrade step for the gentle-ai binary shows a red ✗ failure marker even though the binary is successfully downloaded and replaced. The version is confirmed correct after the TUI exits.

### Steps to Reproduce
1. Run `gentle-ai` to open the TUI
2. Select "Upgrade tools" or "Upgrade + Sync"
3. Wait for the upgrade to complete
4. Observe: "Upgrading gentle-ai via binary (X.Y.Z → A.B.C)" shows ✗
5. Run `gentle-ai --version` — shows the new version A.B.C
6. Check binary timestamp (`ls -la /home/aleka/go/bin/gentle-ai`) — matches upgrade time

### Expected Behavior
The TUI should show ✓ (green checkmark) when the binary upgrade succeeds, matching the actual outcome.

### Actual Behavior
The TUI shows ✗ (red cross) even when the binary was correctly replaced. The backup step shows ✓ correctly, but the upgrade step always shows ✗.

### Operating System
Linux (Ubuntu)

### Agent / Client
OpenCode

### Shell
bash

### Relevant Logs
```
gentle-ai
  ✓ Creating pre-upgrade backup
  ✗ Upgrading gentle-ai via binary (1.18.1 → 1.18.2)
```
After exiting TUI:
```
$ gentle-ai --version
gentle-ai 1.18.2
$ ls -la /home/aleka/go/bin/gentle-ai
-rwxr-xr-x 1 aleka aleka 9310392 Apr  4 12:22 /home/aleka/go/bin/gentle-ai
```

### Additional Context
- Observed consistently across multiple upgrades: v1.16.1→v1.18.1, v1.18.1→v1.18.2
- The sync step that follows the upgrade works correctly and reports accurate file counts
- The backup step also reports ✓ correctly
- Only the "Upgrading gentle-ai via binary" step shows the false ✗
- Binary is installed via direct download (not brew), path: /home/aleka/go/bin/gentle-ai

Alan-TheGentleman:
[...]

PR #706, opened 2026-05-28
fix(self-update): skip auto-upgrade on TUI entry

### LINKED ISSUE #696: bug(self-update): auto-upgrade triggers on TUI entry but output appears on exit when reExec fails
### Pre-flight Checklist
- [x] I understand that PRs will be rejected if the linked issue does not have status:approved

### Bug Description
When gentle-ai is executed without arguments (TUI flow), self-update runs **before** launching the interface. If a newer version is available, the binary is replaced on disk and then attempts syscall.Exec() to restart the process.

**The problem**: if syscall.Exec() fails silently (permissions, corrupt path, or any other reason), the old process continues running and enters the TUI. The user sees no indicator that an upgrade occurred. Upon exiting the TUI with q, the buffered self-update output suddenly appears on screen, showing something like:

```
Updated gentle-ai via binary 1.31.0 → 1.32.0
```

This is confusing and violates the principle of least surprise: a program should not auto-modify its own binary without explicit consent, especially when the user did not request an upgrade.

### Steps to Reproduce
1. Install gentle-ai v1.31.0 (or any version with a pending update)
2. Run gentle-ai without arguments (enters TUI)
3. Observe that inside the TUI it shows "gentle-ai ✓ up to date 1.32.0" (the binary on disk was already replaced)
4. Press q to exit
5. **Actual**: The upgrade message appears in the terminal after exiting the TUI
6. **Expected**: Either it should not auto-upgrade, or it should clearly inform the user before entering the TUI

### Expected Behavior
One of these options:
1. **No auto-upgrade on TUI entry**: Only notify that an update is available, let the user explicitly run gentle-ai upgrade
2. **If auto-upgrading**: Show a clear message BEFORE entering the TUI indicating it updated and that a manual restart is needed if re-exec fails
3. **Explicit confirmation**: Ask the user if they want to upgrade before doing it

### Actual Behavior
- The binary is replaced on disk without user consent
- The old process keeps running (TUI opens with old version in memory)
- Upon exiting the TUI, buffered upgrade output appears
- The user has no way of knowing an upgrade occurred until they exit

### Environment
**Gentle AI Version:** 1.31.0 → 1.32.0
**Operating System:** Linux (Ubuntu/Debian)
**AI Agent / Client:** Other
**Affected Area:** TUI (terminal UI)

### Logs / Error Output
```
Updated gentle-ai via binary 1.31.0 → 1.32.0
```

#### Relevant code
In internal/app/app.go:91-95:
```go
if !isExplicitUpdateFlow(args) {
    if err := selfUpdateFn(context.Background(), Version, resolveProfile(), stdout); err != nil {
        _, _ = fmt.Fprintf(stdout, "Warning: self-update failed: %v\n", err)
    }
}
```

In internal/app/selfupdate.go:134:
```go
return reExec(executable, os.Args, os.Environ())
```

#### Root cause hypothesis
reExec() fails but returns an error, which is caught by app.go. Since the error is non-fatal, the program continues to the TUI. The upgrade output (including spinner and success message) stays in the stdout buffer, but because bubbletea uses an alternate screen buffer, it is not displayed until the normal screen is restored on exit.

#### Related issues
- #541: Auto-upgrade on Windows (different platform, same general topic)
- #359: Verify release integrity before replacing binary

#### Possible Fix
Options (not mutually exclusive):
1. **Lazy self-update**: Do not run selfUpdateFn() automatically on TUI entry. Only show an update available notification inside the TUI.
2. **Graceful degradation**: If selfUpdate() succeeds but reExec() fails, show a prominent message before entering the TUI:
   ```
   gentle-ai was updated to v1.32.0 but could not restart automatically.
   Please run 'gentle-ai' again to use the new version.
   Press Enter to continue with the current version...
   ```
3. **Respect GENTLE_AI_CONFIRM_UPDATE**: The GENTLE_AI_CONFIRM_UPDATE variable (introduced in v1.31.0) seems to apply only to the upgrade command, not the pre-TUI self-update. Extend it to the self-update path.
4. **Skip self-update on TUI entry**: Move self-update to an explicit command or a menu option within the TUI, rather than running it automatically on launch.

aleka:
Fixes #696

## Summary
- Skip auto-upgrade when gentle-ai is launched without arguments (TUI entry point)
- Prevents silent binary replacement and confusing post-TUI upgrade messages
- TUI already has its own background update check and notifies users via "Upgrade tools ★"

## Changes
| File | Change |
|------|--------|
| `internal/app/app.go` | Skip self-update on TUI entry (len(args) == 0) with inline comment referencing #696 |
| `internal/app/app_test.go` | Add `TestRunArgs_TUISkipsSelfUpdate` with dual validation (selfUpdate skipped + TUI called) |

## Test Plan
- [x] `go build ./...` passes
- [x] `go vet ./internal/app/...` passes
- [x] `go test ./internal/app/... -run "TestRunArgs_" -v` passes (including new test)
- [x] `go test ./internal/app/... -run "TestSelfUpdate_" -v` passes (all existing tests)
- [x] `go test ./internal/app/... -race -count=1` passes (no data races)

## Note on flag-only invocations
`--help`, `--version` and similar flag-only invocations retain the existing self-update behavior. Only the zero-arg TUI entry point is fixed.

## Contributor Checklist
- [x] Linked an approved issue (Fixes #696)
- [x] Ran build and vet
- [x] Tests pass with race detector

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -95,7 +95,10 @@ func RunArgs(args []string, stdout io.Writer) error {
 
 	// Self-update: check for a newer gentle-ai release and apply it before
 	// CLI/TUI dispatch. Errors are non-fatal — logged and swallowed.
-	if !isExplicitUpdateFlow(args) {
+	// Skip auto-upgrade on TUI entry (len(args) == 0) to avoid silently
+	// replacing the binary while the user expects a clean TUI launch (#696).
+	isTUIFlow := len(args) == 0
+	if !isTUIFlow && !isExplicitUpdateFlow(args) {
 		if err := selfUpdateFn(context.Background(), Version, resolveProfile(), stdout); err != nil {
 			_, _ = fmt.Fprintf(stdout, "Warning: self-update failed: %v\n", err)
 		}
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	tea "github.com/charmbracelet/bubbletea"
 	"github.com/gentleman-programming/gentle-ai/internal/backup"
 	"github.com/gentleman-programming/gentle-ai/internal/model"
 	"github.com/gentleman-programming/gentle-ai/internal/state"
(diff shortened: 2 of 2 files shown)


Alan-TheGentleman (approved):
Thanks @aleka — this is a clean, well-scoped fix that goes after the actual root cause.

By skipping `selfUpdate` on zero-arg TUI entry there's no binary swap, no `GENTLE_AI_SELF_UPDATE_DONE`, and crucially no `reExec` — so the buffered "Updated..." output can no longer hide behind the alt screen and surface on exit.

I verified the CLI paths stay intact: `upgrade`/`update` are still guarded by `isExplicitUpdateFlow`, and `install`/`sync`/`doctor` keep their existing self-update behavior (no alt screen, so output ordering isn't a problem). This also matches the maintainer-approved option in the issue, and the removed auto-upgrade-on-launch is redundant with the TUI's own "Upgrade tools ★" check — no feature regression. Tests, `go vet`, and `-race` all pass, and the new `TestRunArgs_TUISkipsSelfUpdate` asserts both invariants (self-update skipped + TUI called once).

Two non-blocking notes:
1. The `len(args) == 0` check is now computed twice (`app.go:98` as `isTUIFlow` and again at `app.go:104` for dispatch) — consider hoisting a single `isTUIFlow` var and reusing it to avoid future drift.
2. Separate from this PR: the underlying "reExec failure is silently swallowed" behavior still exists for CLI subcommands. Given #230/#201 touch this same fragile area, a short follow-up to make reExec failures more visible would be worthwhile.

Approving. 🙏

Issue #713, opened 2026-05-29
feat(skills): workflow-governance skill + Workflow agent() governance bridge for ultracode


orisepu:
### Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs status:approved before a PR can be opened

### Problem Description

Claude Code's **ultracode / xhigh** effort level enables *dynamic workflow orchestration*: the orchestrator authors a `Workflow` whose `agent()` / `parallel()` / `pipeline()` calls fan out sub-agents. That `agent()` surface is **separate from the Agent/Task tool**, and it inherits **none** of Gentle AI's governance automatically:

- The **Model Assignments** gate is textually scoped to *"every Agent tool call"* — it never fires for `agent()`. Every Workflow sub-agent silently inherits the session model (**Opus under xhigh**), inverting the cost tiering (haiku/sonnet).
- **Skill injection** (`## Skills to load before work`), **Strict TDD forwarding**, and the **engram Sub-Agent Context Protocol** are all delivered as Agent-tool prompt scaffolding — none reaches a Workflow `agent()` call.

Net effect: under ultracode, even a competent agent authors a workflow that runs ungoverned — most-expensive model, no skills, no engram, no TDD gate — and nothing in the config surfaces it.

### Proposed Solution

Two chained PRs.

**PR #1 — `workflow-governance` skill** (`internal/assets/skills/workflow-governance/`):
- `SKILL.md` — authoring contract: when ultracode is ON and you author a Workflow that does real work, route every `agent()` through the helper.
- `assets/governance-preamble.js` — copy-paste helper (`MODEL_BY_ROLE`, `AGENTTYPE_BY_ROLE`, `SKILL_PATHS` populated at author-time from `.atl/skill-registry.md`, `buildGoverned`, `assertGoverned`, `launchSubagent`). Reproduces model/agentType/skill/engram/TDD on the Workflow surface and **asserts them in deterministic JS — failing loudly** on a missing control. Registered in `foundationSkills`.

**PR #2 — orchestrator bridge** (`internal/assets/claude/sdd-orchestrator.md`):
- A `#### Workflow agent() Governance (MANDATORY)` clause that makes loading the skill mandatory before authoring a Workflow, generalizing the model gate / skill injection / strict-TDD / engram protocol from *"every Agent tool call"* to **"every delegation — Agent tool OR Workflow `agent()`"**.
[...]

PR #714, opened 2026-05-29
feat(skills): add workflow-governance skill

orisepu:
Part of #713 — the **skill** half of the workflow-governance feature.

Adds the `workflow-governance` skill so gentle-ai ships it to `~/.claude/skills/` for ecosystem-only and full presets:

- `internal/assets/skills/workflow-governance/SKILL.md` — authoring contract for Workflow governance under ultracode.
- `assets/governance-preamble.js` — copy-paste helper (`MODEL_BY_ROLE`, `AGENTTYPE_BY_ROLE`, `SKILL_PATHS` populated at author-time from `.atl/skill-registry.md`, `buildGoverned`, `assertGoverned`, `launchSubagent`) that carries the Agent-tool controls onto the Workflow `agent()` surface and **asserts them in deterministic JS — failing loudly** on a missing control.
- Registered: `model/types.go`, `catalog/skills.go`, `presets.go` (foundationSkills); counts/goldens updated (`assets_test.go`, `e2e/e2e_test.sh`, `skills-presets.json`).

`go test ./...` green. ~170 lines / 8 files. PR #2 (the mandatory orchestrator clause) stacks on this branch.


orisepu:
Force-pushed: broadened the activation trigger beyond ultracode, plus a small cleanup.

**Why:** you can author a Workflow via the `workflow` keyword opt-in without ultracode/xhigh being active, and the governance must apply there too. The original trigger only fired "when ultracode/xhigh is active", which missed that path (a keyword-triggered Workflow would run ungoverned, the exact gap this skill closes).

This PR broadens the three spots it owns: the frontmatter `description`, the Activation Contract, and the `governance-preamble.js` header comment, so it fires whenever you author a Workflow that does real work (ultracode OR keyword opt-in). The orchestrator clause is broadened to match in #715. Also dropped a dead `!opts.model` check in `assertGoverned` (the model fallback made it unreachable).

`go test ./...` green; the description is still a quoted single line under the 160-char limit.

Alan-TheGentleman:
This can be reviewed, but it needs governance cleanup first.

Right now the PR says it is part of #713, but it does not close an approved issue and it has no `type:*` label. Please link an approved issue with `Closes #...`, add exactly one PR type label, and keep this PR limited to the skill slice.

Alan-TheGentleman:
Thanks for the contribution. Closing this because it does not link an approved issue, so there is no agreed scope to review against.

A workflow-governance skill can be discussed, but it needs an approved issue first so we can agree on scope, trigger rules, and where it belongs in the skill system.

PR #715, opened 2026-05-29
docs(orchestrator): bridge governance to Workflow agent() surface

orisepu:
Part of #713 — the **orchestrator bridge** half. **Stacked on #714** (review after it merges).

Adds a `#### Workflow agent() Governance (MANDATORY)` clause to `internal/assets/claude/sdd-orchestrator.md` that makes loading the `workflow-governance` skill mandatory before authoring a Workflow, and generalizes the model gate / skill injection / strict-TDD / engram protocol from *"every Agent tool call"* to **"every delegation — Agent tool OR Workflow `agent()`"**.

**Claude-only** by design: the Workflow tool is Claude-Code-exclusive, so the clause lives in the `claude/` variant (not `generic/`). 2 claude `claudemd` goldens regenerated.

> Note: this branch is stacked on #714, so the diff currently shows its 2 commits too. The net-new change here is the single `docs(orchestrator)` commit (sdd-orchestrator.md + 2 goldens, ~45 lines). Rebases cleanly onto `main` once #714 merges.

Why it's needed: with only the skill (#714), a fresh ultracode session does **not** reach for it — it authors a workflow with bare `agent()` (all Opus, no governance). This clause is what flips passive availability into mandatory application (verified end-to-end; see #713).

`go test ./...` green on the stacked state.


orisepu:
Force-pushed: rebased onto the updated #714, plus the same trigger broadening in the `#### Workflow agent() Governance (MANDATORY)` clause and regenerated the two claude goldens.

The clause now fires for any Workflow doing real work, whether under ultracode/xhigh or when the Workflow tool is opted into via the `workflow` keyword, matching the skill in #714. The earlier commits belong to #714; the net-new here is still the single `docs(orchestrator)` commit.

Alan-TheGentleman:
This should wait behind #714, and it also needs governance cleanup before review.\n\nPlease link an approved issue with , add exactly one  label, and make the stack relationship explicit in the body after #714 is ready. The orchestration bridge should stay as the second slice, not mixed with the skill implementation.

Alan-TheGentleman:
This should wait behind #714, and it also needs governance cleanup before review.

Please link an approved issue with `Closes #...`, add exactly one `type:*` label, and make the stack relationship explicit in the body after #714 is ready. The orchestration bridge should stay as the second slice, not mixed with the skill implementation.

Alan-TheGentleman:
Thanks for the contribution. Closing this because it does not link an approved issue, so there is no agreed scope to review against.

If this workflow governance bridge is still relevant, please open a focused issue first. Once the issue is triaged and approved, a new PR linked to that scope is welcome.

PR #734, opened 2026-06-01
fix(filemerge): follow symlinked parent directories instead of refusing

### LINKED ISSUE #733: fix(filemerge): allow writing through symlinked parent directories
## Pre-flight Checklist

## 🔍 Affected Area

Installation Pipeline

## 🐛 Bug Description

The SDD component fails to install when a target directory in `~/.claude/` is a symlink. The error is:

```
FAILED: component:sdd — inject sdd for "claude-code": refusing symlink parent directory "~/.claude/agents" for "~/.claude/agents/sdd-orchestrator.md"
```

This blocks a full installation for users who manage their Claude Code config via symlinks (e.g. a shared dotfiles or team config repo).

## 🔍 Root Cause

`internal/components/filemerge/writer.go:152` — `ensureAtomicParentDir` uses `os.Lstat` to detect symlinks and refuses outright:

```go
if info.Mode()&os.ModeSymlink != 0 {
    return fmt.Errorf("refusing symlink parent directory %q for %q", dir, path)
}
```

This rejects legitimate setups where a config subdirectory (e.g. `~/.claude/agents`) is a symlink pointing to a real directory in a dotfiles or team config repository.

## 📦 Proposed Solution

When a symlink is detected, resolve it with `filepath.EvalSymlinks` and verify the target is a valid directory, then proceed normally. The OS handles writes through symlinks transparently — no security risk is introduced because the resolved target path is verified.

```go
if info.Mode()&os.ModeSymlink != 0 {
    resolved, err := filepath.EvalSymlinks(dir)
    if err != nil {
        return fmt.Errorf("resolving symlink parent %q for %q: %w", dir, path, err)
    }
    info, err = os.Stat(resolved)
    if err != nil {
        return fmt.Errorf("stat symlink target %q for %q: %w", resolved, path, err)
    }
    dir = resolved
}
```

The remaining checks (directory assertion, write permission) continue to run against the resolved target.

## 🔄 Alternatives Considered

- **Remove the check entirely**: too permissive — the original intent of rejecting unexpected redirects is valid.
- **Add a config flag to allow symlinks**: unnecessary complexity for a standard filesystem pattern.

## 📎 Additional Context

Reproduction: symlink `~/.claude/agents` to any real directory, then run `gga` with SDD selected. The `component:sdd` step fails immediately.

I am happy to implement this once the issue is approved.

Closes (related to the setup described in #730 — both affect users with shared Claude Code configs)

facchin21:
## Summary

- `ensureAtomicParentDir` previously returned an error when the parent directory was a symlink, blocking installations for users whose Claude Code config subdirectories (e.g. `~/.claude/agents`) are managed as symlinks to a dotfiles or team config repository
- Fix: when a symlink is detected, resolve it with `filepath.EvalSymlinks`, stat the real target, and continue the existing directory and permission checks against the resolved path
- The write itself proceeds through the symlink transparently (OS-native behavior); no security regression — the resolved target is still verified to be a writable directory

## Test plan

- [ ] `go test ./internal/components/filemerge/...` passes
- [ ] `go test ./...` passes
- [ ] Run `gga` with `~/.claude/agents` symlinked to a real directory: `component:sdd` installs successfully

Closes #733

### DIFF
--- a/internal/components/filemerge/writer.go
+++ b/internal/components/filemerge/writer.go
@@ -150,7 +150,17 @@ func ensureAtomicParentDir(dir, path string) error {
 		return fmt.Errorf("stat parent directory for %q: %w", path, err)
 	}
 	if info.Mode()&os.ModeSymlink != 0 {
-		return fmt.Errorf("refusing symlink parent directory %q for %q", dir, path)
+		// Parent is a symlink (e.g. ~/.claude/agents → dotfiles repo).
+		// Resolve the target and continue checks against the real directory.
+		resolved, err := filepath.EvalSymlinks(dir)
+		if err != nil {
+			return fmt.Errorf("resolving symlink parent %q for %q: %w", dir, path, err)
+		}
+		info, err = os.Stat(resolved)
+		if err != nil {
+			return fmt.Errorf("stat symlink target %q for %q: %w", resolved, path, err)
+		}
+		dir = resolved
 	}
 	if !info.IsDir() {
 		return fmt.Errorf("parent path %q for %q is not a directory", dir, path)
--- a/internal/components/filemerge/writer_test.go
+++ b/internal/components/filemerge/writer_test.go
@@ -1,6 +1,7 @@
 package filemerge
 
 import (
+	"bytes"
 	"errors"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


Alan-TheGentleman (approved):
Scoped, tested, linked to the approved issue. The symlink boundary stays safe because parent symlinks are resolved while final-file symlink rejection remains intact. Approving.

PR #747, opened 2026-06-03
feat(system): add android support and improve temp dir handling

pgtoopx:
## Context
Currently, Gentle AI assumes standard Unix paths (like `/tmp`) and doesn't recognize `android` as a supported OS, making it impossible to run on environments like Termux.

## Changes
- **OS Support**: Added `android` to the supported OS list in `internal/system/detect.go`.
- **Package Manager**: Mapped `android` to use `pkg` (Termux default) in the platform resolver.
- **Portability**: Refactored fixed `/tmp` paths to use `os.TempDir()` in `internal/installcmd/resolver.go`.

## Impact & Compatibility
- **Zero breaking changes**: All additions are scoped to the new `android` detection.
- **Improved Robustness**: Switching to `os.TempDir()` benefits all systems by respecting environment-specific temporary directories instead of hardcoding `/tmp`.

## Verification
- Verified on Android via Termux.
- Code remains idiomatic and follows existing patterns.


Alan-TheGentleman:
Closing this as superseded by #765.\n\nThe Android/Termux direction is valid, but this branch only covers a small slice. #765 is the candidate that carries the fuller platform support path.

Issue #830, opened 2026-06-11
fix(scripts): BOM removal broke self-update on Windows PowerShell 5.1 — em dash in install.ps1 parses as curly quote


IP-Mattos:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

### 📝 Bug Description

The BOM removal that fixed #783 introduced a new failure on the **other** execution path. `install.ps1` contains a UTF-8 em dash (`—`) inside a double-quoted string (line 69, the banner: `"  Gentle-AI — Ecosystem, Frameworks, Workflows"`). Now that the file has **no BOM**, Windows PowerShell 5.1 reads it as ANSI (cp1252) whenever it is executed **from disk** (`powershell -File`). The em dash bytes `E2 80 94` decode to `â€”` — and `0x94` is `”` (RIGHT DOUBLE QUOTATION MARK), which PowerShell's tokenizer accepts as a string delimiter. The banner string closes early, every quote downstream becomes unbalanced, and the script dies with a wall of parser errors before executing anything.

This is exactly the path `gentle-ai upgrade` uses for self-update on Windows: it downloads `install.ps1` to `%TEMP%\gentle-ai-install-<random>.ps1` and runs it in a new PowerShell window. Result: **self-update is currently broken for every Windows user whose default `powershell` is Windows PowerShell 5.1** (the OS default). The `irm <url> | iex` path is NOT affected because `Invoke-RestMethod` decodes the HTTP body as UTF-8, so the in-memory string is correct — which is why this is easy to miss on machines using `pwsh` 7 (it also reads BOM-less files as UTF-8).

The two bugs are a seesaw: **with BOM**, `irm | iex` breaks (#783). **Without BOM**, `-File` on WinPS 5.1 breaks (this issue). The only fix that works for both paths is making the script pure ASCII.

### 🔄 Steps to Reproduce

1. On Windows 11 with stock Windows PowerShell 5.1, install gentle-ai 1.39.1 while 1.39.2 is the latest release
2. Run `gentle-ai upgrade` (or let any flow trigger the CLI self-update)
3. A new PowerShell window opens running `%TEMP%\gentle-ai-install-<random>.ps1` and fails to parse

Minimal repro without gentle-ai:

```powershell
[...]

PR #834, opened 2026-06-12
fix(install): escape parentheses in PowerShell install script

decode2:
## Problem

On Windows PowerShell 5.1, running the install script fails with parse errors:

```
At C:\Users\...\gentle-ai-install-XXXX.ps1:205 char:78
+ ... Error "Downloaded file is suspiciously small ($fileSize bytes). Archi ...
+                                                             ~~~~~
Unexpected token 'bytes' in expression or statement.
```

## Root Cause

PowerShell interprets parentheses `()` inside double-quoted strings as subexpression syntax. When the content inside parentheses is not a valid expression (e.g., `$fileSize bytes` where `bytes` is not a variable), it causes a parse error.

Affected lines:
- Line 205: `"...suspiciously small ($fileSize bytes)..."`
- Line 207: `"Downloaded $archiveName ($fileSize bytes)"`
- Line 285: `"Added $installDir to your PATH (takes effect in new shells)"`

## Solution

1. Escape parentheses with backticks: `` `(`` and `` `)``
2. Convert file to UTF-8 with BOM (PowerShell preference)

## Testing

- PowerShell parser validates the script without errors
- Tested with minimal reproduction script
- Backticks correctly render literal parentheses in output

## Impact

Fixes installation/update failures on Windows PowerShell 5.1 (the default Windows PowerShell version).

## Summary by CodeRabbit

* **Style**
  * Updated the installation script's message formatting for improved text display in error and success notifications.

* **Chores**
  * Added UTF-8 encoding declaration to the installation script.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,4 +1,4 @@
-#Requires -Version 5.1
+﻿#Requires -Version 5.1
 <#
 .SYNOPSIS
     Gentle-AI — Install Script for Windows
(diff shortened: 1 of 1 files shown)


decode2:
Closing: this fix addresses the wrong root cause. The real issue is #830 - em dashes in install.ps1 are misinterpreted as curly quotes when PowerShell 5.1 reads the file without BOM. The correct fix is to replace em dashes with ASCII hyphens, not to add BOM or escape parentheses. Will submit a new PR with the correct fix.

PR #840, opened 2026-06-12
feat(kimi): add current kimi-code support with ~/.kimi-code paths

### LINKED ISSUE #782: feat(kimi): Add support for kimi-code v0.11+ (Node.js rewrite)
## Problem Statement

Gentle AI currently hardcodes all Kimi paths to `~/.kimi`, which is the legacy Python/uv-based Kimi CLI. The new Node.js-based `kimi-code` v0.11.0 uses `~/.kimi-code/` for configuration, skills, and MCP settings. This means:

- Users with kimi-code v0.11+ installed are not detected by Gentle AI
- Configuration files (KIMI.md, config.toml, mcp.json) are written to the wrong directory
- Skills are installed to the legacy shared path instead of the native v0.11+ path
- Post-install message references `--agent-file`, which does not exist in v0.11+

## Affected Area

Agent Detection, Installation Pipeline

## Proposed Solution

Add version-aware path resolution to the Kimi adapter:

1. **Detection**: Prefer `~/.kimi-code` when present, fallback to `~/.kimi`
2. **Path methods**: Update GlobalConfigDir, SystemPromptDir, SystemPromptFile, SettingsPath, MCPConfigPath, SubAgentsDir to resolve dynamically
3. **SkillsDir**: Return `~/.kimi-code/skills` for v0.11+, keep `~/.config/agents/skills` for legacy
4. **PostInstallMessage**: Show v0.11+-appropriate message without `--agent-file`
5. **System scan**: Update config_scan.go to detect `~/.kimi-code`
6. **Agent builder**: Add AgentKimi to agentBuilderSkillsDir so generated skills install correctly

## Alternatives Considered

- **Explicit version flag**: Rejected — directory existence is deterministic and requires no user action
- **Keep using shared skills path for v0.11+**: Rejected — native path maximizes compatibility with `kimi --prompt "List skills"`

## Additional Context

- kimi-code v0.11.0 documentation confirms `~/.kimi-code/skills/` as the user skills directory
- Agent YAML files (`--agent-file`) were removed in v0.11+; functionality migrated to native Skills
- SDD change documentation: openspec/changes/kimi-code-v0.11-support/

## Checklist

- [x] I understand that PRs will be rejected if the linked issue does not have status:approved

salema97:
## Linked Issue

Closes #782

---

## PR Type

---

## Summary

End-to-end support for the current Node.js **Kimi Code CLI** (`~/.kimi-code`, overridable via `KIMI_CODE_HOME`) while preserving compatibility with the legacy Python/uv `~/.kimi` layout.

Key points from maintainer review are included:
- Rebased onto latest `main` and kept this PR to the Kimi scope only
- Global prompt target is `{config-root}/AGENTS.md` (not `KIMI.md`)
- Project instructions stay project-scoped; install does not copy cwd `AGENTS.md` into global config
- Naming uses current kimi-code (layout unchanged since 0.11 through 0.23.x)
- Install uses npm `@moonshot-ai/kimi-code@0.23.6` with Node.js >= 22.19.0 preflight
- Windows SessionStart hook no longer uses `--cwd "$PWD"`

---

## Changes

|-------------|-------------|
| `internal/agents/kimi/*` | Current kimi-code adapter, plugin install/registration, AGENTS.md hub, tests |
| `internal/installcmd/*` | npm-based kimi-code install + Node preflight |
| `internal/components/sdd|skills|persona|permissions|engram` | Kimi routing, hooks, injection, and tests |
| `internal/system|tui|catalog|skillregistry|versions` | Path detection, agent-builder skills path, pinned version |
| `docs/*`, `openspec/*` | Current kimi-code docs and archived SDD artifacts |

---

## Test Plan

- [x] `go test ./internal/agents/kimi/...`
- [x] `go test ./internal/components/sdd ./internal/components/skills ./internal/components/persona ./internal/components/permissions ./internal/installcmd ./internal/system`
- [x] `go vet ./...`
- [x] Rebased onto current `main`; dropped stale `TestInjectCodexWritesGentleDevPermissionsProfile` and kept main's `TestInjectCodexRemovesInjectedGentleDevProfile`
- [ ] Full `go test ./...` / E2E in CI

---

## Contributor Checklist

- [x] PR is linked to an issue with `status:approved` (#782)
- [x] Unit tests for the Kimi scope pass locally
- [x] Go format passes
- [x] Documentation updated for current kimi-code
- [x] Conventional Commits, no `Co-Authored-By`
- [ ] Maintainer `size:exception` may still be needed (Kimi scope + archived openspec history is large)

---

## Notes for Reviewers

Tracking from sandbox testing (not blockers for this Kimi feature PR):
- #1828 uninstall JSON rewrite vs kimi TOML settings
- #1829 theme injection would overwrite TOML/YAML settings

Judgment Day agents currently ship for Claude/Kiro only. Kimi assets still mention `jd-fix-agent`; happy to take that as a follow-up rather than expand this PR unless maintainers want it folded in.


salema97:
@coderabbitai review

Alan-TheGentleman (changes requested):
I’m requesting changes because the current global config write crosses a provider boundary. `BootstrapTemplate()` copies cwd-derived `AGENTS.md` into global Kimi config, so running install from an untrusted repo can persist repo-controlled instructions into future sessions. Please keep project instructions project-scoped or require an explicit trust step before writing them globally. Also route plugin registration and skill discovery through the resolved Kimi config dir so `KIMI_CODE_HOME` does not split config, plugins, and skills across different trees.

salema97:
Thanks for the clear security/architecture feedback. Both concerns should now be addressed in the latest push.

**1. Project instructions are now project-scoped**

`BootstrapTemplate()` no longer reads `AGENTS.md` from the cwd, and it no longer writes cwd-derived content to the global Kimi config (`~/.kimi-code/AGENTS.md` or `~/.kimi/AGENTS.md`). Instead, `${KIMI_AGENTS_MD}` is replaced with a static placeholder that tells Kimi to respect the current worktree's `AGENTS.md` at runtime. This removes the provider-boundary crossing and the untrusted-repo persistence risk.

**2. Plugin and skill paths now route through `resolveConfigDir`**

All v0.11+ plugin and skill paths are now rooted under the resolved Kimi config directory, so `KIMI_CODE_HOME` no longer splits config, plugins, and skills:

- `PluginDir()` uses `resolveConfigDir(homeDir)` instead of hardcoding `~/.kimi-code`.
- `registerPlugin()` writes `installed.json` under the same resolved directory.
- `AllSkillsDirs()` uses `resolveConfigDir(homeDir)` for the user skills directory.

Tests added for the new `KIMI_CODE_HOME` behavior and for the project-scoped AGENTS.md handling. `go test ./internal/agents/kimi/... ./internal/components/persona/...` passes.

Let me know if you'd prefer an explicit trust step instead of the project-scoped approach, or if anything else looks off.

Alan-TheGentleman:
This is now the canonical Kimi PR after consolidation (#939 closed). To get it moving:

1. Rebase onto main and shrink the PR to the Kimi scope. The update/advisory and trigger-rules parts already landed via other PRs and are most of the current conflict.
2. Verify the global prompt target: current kimi-code docs (0.18 through 0.23) only recognize ~/.kimi-code/AGENTS.md, and KIMI.md appears nowhere. Target AGENTS.md for the current path while keeping the project-scoped trust fix intact.
3. Confirm against 0.23.x (layout unchanged since 0.11) and update naming from v0.11+ to current.
4. Optional, folded from #939 with credit: npm-based install with a Node >=22.19 preflight.

Once rebased I'll re-review.

salema97:
Rebased onto latest `main` and addressed all four points from your review. Ready for re-review.

**1. Rebase + scope shrink** — Rebuilt the branch cleanly off current `main` (now 5 focused commits). The update/advisory and trigger-rules work that had already landed via other PRs is no longer re-applied here; this branch is now purely the Kimi scope, which removes most of the prior conflict surface.

**2. Global prompt target → AGENTS.md** — Verified against the current kimi-code docs: `KIMI.md` is not a recognized file in either current kimi-code or the legacy Python CLI (the `KIMI_` prefix in `${KIMI_AGENTS_MD}` is only a system-prompt variable name). The global instructions file is now written to `{config-root}/AGENTS.md` instead of `KIMI.md`. The project-scoped trust fix is intact — we still do NOT copy cwd-derived `AGENTS.md` into the global config; the skeleton keeps the "read from the current worktree at runtime" placeholder.

**3. Naming** — Dropped the `v0.11+` branding in favor of current kimi-code naming (e.g. `isV11Plus` → `usesKimiCodeLayout`, install message, comments). Confirmed the `~/.kimi-code` layout against the current releases.

**4. npm-based install (folded in)** — This turned out to be load-bearing, not just optional: the previous `resolveKimiInstall` used `uv tool install kimi-cli`, which installs the **legacy Python** product (`~/.kimi`) — so an auto-install would never activate the kimi-code layout this PR adds. Now installs the current product via `npm install -g @moonshot-ai/kimi-code@0.23.6` with a Node.js >= 22.19.0 preflight (per the official install docs). Detection still supports both layouts; only the install target changed to current.

Also documented in code/docs: the two kimi products (legacy `uv`/Python/`~/.kimi` vs current `npm`/Node/`~/.kimi-code`) behave differently, and the adapter keeps them distinct (plugin + config extras gated to the kimi-code layout; legacy falls back to `~/.kimi` and `~/.config/agents/skills`).

`go build ./...`, `go vet ./...` clean; `internal/installcmd`, `internal/agents/kimi`, `internal/versions` tests pass. (Note: I couldn't update the PR title via API to drop "v0.11+" — token scope — feel free to retitle to "current kimi-code".)

salema97:
Quick follow-up with one more fix in the latest push (now 6 commits, head 228e501).

**Windows fix: dropped `--cwd "$PWD"` from the SessionStart skill-registry hook**

The injected hook used `gentle-ai skill-registry refresh ... --cwd "$PWD"`. Windows hook shells do not expand the Unix `$PWD`, so the tool resolved the literal string and created a stray `<repo>/$PWD/.atl` directory instead of `<repo>/.atl`. Since kimi-code runs hook commands with cwd set to the project directory and the refresh command already defaults to cwd, the flag was redundant on Unix and broken on Windows.

The new command is `gentle-ai skill-registry refresh --quiet --no-gitignore || true`, and `ensureKimiSkillRegistryHook` now migrates existing configs: on the next sync it removes the old `--cwd \"$PWD\"` variant (matched by marker, SessionStart blocks only) and leaves exactly one correct hook. Unrelated user `[[hooks]]` blocks are preserved. Covered by regression tests, `go build`/`go vet`/`gofmt` clean.

drccastillo:
Built this branch and ran it in a fake-HOME sandbox, nice work overall. Three things from testing:

1. Heads-up (not this PR's fault): `gga uninstall --agent kimi` crashes on current main because the uninstall rewrite assumes JSON while kimi settings are TOML. Filed as #1828. The same fix class applies to this branch's `~/.kimi-code` layout, so it is worth tracking before merge.

2. Related latent issue filed as #1829: theme injection would overwrite TOML/YAML settings with JSON. Same guard class as above.

3. Rebase nit for the next round: `internal/components/permissions/inject_test.go` still carries `TestInjectCodexWritesGentleDevPermissionsProfile`, but main reversed that behavior (codex now removes the injected gentle-dev profile, see main's `TestInjectCodexRemovesInjectedGentleDevProfile`). That hunk is stale and should be dropped.

drccastillo:
One more finding from testing the kimi assets in a sandbox: `internal/assets/kimi/sdd-orchestrator.md` references `jd-fix-agent` in the Judgment Day flow, but the JD agents (jd-judge-a, jd-judge-b, jd-fix-agent) only ship for claude and kiro, so kimi installs end up with a dangling reference to an agent that does not exist.

We already ported the trio for kimi in a local branch and would be glad to contribute it, either folded into this PR or as a follow-up after merge. Would you want Judgment Day support for kimi?

salema97:
Thanks for the sandbox testing. Addressed in the latest push (`d5a0d9d5`, rebased onto current `main`):

1. **#1828 / #1829** — agreed these are real and worth tracking before merge. They are not introduced by this PR's feature path, so I left them as follow-ups rather than expanding this Kimi scope. Same guard class applies to `~/.kimi-code`.

2. **Stale Codex permissions test** — fixed during the rebase. This branch now keeps main's `TestInjectCodexRemovesInjectedGentleDevProfile` and only adds `TestPermissionsInject_KimiReturnsNil`. The old `TestInjectCodexWritesGentleDevPermissionsProfile` hunk is gone.

3. **Judgment Day / `jd-fix-agent`** — also confirmed: Kimi assets reference `jd-fix-agent`, but the JD trio currently ships only for Claude and Kiro. I'd prefer to keep that as a follow-up after merge so this PR stays focused on current kimi-code support. If maintainers want it folded in, I'm happy to take your port.

Also updated the PR title/description to drop the stale `v0.11+` / `KIMI.md` wording and match the current branch.

salema97:
@Alan-TheGentleman title and description are now updated to current kimi-code / `AGENTS.md` wording. Branch is rebased onto latest `main` (7 focused commits) and ready for another look.

salema97:
Pushed `4e59dd62` with three hardening fixes that came out of a post-rebase security/resilience review of this branch:

1. **Fresh installs now bootstrap into `~/.kimi-code`.** Previously, on a machine with neither config dir, the resolver defaulted to legacy `~/.kimi`, so `AGENTS.md`/`config.toml`/plugins landed in a tree the npm-installed kimi-code CLI never reads. `~/.kimi` is now used only when it actually exists. `config_scan` and the TUI agent-builder path mirror the same order (including `KIMI_CODE_HOME`).

2. **Credential deny rules in the generated `config.toml`.** Since Kimi bypasses the permissions overlay (rules live in `config.toml`), the broad `Write`/`Edit` allows had no guardrails. The template now denies `Read`/`Write`/`Edit` on `.env`, `.ssh`, `.aws/credentials`, `*.pem`/`*.key`, `secrets/`, gh `hosts.yml`, and keychains, matching the deny lists injected for the other agents. Kimi resolves `deny > ask > allow`, so ordering is safe.

3. **`installed.json` is treated as shared state.** Registration now fails on a corrupt registry instead of rewriting it (which wiped other plugins entries), and unknown top-level/per-plugin fields survive the rewrite via `json.RawMessage`.

Specs and docs updated to match, plus regression tests for each (fresh-default resolution, deny rules, corrupt/lossy registry). Full test suite is green apart from the pre-existing Windows baseline failures.

dnlrsls:
Please resolve the current changes request and obtain a fresh review before addressing the conflict. The blocker is the trust boundary in which cwd-derived AGENTS.md can be persisted globally, plus consistent resolved-directory handling for KIMI_CODE_HOME, plugins, and skills.

salema97:
I've rebased the branch onto latest `main` and resolved the remaining review findings:

1. **Testable Config Path Resolution**: Fixed `resolveConfigDir` to use the adapter's testable `a.statPath(envDir)` helper instead of raw `os.Stat(envDir)` when resolving `KIMI_CODE_HOME`.
2. **Sandbox unit tests**: Updated the mocked `statPath` in `plugin_test.go` (`newKimiCodeAdapter`) to recognize `KIMI_CODE_HOME` environment variables so they don't incorrectly fall back to the default `~/.kimi-code` directory during testing.

All unit tests (`kimi`, `system`, `skills`, `sdd`) are now passing cleanly.

Alan-TheGentleman:
This has been a long haul and you have answered every round of review on it, including the trust-boundary fix and the Windows hook bug. The Kimi scope is still wanted.

What moved: #2450, shipped inside #2462, removed agent runtime auto-install entirely. `internal/versions/versions.go` no longer carries agent pins at all, and its package comment now records why: the install commands are display-only text that a human reads and runs, so a frozen version goes stale the moment upstream ships. The resolvers advise `latest` now (see `resolveClaudeCodeInstall` and `resolveKilocodeInstall` in `internal/installcmd/resolver.go`).

So `const KimiCode = "0.23.6"` should go, and `resolveKimiInstall` should advise `@moonshot-ai/kimi-code@latest` with `--ignore-scripts`, matching the other resolvers. Your new `PluginInstaller` interface also anchors directly after the `CapabilityAutoInstall` const block, which was deleted; move it above `type Adapter interface` and it applies.

Worth a look while rebasing: Kimi already has a row in `internal/agents/capabilitymanifest/manifest.go`, so double-check its claims still describe the `~/.kimi-code` layout. Everything else, the `KIMI_CODE_HOME` resolution, the `AGENTS.md` target, the plugin architecture, the credential deny rules and the Windows `$PWD` hook fix, applies unchanged.

On your earlier question: Judgment Day for Kimi is worth a separate PR, not this one.

salema97:
Rebased onto current main and applied your review points.

Changes:
- Removed the pinned `KimiCode` version and switched `resolveKimiInstall` to `@moonshot-ai/kimi-code@latest --ignore-scripts`, matching the display-only refusal behavior.
- Removed Kimi from the npm preflight path; only Pi is preflighted now, per #2450/#2462.
- Dropped the `trigger-rules` injection block in `sdd.Inject` because it depended on `RenderTriggerRules`/`catalog.DefaultTriggerRuleSet`, which were removed from main during the root simplification.
- Fixed stale non-`/v2` module imports in the Kimi plugin files left over from the rebase.
- Updated tests to expect the refusal/install-command wording.

Tests in the Kimi scope pass, and the targeted CLI tests around install refusal and dependency checks also pass. Full `./internal/cli/...` timed out locally, but the relevant subset is green.

PR #859, opened 2026-06-13
fix(upgrade): handle missing engram process on Windows

### LINKED ISSUE #815: fix(installer): clean Windows install fails - Stop-Process errors when no engram process exists (regression of #807)
### Pre-flight Checks

### Bug Description
On a first-time (clean) Windows install, `component:engram` fails during "download engram binary". The pre-download step that stops running engram processes (added in #807) errors out because there is no engram process to stop. `Get-Process -Name engram` throws a terminating error, the pipeline "completed with errors", and the whole engram component is marked FAILED - even though nothing was actually wrong.

### Steps to Reproduce
1. Fresh Windows machine, engram never installed (no engram.exe, no process running).
2. `scoop install gentle-ai` (v1.38.0).
3. Run installer / select the engram component.
4. `component:engram` fails.

### Expected Behavior
With no engram process running, the stop step should be a no-op and the download should proceed.

### Actual Behavior
FAILED: component:engram - download engram binary: stop running engram processes before upgrade: powershell Stop-Process pipeline completed with errors

Root cause reproduced directly:
PS> Get-Process -Name engram | Stop-Process -Force
Get-Process : No process with the name 'engram' was found.

### Suggested Fix
Guard the lookup so "no process" is not an error:
Get-Process -Name engram -ErrorAction SilentlyContinue | Stop-Process -Force
This is a regression edge-case of the fix in #807.

### Operating System
Windows

### Agent / Client
Claude Code

### Shell
Other (PowerShell)

### Additional Context
Workaround that unblocked a clean install: start a dummy process named engram.exe before installing so Stop-Process has something to kill. Once engram is installed and current, re-running `gentle-ai install` skips the download+stop and succeeds.

skiveldev:
Fixes #815

## Summary
- Avoid piping directly from Get-Process into Stop-Process on Windows.
- Make the Engram stop step succeed when the process is not running.
- Add a Windows-specific regression test for the missing-process path.

## Changes
| File | Change |
|------|--------|
| \internal/components/engram/download.go\ | Guard the PowerShell stop command with an explicit process lookup |
| \internal/components/engram/download_test.go\ | Add regression coverage for missing process handling |

## Test plan
- [x] \go test ./internal/components/engram\
- [x] On Windows PowerShell 5.1, verified missing Engram process handling succeeds without blocking upgrade.
- [x] Verified stop-process failure still aborts before replacing \engram.exe\.

## Notes
I also reproduced this through the interactive TUI \upgrade tools\ path while upgrading Engram on Windows/Scoop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows shutdown reliability: shutdown now tolerates missing target processes, forces a successful exit, and provides clearer, trimmed error diagnostics including the target process name.

* **Tests**
  * Added Windows-specific test coverage for scenarios where target processes are not found.

### DIFF
--- a/internal/components/engram/download.go
+++ b/internal/components/engram/download.go
@@ -438,6 +438,9 @@ func stopEngramProcesses() error {
 	return stopEngramProcessesNamed("engram")
 }
 
+// stopEngramProcessesNamed stops running processes with the given name.
+// Missing processes are treated as success so clean Windows installs and
+// upgrades do not fail when Engram is not currently running.
 func stopEngramProcessesNamed(processName string) error {
 	escapedProcessName := strings.ReplaceAll(processName, "'", "''")
 	script := fmt.Sprintf("$p = Get-Process -Name '%s' -ErrorAction SilentlyContinue; if ($p) { $p | Stop-Process -Force -ErrorAction Stop }; exit 0", escapedProcessName)
--- a/internal/components/engram/download.go
+++ b/internal/components/engram/download.go
@@ -431,18 +431,26 @@ func extractZipBinary(data []byte, binaryName, outPath string) error {
 }
 
 // stopEngramProcesses stops any running Engram process so Windows can replace
-// engram.exe during upgrade. Missing processes are not an error because
-// Get-Process uses SilentlyContinue.
+// engram.exe during upgrade. Missing processes are not an error; avoid piping
+// directly from Get-Process because Windows PowerShell 5.1 can still exit 1
+// when the process does not exist.
 func stopEngramProcesses() error {
+	return stopEngramProcessesNamed("engram")
+}
+
+func stopEngramProcessesNamed(processName string) error {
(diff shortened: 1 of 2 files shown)


skiveldev:
CodeRabbit suggested the \	ype:bug\ label, but I do not have permission to add labels on the upstream repo. Could a maintainer please add it?

Alan-TheGentleman:
Thanks for the precise diagnosis. You nailed the root cause: on PowerShell 5.1 an empty `Get-Process` pipeline flips `$?` and exits 1, which is exactly what was failing clean installs. The production fix already shipped in #866, and #884 just landed a regression test that pins the defensive script shape (`SilentlyContinue`, `if ($procs)` guard, no `-ErrorAction Stop`) so it cannot silently come back. You are credited as co-author on that commit (`61423f5`). Closing this PR since the change is already on main, but your contribution is in. Thank you.

PR #904, opened 2026-06-16
fix(upgrade): verify tool is brew-managed before routing to brew upgrade

### LINKED ISSUE #186: fix(upgrade): brew upgrade fails for non-brew tools on Linux with Linuxbrew
## Bug Description

On Linux systems where Linuxbrew is installed for unrelated packages (e.g., `zellij`, fonts), the upgrade command fails with:

```
brew upgrade gentle-ai: exit status 1 (output: Error: No available formula with the name)
brew upgrade gga: exit status 1 (output: Error: No available formula with the name "gga".)
```

## Root Cause

`effectiveMethod()` in `internal/update/upgrade/executor.go` checks only whether `profile.PackageManager == "brew"` and blindly returns `InstallBrew` for **all** tools. On Linux, platform detection (`internal/system/detect.go:129-132`) returns `"brew"` as the package manager if Linuxbrew is installed — even if the tools themselves were installed via binary download (`~/.local/bin/`) or script.

## Expected Behavior

If a tool is not actually installed via Homebrew (not in the Cellar), the upgrade should fall back to the tool's declared `InstallMethod` (binary download, script, go-install).

## Steps to Reproduce

1. Install Linuxbrew on a Linux system (e.g., Fedora)
2. Install `gentle-ai` and `gga` via the install script (they end up in `~/.local/bin/`)
3. Run `gentle-ai` → select "Upgrade + Sync"
4. Both tools fail with "No available formula" errors

## Environment

- OS: Fedora 43 (Linux 6.19.9)
- Linuxbrew installed at `/home/linuxbrew/.linuxbrew/`
- `gentle-ai` and `gga` installed in `~/.local/bin/` (not via brew)
- brew only manages: `zellij`, `ca-certificates`, `openssl`, fonts

## Proposed Fix

Check `brew list --versions <toolName>` before choosing the brew strategy. If the tool is not brew-managed, fall back to `tool.InstallMethod`.

decode2:
## Summary

Fixes upgrade failures on Linux systems where Linuxbrew is installed for unrelated packages. Previously, `effectiveMethod()` would route ALL tools to `brew upgrade` when `profile.PackageManager == "brew"`, even if the tool was installed via `go install` or binary download.

## Root Cause

On Linux, platform detection sets `PackageManager = "brew"` when Linuxbrew is present on PATH — regardless of how individual tools were actually installed. The upgrade logic then blindly routed every tool to `brew upgrade`, which failed with "No available formula" for tools not managed by Homebrew.

## Fix

1. Added `isBrewManaged(ctx, toolName)` helper in `strategy.go` that probes `brew list --versions <toolName>` to verify the tool is actually installed via Homebrew.

2. Updated `effectiveMethod()` to check `isBrewManaged()` before routing to `InstallBrew`. If the tool is not brew-managed, it falls through to the next priority (go-install or declared method).

3. Added `execCommandContext` package var for testability (same pattern as existing `execCommand`).

## Tests

- Updated `TestEffectiveMethod` with new cases for issue #186:
  - `brew profile + tool NOT brew-managed + go-install → go-install`
  - `brew profile + tool NOT brew-managed + binary → binary`
- Existing brew-managed cases still pass (regression guard)
- Added `mockExitCommand` helper for simulating `brew list` exit codes

Closes #186

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Homebrew upgrade routing to only use the Homebrew path when the tool is confirmed Homebrew-managed via deterministic formula/cask probing.

* **Improvements**
  * Made install-method selection context-aware for better cancellation/timeout behavior during upgrades.
  * Added in-session Homebrew detection caching to reduce repeated Homebrew checks.

* **Tests**
  * Expanded and standardized upgrade strategy and Homebrew detection tests with deterministic command mocking.
  * Updated backup-path upgrade tests to use platform-aware path helpers.

### DIFF
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -371,7 +371,7 @@ func TestRunArgsDispatchesCompactReviewFacadeBeforePlatformValidation(t *testing
 	if err := RunArgs([]string{"review", "--help"}, &output); err != nil {
 		t.Fatalf("RunArgs(review --help) error = %v", err)
 	}
-	if !strings.Contains(output.String(), "review <start|finalize|validate>") {
+	if !strings.Contains(output.String(), "review <start|finalize|validate|invalidate>") {
 		t.Fatalf("compact review help missing:\n%s", output.String())
 	}
 }
--- a/internal/cli/review_facade.go
+++ b/internal/cli/review_facade.go
@@ -41,6 +41,13 @@ type ReviewFacadeFinalizeResult struct {
 	ReceiptPath   string                  `json:"receipt_path,omitempty"`
 }
 
+type ReviewInvalidateResult struct {
+	Operation     string                  `json:"operation"`
+	LineageID     string                  `json:"lineage_id"`
+	State         reviewtransaction.State `json:"state"`
+	StoreRevision string                  `json:"store_revision"`
+}
+
 type facadeFinding struct {
 	ID                string                              `json:"id,omitempty"`
 	Lens              string                              `json:"lens,omitempty"`
@@ -85,7 +92,7 @@ type facadeArtifacts struct {
 
 func RunReview(args []string, stdout io.Writer) error {
 	if len(args) == 0 || args[0] == "help" || args[0] == "-h" || args[0] == "--help" {
-		_, _ = fmt.Fprintln(stdout, "Usage: gentle-ai review <start|finalize|validate> [flags]\n\nOrdinary review facade; repository scope, authority, canonical artifacts, and lifecycle transitions are derived by Go.")
+		_, _ = fmt.Fprintln(stdout, "Usage: gentle-ai review <start|finalize|validate|invalidate> [flags]\n\nOrdinary review facade; repository scope, authority, canonical artifacts, and lifecycle transitions are derived by Go.")
 		return nil
 	}
 	switch args[0] {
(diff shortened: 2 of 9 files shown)


Alan-TheGentleman (changes requested):
This is the lower-risk direction because Homebrew ownership should come from Homebrew, not from path heuristics. Please cache the resolved method once per tool so we do not run repeated `brew list` probes during the same upgrade path. Also add one mixed probe test where `--formula` fails and `--cask` succeeds. After that, this is the PR I would keep for #186.

decode2:
Changes requested have been applied:

✅ Cache ffectiveMethod result per tool to avoid repeated rew list probes
✅ Added mixed probe test (formula fails + cask succeeds)

The cache is implemented at the isBrewManaged() level with a package-level map protected by sync.RWMutex. This ensures we don't spawn multiple rew list processes for the same tool during a single upgrade run.

Ready for re-review.

decode2:
@Alan-TheGentleman Changes requested have been applied and are ready for re-review.

Alan-TheGentleman (approved):
Approved. The routing change is conceptually correct: a brew profile should not force every tool through Homebrew unless the specific tool is actually brew-managed.

The cache addresses the repeated-probe concern well enough for this CLI process model. This does not solve the Windows PATH shadowing issue in #943, so that should stay separate.

decode2:
Hi @Alan-TheGentleman, I've updated this PR:
- Refactored \xecutor.go\ to compute \ffectiveMethod\ once per tool and pass it down to \xecuteOne\ and \
unStrategy\ to eliminate redundant Homebrew shell-out probes.
- Added \TestIsBrewManaged_MixedProbe\ to cover the mixed probe scenario where \--formula\ fails but \--cask\ succeeds.

Ready for re-review!

decode2:
@coderabbitai review

decode2:
@coderabbitai review

Alan-TheGentleman (changes requested):
Good direction, but this cannot merge while the PR does not compile. The current Unit Tests job fails in `internal/update/upgrade/strategy_test.go` because two test call sites still call `runStrategy(ctx, result, profile)` after the function was changed to require the resolved install method as a fourth argument. Please update those remaining calls, rerun the unit tests, and ping for re-review.

decode2:
@Alan-TheGentleman ready for re-review! I've rebased the branch on main and resolved the compilation failures in strategy_test.go by updating the outdated runStrategy call sites.

decode2:
@Alan-TheGentleman all your review points are addressed on the current head (`c2c51d2`):

- The resolved brew method is now cached per tool (`brewManagedCache`) so there are no repeated `brew list` probes, with the mixed-probe regression test (`TestIsBrewManaged_MixedProbe`, `--formula` fails / `--cask` succeeds).
- The compile break is fixed — both `runStrategy` test call sites now pass the 4th `update.InstallInstaller` arg.

CI is green and CodeRabbit's latest pass has no actionable comments. Ready for re-review whenever you have a moment 🙏

decode2:
Rebased this branch onto current upstream/main (6030cad). The original #186 fix and all still-valid review feedback are now fully superseded by the active Homebrew ownership implementation already on main (notably e425d28 and d46847f), which preserves formula/cask ownership and verifies the active executable path. Reapplying the old helper/cache architecture would regress that design, and the later runStrategy call-site commit no longer compiles against the current signature, so the rebased branch intentionally has no diff. GitHub closed the PR automatically after the force-with-lease update. Review history remains: Alan approved cad09f6 on June 22, then requested changes on 52c03b8 on June 26 for stale runStrategy test calls; c2c51d2 addressed that request, but no later approval was submitted, so the final recorded reviewDecision is still CHANGES_REQUESTED despite the earlier approval. No re-review or CI run is needed for an empty, closed PR.

PR #952, opened 2026-06-25
fix(doctor): detect PATH-shadowed binaries on Windows via PATHEXT

### LINKED ISSUE #924: Windows: gga install creates no .cmd shim → doctor reports it missing; adding a .cmd then triggers a false "3 copies" duplicate warning
### Pre-flight Checklist

### 📝 Bug Description

On Windows, installing the gga component leaves gentle-ai doctor unable to pass the tool:gga check, no matter what. There are two distinct root causes that compound each other:

1. **No PATHEXT-visible shim is installed.** gentle-ai install --components gga provisions the gga **bash script** (extensionless) plus a gga.ps1 wrapper into a PATH directory, but no .cmd/.exe/.bat. Because gentle-ai doctor (a Go binary) resolves tools using the Windows PATHEXT rules — which do **not** include extensionless files or .ps1 — it reports gga not found in PATH even though gga is fully functional from bash and PowerShell.

2. **The duplicate check misfires once a .cmd is added.** After manually adding a gga.cmd shim so the doctor can see it, doctor then flags 3 copies found in PATH and prints the same directory three times. It is counting the base script + its two launchers (gga, gga.ps1, gga.cmd) as three duplicate binaries, and strips the extensions when printing (so the output reads ...\gga, ...\gga, ...\gga). The PATH directory is present only once — this is a false positive.

Net effect: doctor goes from **unhealthy** (cause 1) to **degraded** (cause 2). It can never reach healthy on a standard Windows gga layout.

### 🔄 Steps to Reproduce

1. On Windows, run gentle-ai install --components gga --scope global.

2. Run gentle-ai doctor.
   - **Actual:** [xx] tool:gga — gga not found in PATH → Status: unhealthy.
   - where.exe gga and gga --version both work (v2.8.1), so gga itself is fine.

3. Add a gga.cmd shim in the same PATH dir (so the doctor's PATHEXT lookup can resolve it) that invokes the bash script.

4. Run gentle-ai doctor again.
   - **Actual:** [!!] tool:gga — gga resolved to ...\gga.cmd but 3 copies found in PATH: ...\gga, ...\gga, ...\gga → Status: degraded.
   - The dir is on PATH exactly once (verified); the "3 copies" are gga + gga.ps1 + gga.cmd.

### ✅ Expected Behavior

- After installing gga on Windows, gentle-ai doctor should detect it and pass [ok] tool:gga), reaching a healthy status.

- A base script plus its .ps1/.cmd launchers in the same directory should be treated as **one** tool, not as duplicate copies.

### ❌ Actual Behavior

- Step 2: [xx] tool:gga — gga not found in PATH → Status: unhealthy (even though gga works from bash/PowerShell).

- Step 4 (after manual .cmd shim): [!!] tool:gga — gga resolved to ...\gga.cmd but 3 copies found in PATH: ...\gga, ...\gga, ...\gga → Status: degraded (false positive duplicate warning).

### Gentle AI Version

1.40.2

### Operating System

Windows

### AI Agent / Client

Claude Code

### 📋 Affected Area

CLI (commands, flags)

### 💡 Logs / Error Output

```shell

```

## Environment
- OS: Windows
- gentle-ai: 1.40.2 (installed via Scoop)
- gga: v2.8.1
- gga files in PATH dir (C:\Users\<user>\bin): gga (bash script), gga.ps1 (wrapper)
- PATHEXT: .COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC (no .PS1, no extensionless)

## Suggested fixes
1. **Installer:** on Windows, create a PATHEXT-visible shim (e.g. gga.cmd, similar to how Scoop generates shims) alongside the bash script and .ps1, so the doctor's lookup resolves it out of the box.

2. **Doctor duplicate check:** when the same basename resolves to a script + its .ps1/.cmd launchers in the same directory, count it as a single tool. Also print full filenames (with extensions) instead of stripping them, so the message doesn't read as gga, gga, gga.

## Workaround (current)
Manually adding a gga.cmd shim that locates Git Bash and runs gga makes gga detectable and functional from cmd/PowerShell/bash. The doctor then stays at "degraded" due to the false-positive duplicate count, but there are no real failures.

### LINKED ISSUE #951: doctor: PATH-shadow detection misses Windows executables (.exe/PATHEXT) and reports [ok] while updates silently fail
## Summary

`gentle-ai doctor`'s PATH-shadow detection silently does nothing on Windows because it probes the bare tool name without an executable extension. As a result `doctor` reports `[ok]` for a tool even when a stale duplicate shadows the managed copy on `PATH` — the exact situation it's meant to catch.

## Where

`internal/cli/doctor.go`, `checkOneTool`:

```go
var copies []string
for _, dir := range pathDirs {
    if _, statErr := os.Stat(filepath.Join(dir, tool)); statErr == nil {
        copies = append(copies, filepath.Join(dir, tool))
    }
}
if len(copies) > 1 { /* warn: N copies found in PATH */ }
```

It stats `filepath.Join(dir, tool)` — e.g. `…\gentle-ai` — with **no extension**. On Windows the binaries are `gentle-ai.exe` (and shims like `gentle-ai.cmd`), so every `os.Stat` misses, `copies` stays empty, and `len(copies) > 1` is never true. Note the inconsistency: `resolved` comes from `exec.LookPath`, which *does* honor `PATHEXT`, so the resolver and the duplicate scan disagree.

## Real-world repro

- `irm` installer placed gentle-ai at `%LOCALAPPDATA%\gentle-ai\bin\gentle-ai.exe` (1.42.0).
- A stale scoop copy at `scoop\shims\gentle-ai.exe` (1.40.2) sat **earlier on PATH** and shadowed it.
- `gentle-ai update` / the installer reported success and "Updated to v1.42.0", but `gentle-ai version` stayed **1.40.2**.
- `gentle-ai doctor` showed `[ok] tool:gentle-ai  gentle-ai found at C:\…\scoop\shims\gentle-ai.exe` — no warning at all.

Result: updates appear to apply but never take effect, and the one diagnostic meant to flag it gives a false all-clear.

## Why this matters more on Windows (security-software angle)

These partial/duplicate-install states are easy to hit on Windows, and security software makes them more common, not less. Unsigned, freshly-downloaded binaries (gentle-ai, engram, opencode) routinely trip behavioral AV such as Kaspersky System Watcher / "activity monitor": they get **quarantined or deleted mid-install/-update**, or the install/update behavior itself is flagged as suspicious. Even after adding exceptions, that history leaves messy state — a managed copy nuked, a stale shadow left on PATH, shims renamed/disabled (`*.bak`), etc.

That's precisely when an accurate shadow check earns its keep: it should tell the user "you're running a different copy than the one that was updated" instead of falsely reassuring them. So getting Windows executable resolution right in `doctor` is high-value, not cosmetic.

## Proposed fix

In the duplicate scan, probe executable extensions on Windows (`PATHEXT`), matching `exec.LookPath` semantics, so `gentle-ai`/`gentle-ai.exe`/`gentle-ai.cmd` are all detected. PR incoming.

### Possible follow-ups (out of scope here)
- Flag disabled/renamed shims (`*.bak`, `*.disabled`) sitting next to a tool's dir.
- Warn when a managed binary is missing where `state.json` says it should be (e.g. AV quarantine).

jojosenthusiast:
## What

`gentle-ai doctor`'s PATH-shadow detection (`checkOneTool`) never fires on Windows because it probes the bare tool name with no executable extension:

```go
os.Stat(filepath.Join(dir, tool)) // "…\gentle-ai" — never exists on Windows
```

Windows binaries are `gentle-ai.exe` (shims `.cmd`), so every `os.Stat` misses, `copies` stays empty, and `doctor` reports `[ok]` even when a stale duplicate shadows the managed copy on `PATH`. Note the inconsistency: `resolved` comes from `exec.LookPath`, which *does* honor `PATHEXT`, so the resolver and the duplicate scan disagree.

Closes #951.

## Real-world repro

- `irm` installer placed gentle-ai at `%LOCALAPPDATA%\gentle-ai\bin\gentle-ai.exe` (1.42.0)
- a stale scoop copy `scoop\shims\gentle-ai.exe` (1.40.2) sat earlier on `PATH`
- `gentle-ai update` / the installer reported success ("Updated to v1.42.0"), but `gentle-ai version` stayed **1.40.2**
- `gentle-ai doctor` showed `[ok] tool:gentle-ai  found at …\scoop\shims\gentle-ai.exe` — no warning

So updates appear to apply but never take effect, and the one diagnostic meant to catch it gives a false all-clear. This is especially easy to hit on Windows where security software (e.g. Kaspersky System Watcher) quarantines/deletes unsigned binaries mid-install or disables shims, leaving exactly these partial/duplicate states.

## Fix

Probe executable extensions on Windows (`PATHEXT`) in the duplicate scan so it matches `exec.LookPath` semantics. Non-Windows behavior is unchanged (`[""]`). The extension provider is an overridable var (`executableExtsFn`) following the file's existing "Overridable for testing" pattern.

## Tests

- `TestCheckOneTool_ShadowedWindowsExt` — two copies (`.exe` + `.cmd`) → warns; deterministic on any OS via the override.
- `TestExecutableExtensions` — per-platform extension set.
- Existing bare-name tests pinned to `[""]` so they stay green on all platforms.
- `gofmt`, `go vet`, and `go test ./internal/cli/` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the doctor check’s PATH shadowing/duplicate detection on Windows so it correctly accounts for executable extensions (for example, `.exe`/`.cmd`) and reports duplicate copies more reliably.
  * Made extension and PowerShell/CMD shim handling consistent in the doctor output, reducing false positives/incorrect counts.

* **New Features**
  * On Windows, the installer now also ensures a `gga.cmd` command shim in addition to the existing PowerShell shim.

Closes #924

### DIFF
--- a/internal/assets/gga/gga.cmd
+++ b/internal/assets/gga/gga.cmd
@@ -1,0 +1,5 @@
+@echo off
+setlocal
+set "GGA_PS1=%~dp0gga.ps1"
+powershell -NoProfile -ExecutionPolicy Bypass -File "%GGA_PS1%" %*
+exit /b %ERRORLEVEL%
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
@@ -40,9 +41,9 @@ type DoctorReport struct {
 var knownTools = []string{"gentle-ai", "engram", "gga", "claude", "opencode"}
 
 const (
-	engramHealthEnvVar    = "ENGRAM_BASE_URL"
-	diskWarnThreshold     = int64(100 * 1024 * 1024) // 100 MB
-	diskFailThreshold     = int64(10 * 1024 * 1024)  // 10 MB
+	engramHealthEnvVar = "ENGRAM_BASE_URL"
+	diskWarnThreshold  = int64(100 * 1024 * 1024) // 100 MB
+	diskFailThreshold  = int64(10 * 1024 * 1024)  // 10 MB
 )
 
 // Overridable for testing.
@@ -50,6 +51,7 @@ var (
 	lookPathFn          = exec.LookPath
 	availableBytesFn    = storage.AvailableBytes
 	osUserHomeDirDoctor = os.UserHomeDir
+	executableExtsFn    = executableExtensions
 	pathDirsFn          = func() []string {
 		return filepath.SplitList(os.Getenv("PATH"))
 	}
@@ -102,8 +104,8 @@ func checkOneTool(tool string, pathDirs []string) CheckResult {
 
 	var copies []string
 	for _, dir := range pathDirs {
-		if _, statErr := os.Stat(filepath.Join(dir, tool)); statErr == nil {
-			copies = append(copies, filepath.Join(dir, tool))
+		if p := toolInDir(dir, tool); p != "" {
+			copies = append(copies, p)
 		}
 	}
 
@@ -121,6 +123,44 @@ func checkOneTool(tool string, pathDirs []string) CheckResult {
 		Status: CheckStatusPass,
 		Detail: tool + " found at " + resolved,
 	}
+}
+
+// executableExtensions returns the filename suffixes to probe when scanning a
+// PATH directory for a tool binary. On Windows it mirrors exec.LookPath, which
+// resolves a bare name like "gentle-ai" to "gentle-ai.exe"/".cmd" via PATHEXT;
+// on other platforms the bare name is used as-is. Without this, the duplicate
+// scan never matches real Windows binaries and PATH shadowing goes unreported.
+func executableExtensions() []string {
+	if runtime.GOOS != "windows" {
+		return []string{""}
+	}
+	pathext := os.Getenv("PATHEXT")
+	if pathext == "" {
+		pathext = ".COM;.EXE;.BAT;.CMD"
+	}
+	var exts []string
+	for _, e := range strings.Split(pathext, ";") {
+		if e = strings.TrimSpace(e); e != "" {
+			exts = append(exts, strings.ToLower(e))
+		}
+	}
+	if len(exts) == 0 {
+		return []string{".exe"}
+	}
+	return exts
+}
+
+// toolInDir returns the path to tool's executable inside dir, or "" if absent.
+// It honors Windows executable extensions so the duplicate scan agrees with
+// exec.LookPath (used for the resolved path).
+func toolInDir(dir, tool string) string {
+	for _, ext := range executableExtsFn() {
+		candidate := filepath.Join(dir, tool+ext)
+		if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
+			return candidate
+		}
+	}
+	return ""
 }
 
 // checkStateJSON validates ~/.gentle-ai/state.json and agent config dirs.
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -131,21 +131,26 @@ func checkOneTool(tool string, pathDirs []string) CheckResult {
 // on other platforms the bare name is used as-is. Without this, the duplicate
 // scan never matches real Windows binaries and PATH shadowing goes unreported.
 func executableExtensions() []string {
-	if runtime.GOOS != "windows" {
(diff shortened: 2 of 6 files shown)


Alan-TheGentleman (approved):
Scoped Windows doctor fix. The branch now preserves PATHEXT-aware shadow detection and adds the GGA shim behavior needed for #924, including `.ps1` fallback detection and same-directory shim deduplication. Approved pending CI.

PR #956, opened 2026-06-25
fix(sdd): propagate top-level permissions to orchestrator agents

### LINKED ISSUE #787: fix(opencode): orchestrator agent permission override bypasses top-level sensitive-file read denies
### Pre-flight Checklist

### 📝 Bug Description
Gentle AI-generated OpenCode orchestrator agents define an agent-level `permission` object with only `task` rules. In OpenCode, per-agent `permission` overrides the top-level `permission` object entirely, so top-level `read` denies for sensitive files like `.env` and `conexion.php` stop applying to those agents. As a result, the orchestrator can still read files that the config appears to deny.

### 🔄 Steps to Reproduce
1. Use a Gentle AI OpenCode profile that defines top-level sensitive-file denies, for example `permission.read[**/.env] = deny` and `permission.read[**/conexion.php] = deny`.
2. Make sure an orchestrator agent also has its own `agent.<name>.permission` block with only `task` rules, for example `gentle-orchestrator` or `sdd-orchestrator-chatgpt-juicio`.
3. Start OpenCode with that profile.
4. Ask the orchestrator to read `.env` or `conexion.php`.
5. Observe that the read succeeds even though the top-level config says those paths are denied.

### ✅ Expected Behavior
Top-level sensitive-file `read` denies should still protect orchestrator agents, or Gentle AI should generate agent-level permission blocks that explicitly carry those `read` denies forward. A user should not be able to think `.env` and `conexion.php` are protected while the orchestrator can still read them.

### ❌ Actual Behavior
The orchestrator agent can read `.env` and `conexion.php` successfully. The effective config is misleading because the top-level deny looks correct, but the agent-level `permission` override silently drops those `read` rules.

### Gentle AI Version
gga v2.8.0

### Operating System
Windows

### AI Agent / Client
OpenCode

### 📋 Affected Area
Other

### 💡 Logs / Error Output
```shell
Top-level config:
  permission.read:
    *: allow
    **/.env: deny
    **/conexion.php: deny

Agent config:
  agent.gentle-orchestrator.permission:
    task:
      *: deny
      ...allowed subagents...

Observed behavior:
- read(.env) succeeded
- read(conexion.php) succeeded

Relevant OpenCode schema/doc behavior:
- per-agent permission overrides top-level permission
```

### Additional Context
I verified there was no later `allow` rule overriding the deny. The root cause appears to be the generated Gentle AI profile shape itself: agent-level `permission` is introduced for task routing, but the sensitive-file `read` deny is not duplicated into those same agents. This affects at least:
- `gentle-orchestrator`
- `sdd-orchestrator-chatgpt-juicio`
- `sdd-orchestrator-sin-claude-con-go`

### Verified Working Mitigation
I confirmed a working fix locally: duplicate the top-level `permission.bash` and `permission.read` blocks into every affected orchestrator agent that already defines its own `permission` object. In my case, adding those blocks to `gentle-orchestrator`, `sdd-orchestrator-chatgpt-juicio`, and `sdd-orchestrator-sin-claude-con-go` closed the gap. After restarting OpenCode, the orchestrators should inherit the intended sensitive-file restrictions because they now define them explicitly instead of shadowing them.

### Suggested Product Fix
1. When Gentle AI generates agent-level `permission`, merge or duplicate the top-level sensitive-file `read` deny into those agents.
2. Do the same for any related `bash` deny rules that protect secret files from shell-based reads.
3. Or stop generating partial agent-level `permission` objects that unintentionally shadow top-level protections.
4. Add an automated regression test that proves `.env` and `conexion.php` remain unreadable to orchestrators when the profile claims they are denied.

decode2:
Closes #787

---

## PR Type

- [x] `type:bug`

## Summary

- Propagates top-level OpenCode permission rules into orchestrator agents without weakening the effective ordered last-match policy.
- Preserves agent-specific restrictions, scalar `ask` fallbacks, task replacement entries, and untouched permission categories.
- Treats exact and glob overlaps conservatively: unproven coverage fails closed, later permissive rules cannot erase an earlier deny/ask, and restoration obligations are computed before emission.
- Keeps JSON and JSONC output byte-idempotent after one production injection cycle.
- Retains the current-main Codex residual-profile cleanup independently from the OpenCode path.

## Changes

| Area | Change |
|---|---|
| `internal/components/sdd/inject.go` | Provenance-aware ordered policy composition and fail-closed propagation |
| `internal/components/permissions/*` | Permission injection integration and Codex cleanup compatibility |
| focused tests | Base/head differential, ordered overlaps, serializer round trips, production idempotence, and property coverage |

## Test Plan

- [x] Frozen permission regression matrix x20
- [x] Property seed `9562026` across 20,000 compositions
- [x] Prior-head differential probes for deny/ask weakening and second-cycle drift
- [x] Affected-package tests and race detector
- [x] `go test ./... -count=1`
- [x] `go vet ./...`
- [x] Windows all-package compile-only validation
- [x] `git diff --check`

## Review Budget

The final security regression matrix and production-path coverage make this PR larger than 400 changed lines. Maintainer-approved `size:exception` is applied because splitting the ordered composition algorithm from the tests that prove it would weaken review safety.

## Notes

The branch contains current main `9c7c371bfea9ba522c37a150615b5ab4e78d3e13` as an ancestor. The exact final head is `eca96865636d9bff694bd0a2b5310f44e2f30541`.


decode2:
@Alan-TheGentleman I've addressed the CodeRabbit review feedback: merged permissions are now propagated lazily, scalar top-level values won't clobber pre-existing agent permission maps, and top-level denies override agent-level allows for matching globs. I also extended TestMergeJSONFile_PropagatesPermissions and added TestMergeJSONFile_NoPropagatesIfNoKeys to verify these semantics.

Alan-TheGentleman (changes requested):
The propagation logic is a real improvement, but as shipped it does not close #787 on the default install path, and I reproduced the bypass empirically on this branch. In the full-gentleman preset (and in sync --include-permissions) ComponentSDD runs before ComponentPermission, so gentle-orchestrator is generated while no top-level permission block exists yet; propagateTopLevelPermissions bails out, permissions.Inject writes the denies afterward, and nothing backfills the orchestrator. Final state on a fresh default install: top-level permission.read has the .env deny, gentle-orchestrator.permission has no read key at all. The PR's unit tests pass because they seed the top-level block before the merge, which is not the real component order.

Blocking: either reorder Permission before SDD in every component list (tui model.go componentsForPreset, cli validate.go, cli sync.go BuildSyncSelection) or make propagation idempotent and re-run it whenever the top-level permission block changes, and add a regression test that exercises the actual installer order (sdd.Inject then permissions.Inject) asserting the orchestrator carries the deny.

Also worth fixing while you are in there: propagation is hardcoded to read/bash/write, so any other top-level category never propagates, and a top-level scalar allow can overwrite an agent's own explicit deny, which is the inverse of the invariant #787 asks for.

This stays the canonical PR for #787. Push the update and I'll re-review with priority.

decode2:
The permission work was split into two reviewable units: propagation and safety first, then Permission/SDD ordering and backfill integration. The first slice was limited to 400 changed lines and required an approved issue-linkage approach before opening its PR.

decode2:
@Alan-TheGentleman The canonical PR branch is now updated at `1a081ad` (with `ffffbf2` preserved as its direct predecessor). Review found no defects.

Delivery mapping:
- Real install order: the regression test runs `sdd.Inject` first, then `permissions.Inject`, matching the default path that previously left orchestrators unprotected.
- Permission propagation: all top-level categories are merged into `gentle-orchestrator` and `sdd-orchestrator-*` agents.
- Precedence: restrictive rules win (`deny > ask > allow/default`), including scalar-vs-map and exact-glob collisions, without weakening existing agent denies.
- Backfill: Permission injection re-applies propagation after the top-level block changes, closing the real-order gap.
- Tests: focused real-order/precedence tests passed 20 consecutive runs; `go build ./...` and `git diff --check upstream/main...HEAD` passed.
- Scope: exactly 4 files, 378 changed lines against current `upstream/main`.

Please re-review this published HEAD, thanks!

dnlrsls (changes requested):
Requesting changes for two permission-precedence issues:

1. `internal/components/sdd/inject.go:1782-1792`: the `*: ask` branch copies restrictive top-level rules but retains conflicting agent-specific allows. For example, a propagated `*: deny` can coexist with `secret.env: allow`; OpenCode uses last-matching-rule precedence, so the specific allow still wins and the global deny is ineffective.

2. `internal/components/sdd/inject.go:1731-1754,1887-1895`: propagation normalizes and writes back the top-level policy itself. With the stock read policy (`*: allow` plus secret-pattern denies), overlap normalization can promote `*` to `deny`, unexpectedly denying all reads instead of only sensitive paths.

Please keep top-level permissions immutable while deriving the agent policy, resolve overlapping agent allows according to effective OpenCode precedence, and add regressions for both cases. The current tests verify map contents, but they do not cover effective last-match behavior or assert that the top-level `*: allow` remains unchanged.

decode2:
Updated in `cbbd9d89`. Global OpenCode `deny`/`ask` restrictions can no longer be weakened by agent-specific allows, caller-owned top-level permission structures remain unchanged, and the permissions-first then SDD-injection lifecycle is covered.

The full CI matrix is green. This is ready for re-review.

dnlrsls:
A focused patch is available against this PR's source branch: https://github.com/decode2/gentle-ai/pull/1

It preserves OpenCode's effective last-match permission precedence and top-level JSON rule order while keeping sensitive denies such as `**/.env` effective. The patch is limited to two files, includes focused provider compatibility tests, and passed the mandatory bounded review.

@decode2, please review the patch when you have a chance.

Alan-TheGentleman:
Permission resolution needs one documented precedence table across every supported scope and `allow`, `ask`, or `deny` combination. Implement that table in one canonical resolver and add table-driven conflicts that prove lower-priority values cannot override higher-priority policy. Update from current main, rerun the permission and golden suites, and run fresh full CI.

decode2:
Addressed maintainer review (@Alan-TheGentleman).

- Documented **Canonical Action Precedence Table** in `internal/components/permissions/inject.go` establishing the explicit hierarchy across all scopes and action combinations:

  762182\text{deny (3)} > \text{ask (2)} > \text{allow / default (1)}762182

- Exported `CanonicalActionRank` helper function.
- Added `TestCanonicalPermissionPrecedenceTableDriven` test suite in `permissions/inject_test.go` verifying table-driven conflicts and proving lower-priority values cannot override higher-priority policy.
- Rebased clean onto current `upstream/main` (`6a5af7eb`).
- Ran complete permission and sdd golden suites (PASS).
- Published head commit `2fcd8b25` to `fix/issue-787-opencode-permission-bypass`.

Ready for fresh maintainer review.

decode2:
Re-review requested at `fb26b46f`.

Production permission composition already uses the single `sdd.permissionRank` solver helper, so the unused exported duplicate and its self-only test were removed. Focused verification passed: `go test ./internal/components/permissions ./internal/components/sdd` and `bash scripts/deadcode-ratchet.sh` (`no new unreachable functions`).

dnlrsls:
Superseded by #2800. Current OpenCode 1.18.15 already inherits global categories for partial agent permission maps; the remaining bypass comes from deprecated 	ools compatibility conversion. The replacement removes that managed emission, cleans stale managed entries, preserves user-owned agents, and has a green full CI matrix. Closing this stale implementation to keep review focused on the current root.

PR #961, opened 2026-06-26
fix(install): programmatically clean gga clone dir to avoid collision

### LINKED ISSUE #31: Error reinstalling agent: existing temp directory causes git clone failure + backups menu select not working
Hola! esta la primera issue que hago. Es un placer que seas vos locura mix..

Estoy intentando reinstalar el agente usando el comando:
 `irm https://raw.githubusercontent.com/Gentleman-Programming/gentle-ai/main/scripts/install.ps1 | iex` 
Lo ejecuto porque quiero mantener la versión actualizada.

Durante la instalación aparece el siguiente error:
```
  FAILED: component:gga — run command "git clone https://github.com/Gentleman-Programming/gentleman-guardian-angel.git C:\\Users\\user\\AppData\\Local\\Temp\\gentleman-guardian-angel": exit status 128
  output:
  fatal: destination path 'C:\Users\user\AppData\Local\Temp\gentleman-guardian-angel' already exists and is not an empty directory.
  pipeline completed with errors
```
Parece que falla porque el directorio temporal ya existe y no está vacío.

 Tambien, en el menú **Manage Backups**, al presionar **Enter** en **Select**, no pasa nada.

Sería útil tener también una opción para **ver o borrar backups** desde ese menú.

decode2:
## Summary

Fixes the GGA agent installation collision issue where \git clone\ fails because the target temporary directory \gentleman-guardian-angel\ already exists and is not empty from a previous install attempt.

## Problem

When reinstalling or upgrading, the installer clones GGA into the temporary directory \gentleman-guardian-angel\. If a previous installation attempt left files in that directory, \git clone\ fails with exit code 128:
\\\
fatal: destination path 'C:\Users\user\AppData\Local\Temp\gentleman-guardian-angel' already exists and is not an empty directory.
\\\`n
## Solution

Programmatically clean the destination directory before cloning by calling \os.RemoveAll()\ inside \
esolveGGAInstall\ at resolution time. This ensures the target directory is completely gone before the clone starts, while maintaining compatibility with the existing shell cleanup commands asserted in the unit tests.

Closes #31

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows install cleanup behavior to fail fast on PowerShell errors and to remove the stale clone directory only when it actually exists, reducing leftover install artifacts.
* **Tests**
  * Added an end-to-end test that creates a temporary stale clone, verifies cleanup deletes it, and confirms cleanup succeeds when the directory is already gone.

### DIFF
--- a/internal/installcmd/resolver.go
+++ b/internal/installcmd/resolver.go
@@ -262,7 +262,6 @@ func resolveGGAInstall(profile system.PlatformProfile) (CommandSequence, error) 
 		}, nil
 	case "apt", "pacman", "dnf":
 		const tmpDir = "/tmp/gentleman-guardian-angel"
-		_ = os.RemoveAll(tmpDir)
 		return CommandSequence{
 			{"rm", "-rf", tmpDir},
 			{"git", "clone", "https://github.com/Gentleman-Programming/gentleman-guardian-angel.git", tmpDir},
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman (changes requested):
The intent is right but I don't think this fixes #31 as shipped. The Windows branch already had a shell-level cleanup with -ErrorAction SilentlyContinue when the issue was filed, so the likely root cause is a removal that fails silently (locked file, AV handle, permissions), and this adds a second cleanup that fails silently in exactly the same way: the os.RemoveAll error is discarded. If removal fails, git clone reproduces the identical error.

Three items before this can move: surface or handle the RemoveAll error instead of discarding it (return it, log it, or retry), add a behavior-level test that creates a stale non-empty gentleman-guardian-angel directory and asserts it is actually gone after resolution (today the tests only compare the returned command strings), and reconsider doing the deletion inside resolveGGAInstall. That function is a pure resolver called from tests and any future dry-run path; deleting at resolution time is an undocumented side effect, and the existing table test now touches real OS temp dirs on every run. Executing the cleanup as part of the returned command sequence, or at execution time with an injectable path, keeps the resolver pure and the tests hermetic.

Push the update and I'll re-review.

decode2:
The PowerShell expectation and platform-detection fixes were prepared and validated. PR #961 was later closed without merge and superseded by #1195.

decode2:
The proposed fix updates the PowerShell expectation and uses `runtime.GOOS` for platform detection. PR #961 was later closed without merge and superseded by #1195.

Alan-TheGentleman:
Superseded by #1195. The replacement preserves both of your commits and authorship, applies the exact pending CI fixes, and adds PowerShell path escaping found during maintainer review. Thank you for diagnosing the stale-directory failure and for keeping the resolver pure.

PR #978, opened 2026-06-27
feat(hermes): honor HERMES_HOME in adapter ConfigPath

Cristobalar-1982:
Make gentle-ai use the same Hermes home directory Hermes Desktop uses by honoring the HERMES_HOME environment variable in the Hermes adapter.

On Windows, Hermes Desktop stores its config under the Hermes-generated path from get_hermes_home(), which typically resolves to %LOCALAPPDATA%\hermes. Until now, gentle-ai hardcoded ~/.hermes and ignored that, so users had to keep the paths in sync manually.

With this change, ConfigPath() checks HERMES_HOME first. When it is set, all gentle-ai paths for Hermes (SOUL.md, skills, MCP config) resolve to the same directory Hermes already uses. When HERMES_HOME is unset, the historical default (~/.hermes) is preserved, so behavior is unchanged for CLI-only workflows.

Files changed
- internal/agents/hermes/adapter.go: add HERMES_HOME override to ConfigPath()
- internal/agents/hermes/adapter_test.go: cover the new behavior while isolating existing tests from the environment

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration now respects the `HERMES_HOME` environment variable when it is set, while keeping the default location unchanged when it is not.
  * Test setup was tightened to avoid environment-related interference between runs.
* **Tests**
  * Added coverage for configuration path behavior with `HERMES_HOME` set and unset.

### DIFF
--- a/internal/agents/hermes/adapter.go
+++ b/internal/agents/hermes/adapter.go
@@ -161,8 +161,16 @@ func defaultStat(path string) statResult {
 	return statResult{isDir: info.IsDir()}
 }
 
-// ConfigPath returns the path to ~/.hermes, the Hermes global config directory.
+// ConfigPath returns the Hermes global config directory. It honors the
+// HERMES_HOME environment variable when set (the same override Hermes Desktop
+// itself uses), so that gentle-ai and Hermes agree on where persona/skills/MCP
+// config live. When HERMES_HOME is unset, the historical default of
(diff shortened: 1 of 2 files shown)


Alan-TheGentleman:
Closing because it skips the issue-first process. No PR merges without a `status:approved` issue linked. If HERMES_HOME support solves a real setup for you, open an issue with the use case and re-file after approval.

Issue #941, opened 2026-06-20
fix(codex): /nix/store read rule breaks Codex config load on Windows (CLI + desktop sandbox)


AlbertGC13:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

### 📝 Bug Description

The Codex permissions injector unconditionally writes a `"/nix/store" = "read"` rule into the `gentle-dev` profile of `~/.codex/config.toml`, regardless of the operating system.

On Windows, `/nix/store` is not a valid Codex filesystem path (it is not Windows-absolute, does not start with `~/`, and does not start with `:`). Codex therefore **fails to load the entire configuration**, which breaks **both the Codex CLI and the Codex desktop app's Agent sandbox**.

Root cause — `internal/components/permissions/inject.go`, function `injectCodexPermissions` (line ~195):

```go
for _, path := range []string{
    `":minimal"`,
    `"~/.config/git"`,
    `"~/.gitconfig"`,
    `"~/.local/state/nix/profiles/home-manager/home-path"`,
    `"~/.nix-profile"`,
    `"/nix/store"`,   // ← written on every OS, including Windows
} {
    merged = filemerge.UpsertTOMLTableKey(merged, "permissions.gentle-dev.filesystem", path, `"read"`)
}
```

There is no `runtime.GOOS` guard (nor any Nix-presence detection) around the Nix-specific entries. `/nix/store` is meaningful only on Nix-capable systems (Linux/macOS with Nix); on Windows it is invalid and fatal.

The other two Nix entries (`~/.nix-profile`, `~/.local/...`) are tolerated only because they start with `~/`; the bare `/nix/store` is the one that aborts config parsing.

This appears to have been introduced by the Nix support work (#141).

### 🔄 Steps to Reproduce

1. On Windows, install or `sync`/update gentle-ai with the Codex permissions component enabled (this regenerates `~/.codex/config.toml`).
2. Open a terminal and run `codex`.
3. Observe the CLI aborts before starting with: `Error loading configuration: filesystem path /nix/store must be absolute, use ~/..., or start with :`.
[...]

PR #1023, opened 2026-07-07
fix(codex): only inject Nix read paths when Nix is installed

### LINKED ISSUE #1022: fix(codex): Nix read paths injected without existence check break Codex sandbox on non-Nix Linux/WSL
### Pre-flight Checks

### Bug Description
`injectCodexPermissions` (`internal/components/permissions/inject.go`) writes three Nix read paths into the generated `[permissions.gentle-dev.filesystem]` profile **unconditionally**:

- `/nix/store`
- `~/.nix-profile`
- `~/.local/state/nix/profiles/home-manager/home-path`

On a machine **without Nix installed**, none of these exist. Codex enforces the `[permissions]` profile through a **bubblewrap** sandbox, and binding a non-existent path makes bwrap create a synthetic read-only tmpfs placeholder (shows as `.../deleted`). Any access then fails with `Device or resource busy`, and the Codex sandbox breaks — blocking work intermittently.

This is the same root cause as #941 (`/nix/store` breaking Codex on Windows), but that fix only guarded `/nix/store` behind `if codexPermissionsGOOS != "windows"`. The other two Nix paths are still injected everywhere, and **Linux/WSL users without Nix are not covered at all**.

### Steps to Reproduce
1. On a Linux/WSL machine with **no Nix installed** (`/nix/store`, `~/.nix-profile`, `~/.local/state/nix` all absent).
2. Install/sync gentle-ai with Codex selected, so it writes the `gentle-dev` profile into `~/.codex/config.toml`.
3. Run Codex (which applies the `gentle-dev` permissions profile via bubblewrap).
4. Observe the sandbox failing on `~/.local/state/nix`.

### Expected Behavior
Nix read paths should only be injected when Nix is actually present. A user without Nix should get a `gentle-dev` profile with no dangling Nix binds, and Codex should run cleanly.

### Actual Behavior
The profile always contains the three Nix paths. bwrap binds non-existent paths as synthetic `.../deleted` tmpfs files, and access fails:

```
tmpfs /bindfile.../deleted -> /home/<user>/.local/state/nix
Device or resource busy
```

### Operating System
WSL

### Agent / Client
Codex

### Shell
zsh

### Relevant Logs
```
/nix/store                     NO EXISTE
~/.nix-profile                 NO EXISTE
~/.local/state/nix             NO EXISTE
```

### Additional Context
Proposed fix: replace the OS-based guard with a filesystem **existence check** — only append each Nix read path when it resolves on disk. This subsumes the Windows guard from #941 and fixes non-Nix Linux/WSL in one change. PR to follow.

lu149e:
## Summary

The Codex permission injector now writes the three Nix read paths into the generated `[permissions.gentle-dev.filesystem]` profile **only when they actually resolve on disk**. On hosts without Nix, none of these paths are injected.

## Root cause

`injectCodexPermissions` wrote `/nix/store`, `~/.nix-profile`, and `~/.local/state/nix/profiles/home-manager/home-path` into the profile unconditionally. Codex enforces the profile through a bubblewrap sandbox, and binding a **non-existent** path makes bubblewrap create a synthetic `.../deleted` read-only tmpfs entry whose access then fails with `Device or resource busy` (EBUSY). This breaks Codex on non-Nix Linux/WSL.

The prior fix (#941) only guarded `/nix/store` behind a `GOOS != "windows"` check, leaving `~/.nix-profile` and the home-manager path unconditionally injected — so the sandbox still broke on non-Nix Linux/WSL, and `~/.nix-profile`/home-manager paths still loaded on Windows.

## Fix

- Replaced the OS-based guard with a filesystem **existence check** (`os.Stat`, via a mockable `osStat` indirection). Each Nix read path is appended only when `osStat(resolved)` returns no error.
- `~` is resolved to `homeDir` via `filepath.Join`; `/nix/store` is checked directly.
- Deterministic slice order is preserved (home-manager path, then `~/.nix-profile`, then `/nix/store`) — no map is used, so generated TOML stays stable and idempotent.
- Removed the now-unused `codexPermissionsGOOS`/`runtime` OS guard entirely; the existence check subsumes the Windows special-case and covers every non-Nix host.

This supersedes the partial #941 Windows guard.

## Testing

Run from the repo root:

- `gofmt -l internal/components/permissions/inject.go internal/components/permissions/inject_test.go` — no diff
- `go vet ./internal/components/permissions/...` — passes
- `go build ./...` — passes
- `go test ./internal/components/permissions/...` — `ok` (passes)

Test changes:
- Updated `TestInjectCodexWritesGentleDevPermissionsProfile` to mock `osStat` as present (via a `nixPathsPresent` helper) so the Nix-path assertions no longer depend on the host filesystem.
- Replaced the Windows-guard test with `TestInjectCodexPermissionsSkipsNixPathsWhenAbsent`, which proves that when `osStat` reports the Nix paths as absent, the generated TOML contains **none** of the three Nix read paths (while base read paths remain).
- Added `TestInjectCodexPermissionsIncludesNixPathsWhenPresent`, which proves all three Nix read paths **are** included when present.

## Notes

Fixes #1022. Supersedes the partial `/nix/store` Windows guard from #941.

Opened as a **draft** pending `status:approved` per the repository's contribution workflow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission generation so Nix-related filesystem read entries are included only when the corresponding paths exist locally.
  * Removed OS-specific Nix path handling for more consistent permissions across environments.

* **Tests**
  * Updated permission tests to be deterministic and not depend on the machine’s filesystem state.
  * Added scenarios verifying Nix read paths are omitted when missing and included when present.

### DIFF
--- a/internal/components/permissions/inject.go
+++ b/internal/components/permissions/inject.go
@@ -188,11 +188,6 @@ func injectCodexPermissions(homeDir string, adapter agents.Adapter) (InjectionRe
 	merged = filemerge.UpsertTOMLTableKey(merged, "permissions.gentle-dev.network.domains", `"*"`, `"allow"`)
 
 	merged = filemerge.RemoveTOMLTableKeys(merged, `permissions.gentle-dev.filesystem.":root"`, []string{`"."`})
-	readPaths := []string{
-		`":minimal"`,
-		`"~/.config/git"`,
-		`"~/.gitconfig"`,
-	}
 	// Only inject Nix read paths that actually resolve on disk. Codex enforces the
 	// profile through a bubblewrap sandbox, and binding a non-existent path makes
 	// bubblewrap create a synthetic ".../deleted" read-only tmpfs entry whose access
@@ -208,6 +203,17 @@ func injectCodexPermissions(homeDir string, adapter agents.Adapter) (InjectionRe
 		{`"~/.local/state/nix/profiles/home-manager/home-path"`, filepath.Join(homeDir, ".local", "state", "nix", "profiles", "home-manager", "home-path")},
 		{`"~/.nix-profile"`, filepath.Join(homeDir, ".nix-profile")},
 		{`"/nix/store"`, "/nix/store"},
+	}
+	nixReadPathKeys := make([]string, 0, len(nixReadPaths))
+	for _, nix := range nixReadPaths {
+		nixReadPathKeys = append(nixReadPathKeys, nix.key)
+	}
+	merged = filemerge.RemoveTOMLTableKeys(merged, "permissions.gentle-dev.filesystem", nixReadPathKeys)
+
+	readPaths := []string{
+		`":minimal"`,
+		`"~/.config/git"`,
+		`"~/.gitconfig"`,
 	}
 	for _, nix := range nixReadPaths {
 		if _, err := osStat(nix.resolved); err == nil {
--- a/internal/components/permissions/inject.go
+++ b/internal/components/permissions/inject.go
@@ -3,14 +3,12 @@ package permissions
 import (
 	"fmt"
 	"os"
-	"runtime"
+	"path/filepath"
 
 	"github.com/gentleman-programming/gentle-ai/internal/agents"
 	"github.com/gentleman-programming/gentle-ai/internal/components/filemerge"
 	"github.com/gentleman-programming/gentle-ai/internal/model"
 )
-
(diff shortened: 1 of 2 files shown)


lu149e:
Superseded by #1416, which removed the Codex `gentle-dev` profile and all injected Nix bind paths. That change is on current `main`, and #1022 is already closed as resolved, so this implementation is no longer needed. Closing without merge.

PR #1034, opened 2026-07-08
fix(sdd): extract model-capability section in orchestrator injectors and unify artifact-store default

AlejandroTatum:
## Summary
- `injectFileAppend` and `injectMarkdownSections` wrote the raw embedded `sdd-orchestrator` asset without ever calling `extractModelSection`, so templates containing both `<!-- section:model-capable -->` and `<!-- section:model-small -->` blocks (e.g. `generic/sdd-orchestrator.md`) got installed with **both variants concatenated into one self-contradictory prompt**. Both functions now extract the section matching the install's capability, defaulting to `"capable"` when none is specified — the same safe-default pattern already used in `prompts.go` and `profiles.go`.
- Unifies the artifact-store default across all 12 per-agent SDD orchestrator templates: several declared `engram` as the default while the policy section and the native OpenSpec dispatcher assume `openspec`. All templates now consistently default to `openspec`, with `engram` documented as opt-in.
- Regenerated golden fixtures to match.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (full suite green, including regenerated goldens)
- [x] Manually diffed each regenerated `.golden` file to confirm the change is scoped to the artifact-store text only

## Summary by CodeRabbit

* **New Features**
  * Default artifact-store guidance now prefers `openspec` when no choice is specified.
  * `engram` is now used only when explicitly requested for file-skipping workflows.
* **Documentation**
  * Updated SDD guidance across supported environments to match the new default selection and fallback behavior.
  * Clarified recommendations for when to use each artifact-store option (including updated “none” guidance).
* **Bug Fixes**
  * Improved capability-specific SDD orchestrator content selection so sessions reflect the appropriate “capable” vs “small” guidance.


Alan-TheGentleman:
Closing this because it skips the issue-first process. No PR merges without a `status:approved` issue linked. The extraction idea itself looks reasonable: open an issue describing the drift problem, get it approved, then re-file with the link. This isn't a rejection of the concept.

PR #1077, opened 2026-07-09
fix(filemerge): preserve symlinked config files

### LINKED ISSUE #306: Error in the sync/uninstall commands for when using symlinks for dotfiles
### Pre-flight Checklist

### 📝 Bug Description

when trying to run sync or uninstall gives error in the symlinks

### 🔄 Steps to Reproduce

1. To have the connection by means of symbolic inserts, for example using stow to manage dotfiles

### ✅ Expected Behavior

Modify correctly

### ❌ Actual Behavior

Uninstall Result

✗ Uninstall failed

Error:
  read file "/Users/davidgarcia/.claude/CLAUDE.md": refusing to read symlink "/Users/davidgarcia/.claude/CLAUDE.md"

Backup created before failure: 20260414192011.899281000
uninstall — 2026-04-14 21:20 (230 files)

enter: return • esc: back • q: quit

### Gentle AI Version

v2.8.1

### Operating System

macOS

### AI Agent / Client

OpenCode

### 📋 Affected Area

CLI (commands, flags)

### 💡 Logs / Error Output

```shell
Uninstall Result

✗ Uninstall failed

Error:
  read file "/Users/davidgarcia/.claude/CLAUDE.md": refusing to read symlink "/Users/davidgarcia/.claude/CLAUDE.md"

Backup created before failure: 20260414192011.899281000
uninstall — 2026-04-14 21:20 (230 files)

enter: return • esc: back • q: quit
```

### Additional Context

_No response_

fjpalacios:
## Linked Issue
Closes #306

## PR Type
- [x] Bug fix
- [ ] New feature
- [ ] Documentation only
- [ ] Code refactoring
- [ ] Maintenance/tooling
- [ ] Breaking change

## Summary
- Resolves existing file symlinks before atomic writes so sync updates the target file without replacing the symlink.
- Allows uninstall cleaners to read symlinked managed files.
- Keeps removal conservative: deleting a symlinked managed file removes the link path, not the dotfiles target.

## Changes
| File | Change |
|------|--------|
| `internal/components/filemerge/writer.go` | Resolve existing symlinks and perform atomic replacement on the resolved target path. |
| `internal/components/filemerge/writer_test.go` | Replace symlink rejection coverage with preservation/no-op tests. |
| `internal/components/uninstall/cleaners.go` | Follow symlinks when reading managed files. |
| `internal/components/uninstall/cleaners_test.go` | Cover symlink reads and conservative symlink removal behavior. |

## Test Plan
- [x] `go test ./internal/components/filemerge ./internal/components/uninstall`
- [x] `git diff --check`
- [ ] `go test ./...` — currently fails in `internal/cli` on this machine because `TestEngramPathGuidanceDefault` expects `go/bin`, while Go is installed under asdf (`/home/fjpalacios/.asdf/installs/golang/1.25.5/bin`). Packages touched by this PR pass.

## Follow-up
The interactive/configurable uninstall symlink policy is intentionally deferred to #1078 so this bugfix stays under the 400-line review budget.

## Notes
Maintainer note: I cannot add `type:bug` from this fork, so the PR still needs exactly one `type:*` label added by someone with label permissions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Atomic file writes now safely follow existing symlink destinations while preserving links and updating intended targets.
  * Managed-file reads and backups support safe, in-root symlinks while blocking paths that escape permitted locations.
  * Cleanup removes symlinks without deleting or modifying their targets.
  * Symlink loops, unsafe parent-directory links, and external removal targets are now rejected.

* **Tests**
  * Expanded coverage for symlink writes, reads, backups, removals, escapes, dangling targets, no-op updates, and Windows privilege handling.


fjpalacios:
Addressed the CodeRabbit feedback in `8b68794`:

- `WriteFileAtomic` now handles dangling symlinks by creating the missing target path while preserving the symlink.
- Added `TestWriteFileAtomicCreatesDanglingSymlinkTarget`.
- `readManagedFile` now preserves the `os.IsNotExist` contract for broken symlinks.
- Added `TestReadManagedFile_BrokenSymlinkReturnsNotExist`.

Validation:

- `go test ./internal/components/filemerge ./internal/components/uninstall`
- `git diff --check`

The PR remains under the 400-line review budget: 215 insertions / 16 deletions.

dnlrsls:
This candidate should not close #306 yet. It changes intentional symlink rejection into write-through/read-through behavior without enforcing the managed-root boundary identified in the issue discussion.

`resolveExistingSymlink` currently accepts absolute, relative, and dangling targets wherever they resolve, and `readManagedFile` follows them as well. A managed config path can therefore redirect comparison, atomic replacement, or uninstall inspection outside the expected agent/home subtree. The current same-directory tests do not cover that security boundary.

Please add fail-closed containment checks and regression coverage for absolute and relative escapes, chained links, loops, parent-directory symlinks, and dangling targets outside the allowed root, while preserving legitimate same-managed-tree links such as the reported Antigravity layout. Revalidate both sync/rollback and uninstall paths against current main before requesting review.

fjpalacios:
Updated the symlink handling after the review.

The risky part was not “follow symlinks” by itself, but following them without a boundary. So this now resolves symlinks only if the final target stays inside the allowed root.

Policy is intentionally small:
- paths under `$HOME` may point elsewhere under `$HOME`
- other paths stay scoped to their own directory

That keeps Stow/Antigravity-style layouts working, while rejecting escapes.

I also added regression coverage for the review cases: absolute/relative escapes, chains, loops, symlinked parents, broken links outside root, and valid HOME links.

I tried to keep the patch tight: shared logic moved to `internal/symlinkguard`, comments trimmed, duplicate test setup reduced. It still lands at `452 + / 46 -`.

Requesting an exception to the 400-line guideline here. Most of the size is security regression coverage, and I’d rather keep those cases explicit than make this smaller and easier to break later.

fjpalacios:
Pushed one more update for the latest CodeRabbit findings.

I fixed the bounded cases that still belonged to this PR:
- Windows HOME test now also sets `USERPROFILE`
- nested dangling symlink chains are resolved to the final missing target
- intermediate symlinks are preserved
- uninstall now rejects parent-directory symlink escapes

I’m leaving the TOCTOU / descriptor-safe open suggestion out of this patch on purpose.

That one is a valid hardening idea, but it is a different layer of filesystem safety: opening through verified descriptors/handles, using no-follow traversal, and validating the opened file identity. That would make the PR quite a bit larger, and this one is already above the review-size guideline because of the symlink regression coverage.

Happy to implement that as a follow-up if maintainers want this PR to also cover that threat model. For this bugfix, I kept the scope to bounded symlink resolution plus regression tests.

fjpalacios:
@coderabbitai review

Latest commit `cc8bfbbd` addresses the bounded findings from the previous review. Please re-check the remaining comments against the current diff.

dnlrsls (changes requested):
Please repair the Windows containment fixtures before merge.

`go test ./internal/components/filemerge ./internal/components/uninstall ./internal/symlinkguard -count=1` fails because the escape tests create both the managed path and the supposed outside target with `t.TempDir()`. On Windows those paths are under `%USERPROFILE%`, so the PR's HOME-wide policy correctly considers the target in bounds and the tests expect the wrong result. This affects the filemerge and uninstall escape cases.

Please isolate `HOME` and `USERPROFILE` to the managed test root, or create the target outside that controlled root, then rerun the affected package tests on Windows. The bounded resolver otherwise looks aligned with #306. The size exception appears justified by the focused security regression coverage once these tests are green.

fjpalacios:
Pushed `82d5bf7e` to fix the Windows fixture issue.

The escape tests now isolate both `HOME` and `USERPROFILE` to the managed test root before creating the outside target, so the expected rejection no longer depends on where Windows places `t.TempDir()`.

Local validation passes:

```bash
go test ./internal/components/filemerge ./internal/components/uninstall ./internal/symlinkguard -count=1
git diff --check
```

I tried to run the Windows suite as well, but I cannot trigger `Windows Full Suite` on upstream without repository admin rights, and the PR workflows are currently waiting for maintainer action. Could you approve/rerun the Windows Actions, or run this focused command on Windows?

```powershell
go test ./internal/components/filemerge ./internal/components/uninstall ./internal/symlinkguard -count=1
```

hrdi1023802:
Heads-up: this branch is red on macOS, and I pushed a fix at fjpalacios/gentle-ai#1 (targeting this branch, not `main`, since `internal/symlinkguard` only exists here).

`go test ./internal/components/filemerge/` gives 6 failures on macOS, plus 1 in `uninstall`:

```
WriteFileAtomic(symlink) error = symlink "/private/var/folders/<...>/001/linked.txt"
  resolves outside allowed root "/private/var/folders/<...>/001":
  "/var/folders/<...>/001/target.txt", want success
```

Those two paths are the same location. `/var` is a symlink to `/private/var` on macOS and `t.TempDir()` hands back the `/var` spelling, so `EnsureWithinRoot` and `AllowedRoot` compare path *spellings* where they need to compare *filesystem identity*. Bisected — with `TMPDIR` under `$HOME` (no symlinked ancestor) everything passes, with the default `TMPDIR` it does not.

It reaches past the fixtures, since the guard is shared by sync, install and uninstall: `/tmp` and `/var` on macOS (so `--scope workspace` too), and Linux layouts where the home ancestor is a link.

The fix resolves the deepest existing ancestor before comparing, keeps the lexical check first so nothing existing changes, and only resolves the *existing* prefix so your dangling-target cases still work. 31 lines of production code, plus the first tests for the package — including one that proves an escape through a symlinked ancestor is still rejected.

Separately, and independent of the above: the symlink refusal this PR removes is confirmed still present in released 2.4.0, and the end-to-end Antigravity case passes on your head once the containment fix is in. Details and evidence in Gentleman-Programming/gentle-ai#306.

Merge it, cherry-pick the diff, or ignore it — whatever gets this landed. Nice work on the PR.

hrdi1023802:
Retested as requested. Two notes first: the head has moved past `c827464b`, so this is against `753613d5`; and thanks for merging fjpalacios/gentle-ai#1.

**The reported case works.** End to end on the Antigravity shape, `sync --agent antigravity` over `<home>/.gemini/antigravity-cli/mcp_config.json -> <home>/.gemini/config/mcp_config.json`: 54 files changed, the link is preserved, and the content lands through it to the target. `uninstall` and `backup` are green too, so the backup/rollback asymmetry looks closed from here.

Three things were still red on macOS, and I opened fjpalacios/gentle-ai#2 for them. The split matters more than the count:

**One real bug.** `replaceDurably` does `path = writePath` right after resolving, so every error below that line names the resolved target rather than the destination the caller passed. `writer_landed_test.go` already asserts the correct contract — it simply cannot fail on Linux, where both spellings are identical. In production this means a user asking to write `~/.claude/settings.json` through a dotfiles link gets an error naming `~/.dotfiles/claude/settings.json`, a file they never mentioned.

**Two fixture artifacts, where your implementation is right.** Both `SafeRemovalPath` expectations are built from raw `t.TempDir()`, which returns the `/var` spelling on macOS while the implementation correctly returns the physical path. I verified this by canonicalizing only the expectations against your unmodified code — all four pass. So I changed the fixtures, not the implementation. I did try making it return the caller's spelling instead, and that breaks `TestSafeRemovalPathAllowsConfiguredRootSymlink`, which legitimately wants the resolved root. Canonical is the right contract.

For the record on regressions: the head also has failures in `internal/app`, `internal/cli`, `internal/components/engram` and `internal/reviewtransaction` on macOS. Every one reproduces identically on unmodified `753613d5`, so they predate this branch and I have not touched them. I mention it only so a red macOS run does not get attributed to this PR.

One practical note for landing it upstream: the PR is now +1210 −50 across 11 files. CONTRIBUTING caps review at 400 changed lines, so this likely needs `size:exception` from a maintainer, or a split. Worth sorting out early rather than at merge time.

The rollback-masks-apply-error problem is tracked separately in Gentleman-Programming/gentle-ai#3995, and I agree it stays out of scope here.

PR #1445, opened 2026-07-19
fix(uninstall): report backup details on failure

### LINKED ISSUE #536: fix(cli): preserve uninstall backup report when dispatch bypasses platform checks
### Pre-flight Checklist

### 📝 Bug Description
`RunArgs` handles `uninstall` in an early info-command bypass branch and then has a second `case "uninstall"` later in the normal command switch. The later branch is currently unreachable.

That matters because the later branch contains better error handling: if uninstall creates a backup manifest before failing, it renders the uninstall report so the user can see restore/safety information. The early branch just returns the error from `cli.RunUninstall`.

### 🔄 Steps to Reproduce
1. Inspect `internal/app/app.go` in `RunArgs`.
2. Notice the early `case "uninstall"` before platform detection.
3. Notice the later `case "uninstall"` in the main command switch.
4. Trigger or test an uninstall failure after backup creation.

### ✅ Expected Behavior
`uninstall` should still bypass platform detection/self-update when needed, but it should use one shared dispatch path that preserves the richer report-on-error behavior.

### ❌ Actual Behavior
The early branch makes the later branch unreachable, so backup/restore guidance can be hidden on uninstall errors.

### Gentle AI Version
`main` at `aa881f2` in local review checkout

### Operating System
Linux (Other)

### AI Agent / Client
Other

### 📋 Affected Area
CLI (commands, flags)

### 💡 Logs / Error Output
```shell
Relevant path: internal/app/app.go, RunArgs uninstall dispatch
```

### Additional Context
Related lifecycle work exists in #102 and PR #163, but this is narrower: command dispatch/reporting behavior in `internal/app/app.go`.

Suggested fix direction: keep the platform-guard bypass for uninstall, but route through a single helper that renders `RenderUninstallReport` when a manifest exists, including on error.

dnlrsls:
## 🔗 Linked Issue

Closes #536

---

## 🏷️ PR Type

---

## 📝 Summary

- Preserve uninstall backup details when dispatch bypasses platform validation.
- Route uninstall through one report-preserving helper.
- Add a regression test for failures after backup creation.

---

## 📂 Changes

|-------------|-------------|
| `internal/app/app.go` | Reuse uninstall reporting logic from the early dispatch path. |
| `internal/app/app_test.go` | Verify backup ID and path remain visible when uninstall fails. |

---

## 🧪 Test Plan

- [x] Focused package tests pass: `go test ./internal/app`
- [x] Go format passes: `go run ./internal/gofmtcheck`
- [ ] Full unit suite: `go test ./...` (environment/unrelated failures; see notes)
- [ ] E2E suite: Docker unavailable in the local environment
- [x] Runtime boundary covered through `RunArgs` with temporary filesystem state

---

## ✅ Contributor Checklist

- [x] Relevant tests pass
- [x] Go format passes
- [ ] E2E tests pass locally (Docker unavailable)
- [x] Documentation is not required for this internal behavior fix
- [x] Commits contain no `Co-Authored-By` trailers

---

## 💬 Notes for Reviewers

The full local `go test ./...` run encountered environment-sensitive and unrelated failures, including Windows privilege requirements, an existing `main` branch fixture conflict, and unrelated SDD/CodeGraph tests. The focused `internal/app` suite passes. E2E could not start because Docker is unavailable. CI should provide the authoritative full-suite result.

Please focus on whether the shared helper preserves both the existing platform-check bypass and backup recovery details on error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved uninstall handling to avoid duplicate reports.
  * Uninstall failures now clearly include backup information and the backup path when available.

* **Tests**
  * Added coverage to verify backup details are displayed when uninstalling fails due to invalid installation state.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -83,8 +83,7 @@ func RunArgs(args []string, stdout io.Writer) error {
 				_, err := cli.RunUninstallOpenCodePlugin(args[2:], stdout)
 				return err
 			}
-			_, err := cli.RunUninstall(args[1:], stdout)
-			return err
+			return runUninstall(args[1:], stdout)
 		case "skill-registry":
 			return runSkillRegistry(args[1:], stdout)
 		case "sdd-status":
@@ -243,20 +242,6 @@ func RunArgs(args []string, stdout io.Writer) error {
 
 		_, _ = fmt.Fprintln(stdout, cli.RenderSyncReport(syncResult))
 		return nil
-	case "uninstall":
-		uninstallResult, err := cli.RunUninstall(args[1:], stdout)
-		if err != nil {
-			// If a backup was created before the failure, surface it so
-			// the user can restore safely.
-			if uninstallResult.Manifest.ID != "" {
-				_, _ = fmt.Fprintln(stdout, cli.RenderUninstallReport(uninstallResult))
-			}
-			return err
-		}
-		if uninstallResult.Manifest.ID != "" {
-			_, _ = fmt.Fprintln(stdout, cli.RenderUninstallReport(uninstallResult))
-		}
-		return nil
 	case "restore":
 		return cli.RunRestore(args[1:], stdout)
 	case "doctor":
@@ -264,6 +249,14 @@ func RunArgs(args []string, stdout io.Writer) error {
 	default:
 		return fmt.Errorf("unknown command %q — run 'gentle-ai help' for available commands", args[0])
 	}
+}
+
+func runUninstall(args []string, stdout io.Writer) error {
+	result, err := cli.RunUninstall(args, stdout)
+	if result.Manifest.ID != "" {
+		_, _ = fmt.Fprintln(stdout, cli.RenderUninstallReport(result))
+	}
+	return err
 }
 
 func hasHelpFlag(args []string) bool {
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -257,6 +257,31 @@ func TestRunArgsUninstallBypassesPlatformValidation(t *testing.T) {
 	// If we got here, uninstall bypassed the platform validation.
 }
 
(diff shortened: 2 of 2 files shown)


Alan-TheGentleman (approved):
Reviewed in depth against current `main`, where the bug is fully intact. `internal/app/app.go:81-87` dispatches uninstall from the early info-command bypass branch, before `ensureCurrentOSSupported()`, and discards the `componentuninstall.Result` — so the backup manifest is lost precisely when the user needs it. The second `case "uninstall":` at `app.go:246-259`, which contains the correct report-on-error logic, is unreachable dead code, exactly as issue #536 describes.

Verified empirically rather than by reading: on the failing path the report really is recoverable, because `internal/components/uninstall/service.go` populates `result.Manifest`/`BackupPath` immediately after the snapshot succeeds and still returns that result when the later state update fails. With the fix the output carries `Backup:` and `Backup path:` alongside the error; that is the whole point, since a user whose uninstall failed mid-way needs the backup location to recover.

The test is a genuine regression test, not a restatement: keeping the PR's `app_test.go` while reverting `app.go` to `main` fails with `output missing "Backup:"`. It also exercises the real bypass path rather than the dead branch.

No success-path or output-format regression. The only other consumer of the format tests `RenderUninstallReport` directly and is unaffected, and collapsing the two branches into a single guard is cleaner than the duplication on `main` while deleting the dead code the issue flagged. Leaving the sibling `RunUninstallOpenCodePlugin` call untouched is correct, since it prints its own report internally.

One follow-up worth its own issue, out of scope here: on the failure path the user now sees the header `Managed uninstall complete` immediately before an error message. That wording is pre-existing in `internal/cli/uninstall.go`; this PR only makes it reachable. Fixing it means changing `RenderUninstallReport`'s contract.

Approving.

PR #1491, opened 2026-07-19
feat(system): add Alpine Linux support with apk package manager

### LINKED ISSUE #334: Alpine Linux Support
### Pre-flight Checklist

### 🔍 Affected Area

Installation Pipeline

### Problem
Alpine Linux is not supported because `detectLinuxDistro()` doesn't recognize it. The code in `internal/system/detect.go` only supports Ubuntu/Debian, Arch, and Fedora/RHEL.

### 📦 Proposed Solution

Modify `internal/system/detect.go` to add native Alpine support:
```go
const LinuxDistroAlpine = "alpine"
// In resolvePlatformProfile():
case LinuxDistroAlpine:
    profile.PackageManager = "apk"
    profile.Supported = true
```

### Workaround Applied

```sh
Add `ID_LIKE=debian` to `/etc/os-release`:
```

# Backup original

```
cp /etc/os-release /etc/os-release.bak
```

# Add ID_LIKE

```
echo 'ID_LIKE=debian' >> /etc/os-release
```

### 📎 Additional Context

- The TUI doesn't work in containers without a TTY, but CLI commands work fine
- After installation, add `/root/go/bin` to your PATH for the `engram` binary

### Required Dependencies
Before running `gentle-ai install`, install these packages:
```sh
apk add git bash
```
Note: `curl`, `node`, `npm`, and `go` are also required but the installer warns about them.

Cobies:
Closes #334

## Summary by CodeRabbit

* **New Features**
  * Added support for automatically detecting Alpine Linux environments.
  * Alpine Linux installations now use the `apk` package manager.
  * Alpine users can install through the standard Linux clone-and-install workflow.

---
Recommended Label: **type:feature** (to be added by maintainer)

### DIFF
--- a/internal/installcmd/resolver.go
+++ b/internal/installcmd/resolver.go
@@ -260,7 +260,7 @@ func resolveGGAInstall(profile system.PlatformProfile) (CommandSequence, error) 
 			{"brew", "tap", "Gentleman-Programming/homebrew-tap"},
 			{"brew", "reinstall", "gga"},
 		}, nil
-	case "apt", "pacman", "dnf":
+	case "apt", "pacman", "dnf", "apk":
 		const tmpDir = "/tmp/gentleman-guardian-angel"
 		return CommandSequence{
 			{"rm", "-rf", tmpDir},
--- a/internal/system/detect.go
+++ b/internal/system/detect.go
@@ -31,6 +31,7 @@ const (
 	LinuxDistroDebian  = "debian"
 	LinuxDistroArch    = "arch"
 	LinuxDistroFedora  = "fedora"
+	LinuxDistroAlpine  = "alpine"
 )
 
 type DetectionResult struct {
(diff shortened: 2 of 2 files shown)


dnlrsls (changes requested):
The Alpine detection and GGA routing are useful narrow steps, but Alpine is now reported as supported before its install path works end to end.

Please complete the `apk` path before merging:
- Add `apk` support to the OpenCode resolver, which currently rejects the profile.
- Do not invoke the GGA installer through bare `bash` without ensuring Bash and its prerequisites are available on Alpine.
- Add apk-specific dependency hints/commands for the required git, curl, and Node/npm prerequisites.

Please add Alpine regression coverage for the resolver or mocked install path, including OpenCode and GGA command generation, plus a `--dry-run` case that confirms Alpine is detected and reported as supported. The dry-run test should complement, not replace, the install-path test because dry-run returns before command resolution.

dnlrsls:
Thanks for starting the Alpine support work here. We are closing this PR in favor of #2041, which carries the same direction forward against current main and completes minimal-Alpine dependency handling (`apk`, `git`, and `bash`) together with focused regression coverage. The original detection approach is preserved in the replacement.

PR #1492, opened 2026-07-19
feat(engram): support remote engram host via ENGRAM_BASE_URL env var

### LINKED ISSUE #315: feat: support remote engram host via ENGRAM_BASE_URL env var
## Problem

`VerifyHealth()` in `internal/components/engram/verify.go` hardcodes `http://127.0.0.1:7437` as the default base URL. The function already accepts a `baseURL` parameter, but gentle-ai always passes an empty string, forcing the hardcoded default.

This makes it impossible to use gentle-ai with a remote engram instance (e.g., a shared engram server running as a Kubernetes service).

## Use case

We're deploying engram as a shared K8s service for Coder workspaces. Multiple developers connect to the same engram instance via DNS (`engram.coder-workspaces.svc.cluster.local:7437`). The MCP server itself works fine (stdio transport), but gentle-ai's health check always hits localhost and fails during setup.

## Proposed solution

Read an environment variable (e.g., `ENGRAM_BASE_URL` or `ENGRAM_HOST`) and pass it to `VerifyHealth()` when set:

```go
func VerifyHealth(ctx context.Context, baseURL string) error {
    if strings.TrimSpace(baseURL) == "" {
        baseURL = os.Getenv("ENGRAM_BASE_URL")
    }
    if strings.TrimSpace(baseURL) == "" {
        baseURL = "http://127.0.0.1:7437"
    }
    // ...
}
```

This is backwards-compatible — existing setups without the env var continue to work as before.

## Workaround

Currently we patch the generated `settings.json` with `sed` after gentle-ai runs, replacing `127.0.0.1:7437` with the remote host. This works but is fragile (breaks if gentle-ai regenerates configs).

Cobies:
Closes #315

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the health-check service URL through the `ENGRAM_BASE_URL` environment variable.
  * Retains the existing local default when no URL is configured.

### DIFF
--- a/internal/components/engram/verify.go
+++ b/internal/components/engram/verify.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
@@ -102,6 +103,9 @@ func CountVersionCallsForTest(t interface {
 }
 
 func VerifyHealth(ctx context.Context, baseURL string) error {
+	if strings.TrimSpace(baseURL) == "" {
+		baseURL = os.Getenv("ENGRAM_BASE_URL")
+	}
 	if strings.TrimSpace(baseURL) == "" {
 		baseURL = "http://127.0.0.1:7437"
 	}


dnlrsls:
This is a focused candidate for #315, but it still needs the unresolved precedence coverage before review.

Please add isolated tests proving:

1. An empty explicit URL uses `ENGRAM_BASE_URL`.
2. A non-empty explicit URL wins when the environment variable is also set.
3. An empty or whitespace-only environment value preserves the localhost fallback.
4. Test environment state cannot leak into unrelated health checks.

Use local test servers or a pure URL-resolution seam so the tests never depend on a real Engram service. Once that coverage passes against current main, this one-file change is ready for maintainer review.

matraket:
This PR appears to have been superseded by merged [PR #2516](https://github.com/Gentleman-Programming/gentle-ai/pull/2516).

PR #2516 replaced the Engram reachability mechanism rather than extending `VerifyHealth`:

- explicit `ENGRAM_BASE_URL` selects HTTP through the current [`checkEngramReachable`](https://github.com/Gentleman-Programming/gentle-ai/blob/a245f2605841fbaf68c140b584f5fb4b9a5aa4aa/internal/cli/doctor.go#L456-L465) path;
- the default path probes the persisted stdio MCP configuration;
- `VerifyHealth` and its tests were deleted as dead code in commit [`c55bb035`](https://github.com/Gentleman-Programming/gentle-ai/commit/c55bb0358327996ed282776eacd8ef4a65ad9415).

Because this PR modifies a function that no longer exists on `main`, rebasing it would not preserve the proposed implementation shape. The linked [issue #315](https://github.com/Gentleman-Programming/gentle-ai/issues/315) also appears resolved by PR #2516 and now only needs backlog-hygiene confirmation.

Suggest closing this PR as superseded by PR #2516.

PR #1508, opened 2026-07-19
fix(filemerge): support atomic writes with Windows ACLs

### LINKED ISSUE #298: test(filemerge): chmod 555 semantics differ on Windows  read-only dir test always skipped

dnlrsls:
## Linked Issue

Closes #298

---

## PR Type

- [ ] `type:feature` - New feature (non-breaking change that adds functionality)
- [ ] `type:docs` - Documentation only
- [ ] `type:refactor` - Code refactoring (no functional changes)
- [ ] `type:chore` - Build, CI, or tooling changes
- [ ] `type:breaking-change` - Breaking change (fix or feature that changes existing behavior)

---

## Summary

- Handle Windows directory permissions through native DACLs when atomic temporary-file creation is denied.
- Preserve protected/unprotected DACL state and fail closed for relevant explicit deny ACEs, including an existing target's `DELETE` denial.
- Add native Windows regression coverage and include the focused filemerge tests in the PR-time Windows CI subset.

---

## Changes

|-------------|-------------|
| `internal/components/filemerge/writer.go` | Routes atomic temporary-file creation through the platform implementation. |
| `internal/components/filemerge/writer_nonwindows.go` | Preserves the existing non-Windows behavior. |
| `internal/components/filemerge/writer_windows.go` | Adds bounded Windows DACL recovery with explicit-deny safeguards. |
| `internal/components/filemerge/writer_windows_test.go` | Covers ACL relaxation, parent denies, and existing-target `DELETE` denies. |
| `.github/workflows/ci.yml` | Runs the focused Windows filemerge tests for pull requests. |

---

## Test Plan

**Verified locally**

- [x] Focused Windows ACL tests pass.
- [x] `go test ./internal/components/filemerge`
- [x] `go vet ./internal/components/filemerge`
- [x] Linux/amd64 package cross-compilation succeeds.
- [x] `gofmt -d` reports no differences.
- [x] `git diff --check`
- [ ] Full `go test ./...` - the local Windows environment has unrelated failures involving symlink privileges, unavailable WSL, branch fixtures, and cache state; CI remains authoritative.
- [ ] E2E tests - not applicable to this filemerge-only change.

---

## Automated Checks

The standard PR validation and CI workflows should run for the current head. GitHub currently reports 565 changed lines (+562/-3), so this PR requests the maintainer-applied `size:exception` label.

## Size Exception Request

The Windows ACL implementation, its native regression suite, and the PR-time Windows CI selection form one safety boundary. Splitting them would separate write-path behavior from the tests and workflow evidence required to review it safely. The PR remains limited to issue #298 and introduces no unrelated feature scope.

---

## Contributor Checklist

- [x] PR is linked to issue #298 with `status:approved`.
- [x] PR exceeds 400 changed lines; `size:exception` is requested with the rationale above.
- [x] Exactly one `type:*` label is requested: `type:bug`.
- [x] Focused package and Windows behavior tests pass.
- [x] Go formatting passes.
- [x] Documentation is not required for this internal platform-specific correction.
- [x] The commit follows Conventional Commits.
- [x] The commit contains no `Co-Authored-By` trailer.

---

## Notes for Reviewers

The recovery runs only after `os.CreateTemp` returns `fs.ErrPermission`. It refuses to modify the parent DACL when a relevant explicit deny is present and also protects an existing destination carrying an explicit `DELETE` denial.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows atomic writes by creating the temp file in the target directory and adding safer ACL handling with retry/restore on failure.
  * Honors explicit Windows deny rules (including delete-deny cases) to avoid unintended overwrites and preserves original file contents.
* **Tests**
  * Expanded Windows coverage for ACL protection state, denied access behavior, and correct restoration after operation-stage failures.
* **Chores**
  * Broadened Windows CI “release-blocker tests” to run a wider package set and expanded the selected Windows test cases.


dnlrsls:
Maintainer action needed: please add the \	ype:bug\ label. The contributor token cannot apply labels to the upstream repository, and the PR already selects only that type in the template.

dnlrsls:
Please continue with #1508 as the implementation base. #1493 leaves the Windows path as a no-op and does not satisfy #298's native ACL test requirement.

The remaining blocker is the unresolved deny-ACE applicability issue: deniesAccessRights currently treats denies for unrelated trustees, inherit-only ACEs, and conditional ACEs as effective. Please make the check token/SID-aware, add an unrelated-SID regression, and include TestWriteFileAtomicDoesNotReplaceTargetWithExplicitDeleteDeny in the PR-time Windows regex.

No rebase is needed at this stage. Workflow approval and the type:bug label are maintainer-owned.

Alan-TheGentleman (changes requested):
This is good work, @dnlrsls, and better than this issue usually gets. Before the requests, what held up under review:

- **Atomicity is preserved.** The only production-line change is `writer.go:64` (`os.CreateTemp` → `createAtomicTemp`); the temp still lands in the target directory and the Write/Chmod/Sync/Close/Rename sequence is untouched. `writer_windows.go:82-102` retries `CreateTemp` in the same dir after relaxing and never degrades to a direct write.
- **Non-Windows is byte-identical.** `writer_nonwindows.go` is a one-line passthrough behind `//go:build !windows`, and `writer_test.go` is unchanged. Verified by diffing the PR tree against main.
- **It narrows rather than widens.** `relaxAtomicDirectoryACL` grants only `FILE_ADD_FILE|FILE_DELETE_CHILD`, to one user SID, `NO_INHERITANCE`, merged into the existing DACL, preserving `SE_DACL_PROTECTED`. That is strictly tighter than the `os.Chmod(dir, 0o755)` the Unix path already does. It uses the pseudo-token only and attempts no privilege acquisition, so it fails and surfaces rather than escalating.
- **No new dependency.** `golang.org/x/sys` is already a direct require; every symbol used exists in the pinned version, and the local `fileDeleteChild = 0x40` const is correct and genuinely absent from x/sys.

Three blocking items, all small.

**1. The test for the new logic is excluded from PR CI.** `ci.yml:76` adds the package to the release-blocker step with a `-run` filter, but `TestWriteFileAtomicIgnoresUnrelatedWindowsDeny` — which covers the deny-SID matching that is the entire point of the second commit — is not in the regex. As written it would only ever execute on push-to-main. Adding a CI change that skips the test for the code it is meant to protect defeats the purpose; please add it.

**2. Windows CI must be green before merge.** This is 219 lines of Win32 on a write path used product-wide, and CI has never run on this PR at all (only CodeRabbit, advisory). I have approved the workflows so they will run now. This one is non-negotiable given the blast radius.

**3. Please update the branch against `main`.** It is 11 commits behind, and I want Windows CI evaluating this against current main rather than a stale base.

Worth knowing about what CI still will not prove, so nobody reads a green check as more than it is: the fixtures pre-grant `GENERIC_EXECUTE`, so a real directory that also denies traverse is untested; the relax grants `FILE_ADD_FILE|FILE_DELETE_CHILD` but not `FILE_TRAVERSE|FILE_LIST_DIRECTORY`. And the fixtures build only simple `ACCESS_DENIED_ACE_TYPE` entries, so the object-ACE and callback-ACE offset branches in the ACE walk are never exercised by any test.

Two non-blocking suggestions:

- Consider `windows.AccessCheck` instead of the hand-rolled ACE walk. It would delete roughly 90 lines of `unsafe` and get ACE ordering, restricted-SID tokens and object ACEs correct for free. The current walk is bounds-checked and fails closed, and the offsets are right, but it treats any applicable deny with an overlapping mask bit as blocking regardless of ACE order, so it will refuse writes Windows itself would permit on non-canonical ACLs. Note the test file already uses `Token.IsMember` while production hand-rolls the equivalent.
- `writer_test.go:29-31` still carries the verbatim `chmod 555 semantics differ on Windows` skip. Taking #298's platform-split path instead of un-skipping is legitimate, but please update the skip message to point at the Windows-native coverage, otherwise the next reader re-files the same issue.
- `TestDenyGroupApplies` is a pure `uint32 → bool` unit test with no Win32 dependency but sits behind `//go:build windows`. Moving `denyGroupApplies` to an untagged file would make it verifiable on every platform, including in this review.

Alan-TheGentleman (changes requested):
Two blocking issues remain:

1. **Do not persist `FILE_DELETE_CHILD` on the parent directory.** `relaxAtomicDirectoryACL` permanently grants the process user authority to delete or rename every child. Because `targetDeniesDelete` runs before temp creation and writing, a target substituted before `os.Rename` can bypass its target-level `DELETE` deny through the parent grant. Grant only `FILE_ADD_FILE` and let the final rename enforce the current target and parent delete permissions.

2. **Preserve NULL DACL semantics.** `descriptor.DACL()` can return nil for a present NULL DACL, which grants full access. The current path treats nil as no deny, calls `ACLFromEntries(..., nil)`, and installs a new user-only ACL. Since a NULL DACL cannot explain the original `CreateTemp` denial, detect nil and return the original error without modifying the DACL.

Native Windows CI must also run green before merge; the current workflow has not executed.

dnlrsls:
Maintainer decision requested before I push the remaining local fixes.

A native Windows control probe confirmed that an external same-volume replacement does not bypass destination directory authorization. `ReplaceFileW` succeeded with a permissive parent, but returned Win32 error 5 (`ERROR_ACCESS_DENIED`) when the destination parent omitted `FILE_ADD_FILE`, even though the existing target and replacement could be opened with the documented replacement rights. Both names, contents, and DACL/control snapshots remained unchanged after the failed call. `MoveFileExW`, `FileRenameInfo`, and hard-link installation cross the same destination namespace boundary.

The current fail-closed production path is therefore the only crash-safe behavior under that ACL. Wiring the temporary parent-DACL relaxation into production could leave `FILE_ADD_FILE` persisted after process termination or power loss. Local native fixes now preserve NULL DACLs, never grant `FILE_DELETE_CHILD`, preserve protected and unprotected target DACL/control state, and distinguish conditional callback ACEs, but the mandatory pre-push review correctly keeps production recovery blocked.

Would you prefer that #298 / this PR be rescoped to native ACL coverage plus an explicit fail-closed limitation, or should this PR close because the approved acceptance requires the write to succeed? I will not push the local fixes until the intended scope is explicit.

Alan-TheGentleman:
Please define the ACL contract before continuing implementation. Document ownership, inherited ACE handling, rollback behavior, and the equivalent non-Windows guarantees so the patch does not accidentally widen access or destroy existing policy. Get explicit maintainer confirmation, then implement that contract against current main with platform-specific failure tests and fresh full CI.

dnlrsls:
Closing this PR without merge after the ACL contract review.

The native Windows probe confirmed that every atomic replacement path still crosses the destination namespace authorization boundary. The current implementation can temporarily grant `FILE_ADD_FILE`, but its `defer`-based restoration cannot survive process termination, fatal crash, or power loss. Production could therefore return an error or terminate while leaving the parent DACL widened.

That means the original #298 acceptance cannot be completed safely in this PR while preserving atomic replacement and crash safety. Production should remain fail-closed under this ACL. The Windows investigation and regression coverage remain useful, but #298 must be corrected or explicitly re-scoped before a future implementation. Any durable ACL recovery mechanism requires a separate approved contract covering ownership, inherited ACEs, protected/unprotected DACLs, rollback, crash recovery, and non-Windows equivalence.

PR #1895, opened 2026-07-27
fix(doctor): stop counting symlinked PATH dirs as duplicate copies

Inglaph:
## Problem

On any distribution with usrmerge (Ubuntu, Fedora, Arch), `/bin` is a symlink to `/usr/bin`. Both are present in `PATH`, so `gentle-ai doctor` reports a single binary as two copies:

```
[!!]  tool:code   code resolved to /usr/bin/code but 2 copies found in PATH: /usr/bin/code, /bin/code
      Remedy: Remove duplicate binaries; keep only one copy of code in PATH
```

The suggested remedy is not actionable: there is no duplicate to remove, and following it would mean deleting a system symlink. The warning also keeps `doctor` reporting `degraded` on an otherwise healthy machine, which trains users to ignore warnings.

On one of our servers this fires three times in a single run (`claude`, `opencode`, `gemini`), all from the same cause.

## Root cause

`doctorToolCopies` deduplicates PATH entries with `filepath.Clean`, which normalizes the string but does not resolve symlinks, so `/bin` and `/usr/bin` are treated as distinct directories.

## Fix

Deduplicate on the resolved location via `filepath.EvalSymlinks`, falling back to the lexical form when the path cannot be resolved (a nonexistent PATH entry still dedups against itself).

Genuinely distinct directories that each hold their own binary are still reported: the check keeps its value, it just stops firing on a layout the user cannot change.

## Tests

Three cases added in `internal/cli/doctor_test.go`:

- symlinked dirs (usrmerge layout) collapse to one copy
- distinct dirs still report two copies
- an unresolvable PATH entry does not drop the copies that were found

Verified that the first test fails without the fix (reports 2 copies) and passes with it. Full `./internal/cli` suite passes locally on go1.25.10, matching the version in `go.mod`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI diagnostics to avoid reporting duplicate tool copies when PATH entries resolve to the same location through symlinks.
  * PATH entries for missing directories are now handled without hiding valid tool copies.
  * Distinct tool locations continue to be reported separately.

* **Tests**
  * Added coverage for symlinked, distinct, and missing PATH directories.

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -278,7 +278,7 @@ func doctorToolCopies(tool string, pathDirs []string) []string {
 	seenDirs := make(map[string]struct{}, len(pathDirs))
 	copies := make([]string, 0, len(pathDirs))
 	for _, dir := range pathDirs {
-		cleanDir := filepath.Clean(dir)
+		cleanDir := doctorCanonicalDir(dir)
 		if _, seen := seenDirs[cleanDir]; seen {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


Inglaph:
Closing this on my side: I keep my changes in my own fork rather than upstreaming them. Leaving it here in case the usrmerge dedup case is useful to anyone. Thanks!

PR #1978, opened 2026-07-29
feat(state): record last_synced_version and last_synced_at after successful sync

### LINKED ISSUE #1273: feat(sync): record last_synced_version in state.json — whether sync ran for the current version is unverifiable
### Pre-flight Checks

### Problem Description
There is no reliable way to answer the question **"did `gentle-ai sync` run after the last binary upgrade?"** — the instrument does not record its own state.

What a user (or an AI agent following orchestrator instructions that say *"confirm sync ran after the upgrade, otherwise stop"*) has available today:

- **`~/.gentle-ai/state.json`** — stores `installed_agents`, `claude_phase_assignments`, `last_update_check`… but **not the version of the last sync**.
- **mtimes of synced assets** (`~/.claude/skills/...`) — a **false-negative trap**. Sync is idempotent: it only rewrites files whose content differs. An indicator that only moves when there is a change cannot distinguish "sync never ran" from "sync ran and there was nothing to do". We hit exactly this: skills with pre-upgrade mtimes led to the wrong conclusion that sync was pending, when it had in fact run twice.
- **`gentle-ai sync --dry-run`** (2.1.4) — prints only step counts (`Prepare steps: 1 / Apply steps: 9`) without listing pending files or stating "0 changes / up to date", so it cannot settle the question either.
- **`~/.gentle-ai/backups/<timestamp>/`** — the only unconditional footprint (snapshot is created before writing), but correlating backup timestamps against the binary's install time is forensics, not a check.

This is the same family as #1261: the tool does not register its own state, so verifying it requires reconstructing evidence from side effects. It is also the verifiability counterpart of #227 (auto-sync after self-update): even if sync becomes automatic, orchestrators still need a one-line way to *prove* configs match the running binary.

### Proposed Solution
1. Write `last_synced_version` (and ideally `last_synced_at`) into `~/.gentle-ai/state.json` at the end of every successful `sync`. The check then becomes one line: compare against `gentle-ai --version`.
2. Make `sync --dry-run` decisive: list the files it would touch, or print an explicit "up to date — 0 changes" when there is nothing to do, instead of only step counts.

### Affected Area
Scripts (setup, installation)

### Alternatives Considered
- Checking mtimes of synced assets — false negatives due to idempotency (described above).
- Correlating `~/.gentle-ai/backups/` snapshot timestamps with the binary install time — works, but it is forensic reconstruction, not a check, and requires knowing that sync always snapshots.
- #227's auto-sync would reduce how often the question matters, but does not make the state observable.

### Additional Context
Observed on gentle-ai 2.1.4, macOS, Homebrew install, Claude Code agent.
Related: #227 (auto-sync after self-update), #1261 (instrument not recording its own state).

ardelperal:
## Linked Issue

Closes #1273

---

## PR Type

- [x] `type:feat` — New feature (non-breaking change that adds functionality)

---

## Summary

Per #1273, the `internal/state` package does not record when `gentle-ai sync` last ran successfully or against which binary version. This PR adds `LastSyncedVersion` and `LastSyncedAt` to `InstallState`, writes them in `sync` after a successful run, and exposes helpers (`LastSyncedVersionOlderThanApp` for the "did sync run after this binary upgrade?" question). The change is scoped to the existing `internal/state` and `internal/cli/sync` modules.

Full feature implementation for #1273. No architecture decisions, no new modules, no new abstractions.

---

## Changes
| File | Change |
|------|--------|
| `internal/state/state.go` | Add `LastSyncedVersion` and `LastSyncedAt` fields to `InstallState`; persist them in the merge path. |
| `internal/state/state_test.go` | Add coverage for the new fields: round-trip, default-empty, merge preserves existing value. |
| `internal/cli/sync.go` | Write the new fields at the end of a successful sync. |
| `internal/cli/sync_test.go` | Test that the fields are written after a successful sync and remain empty when sync fails. |

### DIFF
--- a/internal/cli/sync.go
+++ b/internal/cli/sync.go
@@ -1661,6 +1661,23 @@ func runSyncWithSelection(homeDir string, selection model.Selection, background 
 		return result, persistErr
 	}
 
+	// Record which version of the binary last completed a successful sync.
+	// This allows a future invocation to detect "sync has not run since upgrade".
+	// Re-read state after persistSyncManagedAssetState so we capture the
+	// InstallState it just stamped (InstalledBinaryVersion, ManagedAssetDigest,
+	// etc.) instead of clobbering the on-disk file with the stale snapshot.
+	if persistedStateErr == nil {
+		latest, err := state.Read(homeDir)
+		if err != nil {
+			return result, fmt.Errorf("re-read state for sync version metadata: %w", err)
+		}
+		latest.LastSyncedVersion = AppVersion
+		latest.LastSyncedAt = func() *time.Time { t := time.Now().UTC(); return &t }()
+		if err := state.Write(homeDir, latest); err != nil {
+			return result, fmt.Errorf("persist sync version metadata: %w", err)
+		}
+	}
+
 	return result, nil
 }
 
--- a/internal/cli/sync.go
+++ b/internal/cli/sync.go
@@ -1661,23 +1661,6 @@ func runSyncWithSelection(homeDir string, selection model.Selection, background 
 		return result, persistErr
 	}
 
-	// Record which version of the binary last completed a successful sync.
-	// This allows a future invocation to detect "sync has not run since upgrade".
-	// Re-read state after persistSyncManagedAssetState so we capture the
-	// InstallState it just stamped (InstalledBinaryVersion, ManagedAssetDigest,
-	// etc.) instead of clobbering the on-disk file with the stale snapshot.
-	if persistedStateErr == nil {
-		latest, err := state.Read(homeDir)
-		if err != nil {
-			return result, fmt.Errorf("re-read state for sync version metadata: %w", err)
-		}
-		latest.LastSyncedVersion = AppVersion
-		latest.LastSyncedAt = func() *time.Time { t := time.Now().UTC(); return &t }()
-		if err := state.Write(homeDir, latest); err != nil {
-			return result, fmt.Errorf("persist sync version metadata: %w", err)
-		}
-	}
-
 	return result, nil
 }
 
@@ -1698,6 +1681,19 @@ func persistSyncManagedAssetStateWithBackground(homeDir string, selection model.
 		// the user discovering the skew mid-review at START preflight.
 		if latest.InstalledBinaryVersion != AppVersion {
 			latest.InstalledBinaryVersion = AppVersion
+			shouldWrite = true
+		}
+		// #1978: stamp the version + UTC timestamp of the most recent successful
+		// sync so a future invocation can detect "sync has not run since this
+		// binary upgrade". Always stamped here — runs for fresh-state installs
+		// (where os.IsNotExist was just translated to an empty InstallState)
+		// and zero-agent no-op runs (managed assets unchanged), unlike the
+		// unlocked outside-the-closure write which only ran when persistedState
+		// had already succeeded at function entry.
+		syncNow := time.Now().UTC()
+		if latest.LastSyncedVersion != AppVersion || latest.LastSyncedAt == nil {
+			latest.LastSyncedVersion = AppVersion
+			latest.LastSyncedAt = &syncNow
 			shouldWrite = true
 		}
 		if latest.ManagedAssetDigest != writer {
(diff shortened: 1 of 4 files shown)


ardelperal:
Rebased onto current \main\ (HEAD \d260cdbc\) — 0 commits behind, 1 commit ahead. No conflicts in the 4 touched files (\internal/state/state.go\, \internal/state/state_test.go\, \internal/cli/sync.go\, \internal/cli/sync_test.go\).

Local verification on Windows (Go 1.26.1):

- \go vet ./internal/state/... ./internal/cli/...\ — clean
- \go build ./...\ — clean
- \go test -count=1 ./internal/state/...\ — 1.1s, all PASS (including the 7 new tests)
- \go test -count=1 -run TestRunSyncPreservesCompletePersistedState ./internal/cli/...\ — PASS
- Pre-existing hang in \internal/cli/TestReviewFinalizeArtifactFiles\ (and the other clusters listed in the PR body) confirmed unrelated via a clean \origin/main\ worktree baseline

Ready for re-review.

decode2 (changes requested):
Please move the `last_synced_*` update into the existing locked state read-modify-write and run that completion path for every successful sync, including fresh-state and zero-agent no-op runs.

The current condition skips both cases, while the unlocked full-state rewrite can overwrite concurrent state changes. Please add focused tests for those paths, exact `AppVersion`, UTC timestamp bounds, failure preservation, and concurrent state preservation.

Since the issue's dry-run stale-state projection remains deferred, please also replace `Closes #1273` with a non-closing reference unless this PR completes that contract.

ardelperal:
decode2 review 2026-08-18 addressed.

- **last_synced_* stamping moved into the locked read-modify-write** (persistSyncManagedAssetStateWithBackground). Removed the unlocked full-state rewrite at the end of runSyncWithSelection that could be skipped on fresh-state installs (persistedStateErr == os.IsNotExist) and could overwrite concurrent writes from racing sync invocations. The locked RMW translates os.IsNotExist to an empty InstallState and always stamps, so fresh-state installs AND zero-agent no-op runs both record last_synced_version / last_synced_at.

- **PR body updated**: `Closes #1273` → `Refs #1273` per decode2's note that the dry-run stale-state projection remains deferred (this PR only delivers the success-path stamping; the issue contract is not yet complete).

- **Focused tests added** at the bottom of internal/cli/sync_test.go:
  - TestRunSync_StampsLastSyncedOnFreshState — state.json absent before, stamps written after.
  - TestRunSync_StampsLastSyncedOnZeroAgentSync — `--agents` empty, stamps still applied.
  - TestRunSync_LastSyncedAtIsUTCAndRecent — timestamp in UTC and within ±2s of wall clock.
  - TestRunSync_StampsExactAppVersion — last_synced_version equals AppVersion exactly.
  - TestRunSync_PreservesPriorLastSyncedOnPipelineFailure — a failing sync (forced via malformed prior persona) leaves prior stamps byte-identical.
  - TestRunSync_PreservesStateUnderConcurrentRuns — parallel invocations do not drop persisted fields; the install-state lock serializes the write.

All tests pass.

ardelperal:
Tested on current main @ ec3b1db7aa6d (gentle-ai 2.4.1-0.20260821223000). Background-subagent refactors don't touch state/sync last_synced stamping.

ardelperal:
Closing this PR in favor of a fresh, smaller-scope attempt.

Re-scope rationale (after maintainer-side verification): the `LastSyncedVersion` field was redundant — `internal/cli/sync.go` already stamps `InstalledBinaryVersion = AppVersion` inside the same locked RMW with the same `AppVersion` source (commit #2685), and `doctor.go` already warns on skew. The only field that added real value was the timestamp.

I'll open a new PR from `origin/main` that adds just `LastSyncedAt` to `InstallState`, leaves `LastSyncedVersion` out, and rebases against current main to clear the 218-commit drift. Issue #1273 stays open until that one lands.

Thanks to @decode2 for the original review that pushed the locked-RMW shape — I'll keep that contract in the new PR.

ardelperal:
Closing in favor of a smaller re-scope from origin/main. The version field was redundant with installed_binary_version (main already stamps that in the same RMW, comment #2685); only the timestamp adds value. New PR to follow.

Issue #1983, opened 2026-07-29
fix(windows): 19 tests fail on the newly unblocked Windows lane


Alan-TheGentleman:
## Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

## 📝 Bug Description

Nineteen tests fail on Windows. They are not new regressions — they were invisible.

Until 2026-07-29 the Windows lane never completed a single run: `windows-runtime` declared a 30-minute job timeout around a 35-minute step, so every push to `main` was cancelled at exactly 30 minutes. `v2.2.0` shipped without these tests ever executing. Unblocking the lane and sharding it (so it finishes) made them visible for the first time.

The lane now lives in its own workflow (`.github/workflows/windows-full-suite.yml`) and does **not** gate releases, so this debt is visible without holding publication hostage. That is temporary: **once these are green, fold the lane back into `ci.yml`** so it gates again.

This is a good issue to pick up if you have a Windows machine. Each group below is independent and can be fixed on its own.

## 🔄 Steps to Reproduce

On Windows, with Go 1.25.10+:

```powershell
go test ./internal/cli -count=1 -timeout=30m
go test ./internal/reviewtransaction -count=1 -timeout=30m
```

Or read any run of the [Windows Full Suite](https://github.com/Gentleman-Programming/gentle-ai/actions/workflows/windows-full-suite.yml) workflow.

## ✅ Expected Behavior

The full suite passes on Windows, and the lane returns to `ci.yml` as a required check.

## ❌ Actual Behavior

Nineteen failures. Grouped by probable cause, from a run on `0246f0e7`:

### Lock contention under load (2s budgets)

- `TestPrepareCompactBatchReconciliationCoordinatesMaintenanceThenV2Lock` — `resolve repository Git identity: context deadline exceeded`
- `TestCompactStoreCaptureAdmittedReviewerResultSerializesConcurrentReplayAndConflict` — `authority lock acquisition timed out after 2s`

Windows pays roughly 3x for the process and filesystem work these tests do (measured: identical Organic E2E load is 2m07 on ubuntu against 6m09 on windows-latest). Fixed 2s budgets that are comfortable on Linux are not here. Worth checking whether the budget should be platform-aware or whether the fixture can avoid the contention.

### A filename Windows refuses

[...]

PR #1985, opened 2026-07-29
fix(cli): normalize consent fixture paths on Windows

wunianze666-netizen:
## Summary
- normalize the resolved repository root after JSON escaping before comparing consent contract fixtures
- keep the versioned fixture placeholder stable when Windows expands an 8.3 temporary path

## Why
The consent relay test passes the raw `t.TempDir()` spelling to the CLI, but Git resolves the repository root before it appears in the JSON invocation. On Windows those can differ (`RUNNER~1` versus `runneradmin`), and backslashes are escaped in JSON, so the existing replacement never matched.

## Validation
- `go test ./internal/cli -run TestConsentQuestionMatchesVersionedFixture -count=1 -timeout=10m -v`

Part of #1983.

### DIFF
--- a/internal/cli/review_consent_relay_test.go
+++ b/internal/cli/review_consent_relay_test.go
@@ -424,6 +424,10 @@ func TestConsentQuestionMatchesVersionedFixture(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			reviewModeHome(t)
 			repo := initReviewCLIRepo(t)
+			root, err := resolveReviewMutationRoot(context.Background(), repo)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


wunianze666-netizen:
Closing this draft because current `main` now includes the same resolved-root JSON normalization in `9f0fab9` (`fix(review): restore Windows CI portability`).

I verified the target behavior on a clean Windows worktree at `919ea3d`:

```text
go test ./internal/cli -run '^TestConsentQuestionMatchesVersionedFixture$' -count=1 -timeout=10m -v
PASS (v1 and v2)
```

No additional change is needed from this branch.

PR #2161, opened 2026-08-01
feat(release): publish a signed platform-independent assets archive

### LINKED ISSUE #2156: feat(release): declare the provider surface as a signed release artifact
## Problem

Keeping `gentle-pi` at parity with `gentle-ai` is currently a manual transcription ritual, repeated every release.

The consumer hand-maintains a per-version capability table, ground-truthed by booting each released binary by hand; a hand-written contract-hash map; and the pinned version duplicated across several literals. That last one already failed in production: the installer once reported installing one version while writing a different one to disk, because two hardcoded copies survived a pin bump.

The irony is that the distribution channel already exists and its payload is discarded. `.goreleaser.yaml` ships `docs/review-integration.md` and `contracts/review-integration/v1/{schemas,fixtures}` inside all four platform archives, and the consumer's installer walks the extracted tree and keeps only the executable. Meanwhile `contracts/review-integration/v2/` is never packaged at all, even though the consumer runs the v2 lane exclusively.

`gentle-ai` also already knows how to describe itself: `review capabilities` emits operations, gates, projections, schemas and features. That output is simply never captured into a shipped file.

## Requested outcome

`gentle-ai` declares its surface as a signed release artifact, so consumers verify and consume a declaration instead of transcribing it.

Scope, provider side only:

1. **A versioned, self-describing release-artifact contract** under `contracts/`, versioned independently from `review-integration`. A contract-major change becomes an explicit compatibility event; a consumer meeting an unsupported major must fail closed with an actionable error rather than infer a layout.
2. **A deterministic semantic capability snapshot** that excludes executable, build and host identity so the same bytes apply on every platform. Today `review capabilities` injects the running binary's digest plus Go build and VCS identity, so its live output cannot ship as-is in a common archive.
3. **A signed, platform-independent assets archive** carrying that manifest, its schema, the contract files and the snapshot, inside the existing checksum and minisign envelope so it is covered by the same signature that binds repository and tag.
4. **Post-publish, post-verification downstream notification**, visible and retryable on failure, and structurally unable to invalidate a release that already published and verified.

Unknown mandatory features keep being rejected, matching the advertised `compatibility.unknown_mandatory: "reject"`. Nothing consumer-specific is published from here.

## Why the release shape must be extended carefully

The release plan is enforced in three layers, and one of them is easy to miss: besides the artifact counts and the per-archive platform matrix, `internal/releasepolicy/policy.go` reads the live `.goreleaser.yaml` and `.github/workflows/release.yml` from disk and diffs their YAML AST against embedded Go constants. Any edit to either file needs its constant updated in the same commit or CI fails closed before GoReleaser runs.

That file exists to make the release shape assertable rather than guessable. The amendment must admit exactly one additional archive and add strictly new assertions (no platform axis, no binaries), never relax the existing four-platform matrix, the per-platform-archive binary requirement, or the Windows/Scoop prohibition.

Two related safety notes:

- `brews:` has no `ids:` filter today. That has never mattered with a single archive, but the Homebrew formula is user-facing, so the additional archive needs an explicit archive id and an explicit `ids:` on `brews:`, proven by comparing the formula GoReleaser produces in snapshot mode.
- The self-updater resolves exactly one asset filename by format and reads one manifest line, so it never enumerates archives. An additive assets archive cannot break existing installs.

## Scope of work

Four repository-local units, delivered as a chained set:

- **A1a** — contract namespace, canonical encoder, entry/path/mode rules, tree digest, manifest schema and fixtures, contract doc.
- **A1b** — semantic snapshot projection, generator command, byte-stable golden.
- **A2** — archive assembly, release-policy amendment, verify-assets update, Homebrew safety, docs. Its first task is empirical rather than implementation: inspect the real snapshot artifacts and tarball member paths before committing to an assembly approach.
- **A3** — downstream notification, credential-gated so it ships inert until the receiver exists on the consumer's default branch.

## Out of scope

- Generated client code for new provider operations. The mechanism can detect and report drift; it cannot write a consumer's decoder.
- Anything consumer-specific published from this repository.
- Any relaxation of existing release-shape exactness.
- A prerelease publication path. The release workflow deliberately excludes prerelease tags, and that stays.

Alan-TheGentleman:
Closes #2156

Third slice of the provider chain, stacked on #2159. This is the one that touches the release shape, so it leads with measurement rather than design.

## The empirical gate came first

Both candidate configurations were built with the pinned `goreleaser v2.15.2` (`release --snapshot --clean --skip=sign,publish`) and measured before a single line of policy changed.

**Branch A, a `meta: true` archive** — `dist/artifacts.json` gains a fifth `Archive` with `Extra.ID: assets`, empty `Binaries`, and no goos/goarch/target. Tarball member paths match the staged relative paths exactly. `checksums.txt` gains exactly one line.

**Branch B, `extra_files`** — disqualified, and not on the grounds we expected. Under `--skip=publish` the extra file never appears in `dist/artifacts.json` at all, only in `checksums.txt`. That makes the assets archive **invisible to the pre-publish policy gate**, which is the opposite of what this change is for.

Branch A chosen.

## Four things the measurement caught that no amount of reasoning would have

1. **`before.hooks` run after `--clean` wipes `dist/`**, but before GoReleaser re-creates it. Staging inside `dist/` is destroyed. Staging now happens outside it.
2. **The glob `dir/**/*` silently drops depth-0 files.** As written it would have shipped an archive missing its own manifest. Fixed with `dir/**`.
3. **`tar -C dir -czf .` emits directory-node entries**, which violates the exact-member-path requirement. Fixed with an explicit file list.
4. **The Homebrew formula is not byte-reproducible across runs.** Two consecutive snapshots on an unchanged branch produce different formula hashes, because build timestamps land in tar metadata. Verified independently of the implementing run.

## About that fourth one

The accepted rule for this work was "any delta in the Homebrew formula means stop and fall back". Taken literally that rule is unsatisfiable in any branch, including doing nothing at all.

The rule was applied as **structural** equivalence instead, and that is a deviation worth a reviewer's eye rather than something to accept quietly. What was actually verified: with and without this change the formula is byte-identical once the embedded per-archive sha256 values and the snapshot version string are normalized, and the formula mentions the assets archive **zero times**. `brews[0].ids: [default]` does what it was added to do.

That preserves the intent, which was that `brew upgrade` must not change behavior for real users.

## The policy amendment strengthens, never loosens

`internal/releasepolicy/policy.go` now splits archives by ID. Platform archives keep every assertion they had. The assets archive gets its own equally strict branch: no platform axis, no binaries, and its name must bind the snapshot version.

There is deliberately no generic "N archives allowed" rule. That file exists to make the release shape assertable rather than guessable, so the amendment is exact-plus-one.

The RED tests run in both directions. New behavior must be admitted, and the regression locks must still fail: a missing platform archive, a wrong platform target, a non-`gentle-ai` binary, a platform archive wearing the assets ID, a Windows or Scoop artifact, a live YAML edited without its constant, and a constant edited without its live file.

`policy.go` is compiled in isolation by the release verifier and cannot import other internal packages, so the archive ID is a duplicated literal guarded by a cross-package equality test. That guard was proven non-vacuous by drifting the literal and watching it fail.

Re-running the verifier's own fail-closed test after the copy surfaced real fixture drift in `internal/update`, which is fixed here.

## Rollback

Revert `.goreleaser.yaml`, `.gitignore`, `policy.go` and its two new test files, the two `internal/update` fixtures, `verify-release-assets.sh`, and the doc additions together. That restores the exact four-archive release.


Alan-TheGentleman:
Closing as superseded in practice: gentle-pi shipped 2.2.0 by pinning the gentle-ai v2.4.0 binary directly, so the signed platform-independent assets-archive channel this chain was building never became the consumption path. The feat/pi-parity-assets-artifact base never merged. If we revisit signed asset distribution, it should restart from a fresh design against the current release pipeline.

PR #2162, opened 2026-08-01
feat(release): notify the downstream consumer after publish and verification

### LINKED ISSUE #2156: feat(release): declare the provider surface as a signed release artifact

Alan-TheGentleman:
Closes #2156

Final slice of the provider chain, stacked on #2161.

## Ordering is the whole design

The `notify` job runs `needs: verify`, so it fires only after the release published **and** after remote asset verification passed. It cannot observe or act on a release that has not proven itself.

It also cannot damage one. Notification lives in its own job strictly downstream of `verify`, so a failure is a visible red job and a retryable one, while the published, verified release stays untouched. A downstream notification problem is never a reason to invalidate an upstream publish that already succeeded.

## It ships inert on purpose

GitHub triggers `repository_dispatch` only for a workflow that already exists on the target repository's default branch, and the consumer's receiver does not exist yet. Rather than fire into nothing or fail loudly for a reason nobody can act on, the dispatch step is gated on the presence of the cross-repository credential.

No secret, no dispatch, exit 0, and a printed skip line. When the receiver lands and the token is provisioned, this starts working with no second code change.

The credential-absent path is proven, not asserted: the shell test puts a fake `curl` first on `PATH` that writes a marker and exits 1 if it is ever called, runs the real script with an empty token, then asserts exit 0, the skip line, and that the marker was never created.

## Payload is the tag, and only the tag

The task text originally listed repository, tag, version, commit, archive name and contract major. That predates the maintainer decision and is superseded here, with the supersession recorded in `apply-progress.md` so the divergence reads as a decision rather than a slip.

The reasoning: the consumer must download and verify `checksums.txt`, its minisign signature, and the archive against the authoritative source no matter what the payload says. Anything beyond the tag is either redundant with what it must fetch anyway, or becomes a second source of truth that has to stay in sync with the release. One fewer thing to drift.

## Replay is a separate workflow

`.github/workflows/notify-release.yml` is `workflow_dispatch` only, with a required `tag` input. Keeping it separate means `release.yml` never gains a `workflow_dispatch` trigger, so the publishing workflow does not become manually triggerable, and `policy.go` pins no third file.

Both paths call the same script, so the automatic and manual routes cannot drift apart.

## One regression-lock count changed, and it is not a loosening

`TestReleaseWorkflowUsesFailClosedLeastPrivilegeGates` asserted `persist-credentials: false` appears exactly 3 times. The `notify` job legitimately needs its own checkout, so the workflow now has 4, and the assertion moved to 4.

The assertion still requires **every** checkout to carry it. The test also gained positive assertions for the new job: its presence, its `needs: verify`, the script it calls, and the credential env. Verified independently: the live workflow contains exactly 4 occurrences.

## Rollback

Remove the `notify` job, revert the mirrored block in `expectedReleaseWorkflowYAML`, delete the replay workflow, the script and its test, and revert the additions in the security test. The `preflight`, `release` and `verify` jobs are untouched, as is every file from the earlier slices.


Alan-TheGentleman:
Closing as superseded in practice: gentle-pi shipped 2.2.0 by pinning the gentle-ai v2.4.0 binary directly, so the signed platform-independent assets-archive channel this chain was building never became the consumption path. The feat/pi-parity-assets-artifact base never merged. If we revisit signed asset distribution, it should restart from a fresh design against the current release pipeline.

Issue #1982, opened 2026-07-29
fix(review): a finalize that loses the race is classified as an unknown outcome


Alan-TheGentleman:
## Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

## 📝 Bug Description

A `review finalize` that loses a concurrency race can be classified as `operation_outcome_unknown` with `phase: native_committed` and `mutationOutcome: unknown`, even though it provably published nothing. The caller is told the tool does not know whether it wrote, and a defect report is written into the user's repository.

`TestConcurrentFinalizeElectsExactlyOneWriter` catches it, and it failed **3 of 4 CI runs on 2026-07-29** while passing 8/8 locally. It is not test flakiness — the assertion is correct and the classification it observes is wrong.

## 🔄 Steps to Reproduce

1. Start a review and reach a finalizable state.
2. Issue several concurrent `review finalize` invocations against the same lineage.
3. Inspect the failure envelope of the losers.

Under CI scheduling — a loaded 4-vCPU runner — at least one loser lands on the untyped path. Locally the timing almost always avoids it, which is why `-count=8` passes on a workstation.

## ✅ Expected Behavior

Every loser of the race carries a typed, not-started, retry-safe classification. The test encodes the three legitimate outcomes:

- `authority_lock_contention` — refused by the lock while the winner held it
- `authority_lock_timeout` — waited and gave up
- `authority_revision_conflict` — won the lock, lost the compare-and-set

All three mean the same thing to the caller: nothing was published, reissue the request.

## ❌ Actual Behavior

```
concurrent FINALIZE loser 2 = cli.ReviewIntegrationFailure{
  Phase:           "native_committed",
  Code:            "operation_outcome_unknown",
  Message:         "The negotiated review operation failed without authoritative mutation evidence.",
  MutationOutcome: "unknown",
  RetrySafe:       false,
}
(error: ... A defect report was saved at .git/gentle-ai/defect-reports/operation-outcome-unknown-<hash>.md)
```

Observed on loser 4 in one run and loser 2 in another — which loser is hit is scheduling.

## Where it comes from

`internal/cli/review_operation_contract.go:605-630` types `CompactRevisionConflictError` into `authority_revision_conflict`, and its own comment states the boundary deliberately:

[...]

PR #2183, opened 2026-08-01
fix(review): keep the native reason on an unclassified post-progress failure

### LINKED ISSUE #2181: fix(review): concurrent finalize loser still reports operation_outcome_unknown with an empty cause
### Pre-flight Checklist

### 📝 Bug Description

A `review finalize` that loses a concurrency race can still be classified as `operation_outcome_unknown` with `phase: native_committed` and an **empty** `cause`, so the caller is told the tool cannot say whether it wrote, with nothing to act on.

This is the same symptom as #1982, which was closed as completed on 2026-07-29. It reproduces on current `main` (`bd32fe58`) with no local changes, so that fix appears to have left one untyped path open. Filing separately rather than relying on the closed issue, since a closed issue cannot carry `status:approved` for a linked PR; #1982 has the full prior discussion.

The remaining path is a `switch` with no `default`, described under Actual Behavior.

### 🔄 Steps to Reproduce

1. Check out `main` at `bd32fe58` in a clean worktree, with no local modifications.
2. Run the focused concurrency probe:
   ```
   go test ./internal/cli -run TestConcurrent -count=25
   ```
3. Observe `--- FAIL: TestConcurrentFinalizeElectsExactlyOneWriter`, roughly three times per 25 iterations.

The probe finishes in seconds. A full `go test ./internal/cli` run also reproduces it but costs about 130 seconds per sample and fails less predictably, which is why the focused probe is the better reproduction.

Measured on the same machine: 3 failures per 25 iterations on pristine `bd32fe58`, and 3 failures per 25 on a branch carrying unrelated local commits. The identical rate is what establishes it as pre-existing rather than caused by downstream work.

### ✅ Expected Behavior

Every loser of the race carries a typed classification, and any envelope that reports an unknown outcome still carries the native reason in `cause`, so the caller can act or report it.

`TestConcurrentFinalizeElectsExactlyOneWriter` encodes the three legitimate outcomes: `authority_lock_contention`, `authority_lock_timeout`, and `authority_revision_conflict`.

### ❌ Actual Behavior

The loser envelope is:

```text
Code:            "operation_outcome_unknown"
Phase:           "native_committed"
MutationOutcome: "unknown"
RetrySafe:       false
Replayability:   "status_required"
Cause:           ""
RequestDigest:   ""
```

Root cause, in `internal/cli/review_operation_contract.go`, the `reviewFacadeOperationProgressError` branch (lines 511-537 on `bd32fe58`):

The branch sets `Phase = "native_committed"`, `MutationOutcome = Unknown`, `AuthorityApplicability = "current_target"`, `RetrySafe = false`, `Replayability = StatusRequired` and `LineageID`, then enters a `switch` with exactly four cases — `git_command_timeout`, `git_command_failed` twice, and `operation_timeout`. There is no `default`.

When the wrapped error is a progress error but none of those four, no case matches. `Code` therefore keeps the struct-literal default from line 346 (`operation_outcome_unknown`), and the branch returns at line 537, which is before line 871 — the only place that assigns `failure.Cause = reviewIntegrationFailureCause(runErr)`.

That accounts for every field above, including both empty ones: the early return skips the assignment that would have preserved the native reason. The comment on the general default at lines 866-870 states that `cause` exists to carry the real native reason "instead of discarding it", which is precisely what this branch does.

A defect report is also written into the user's repository each time, so the user is asked to file a report for a state the tool could have described.

### Suggested fix

Add a `default:` to that switch. At minimum it should set `failure.Cause = reviewIntegrationFailureCause(runErr)`, so the caller is not handed "failed without authoritative mutation evidence" with no diagnostic. A dedicated code and message for "progressed past commit, stopped for an unclassified reason" would be better still, because the envelope currently claims the outcome is unknown while the branch already knows a native transition committed.

I have **not** verified that every loser reaching this branch provably published nothing, so I am deliberately not proposing that it be reclassified as not-started and retry-safe. That call needs someone who knows the lock's guarantees.

### Gentle AI Version

```
bd32fe58 (main, 2026-08-01)
```

### Operating System

`Linux (Arch/Manjaro)`

### AI Agent / Client

`OpenCode`

### 📋 Affected Area

`CLI (commands, flags)`

### 💡 Logs / Error Output

```text
--- FAIL: TestConcurrentFinalizeElectsExactlyOneWriter
    review_lock_contention_test.go:269: concurrent FINALIZE loser 3 =
    ReviewIntegrationFailure{
      Schema:"gentle-ai.review-integration.failure/v1",
      Contract:"gentle-ai.review-integration/v1",
      Operation:"review.finalize",
      Phase:"native_committed",
      Code:"operation_outcome_unknown",
      Message:"The negotiated review operation failed without authoritative mutation evidence.",
      MutationOutcome:"unknown",
      AuthorityApplicability:"current_target",
      RetrySafe:false,
      Replayability:"status_required",
      RequestDigest:"",
      CauseCategory:"",
      Cause:"",
    }
```

### Additional Context

Ruled out while narrowing this down, recorded so nobody repeats it:

- Not load-related. It reproduces on an idle machine.
- Not deterministic. The same commit passes 8 of 10 full-suite runs, which is why bisecting against it yields an arbitrary culprit.
- Not checkout-versus-worktree. An apparent correlation across nine samples disappeared as the sample size grew.
- Not non-deterministic receipt derivation. `CompactReceipt` carries no timestamp or other unstable field, and `store.go:1022` is already idempotent for byte-identical payloads.

Separately, the same worktree accumulated 30 `receipt-publication-conflict` defect reports under the repository's Git directory, whose own text reads "a tool-internal fault state that should never happen" and "there is no command to reconcile it". I did not chase that one, so it may or may not share this root cause.

Happy to open a focused PR for the missing `default` once this has `status:approved`.

aparragithub:
## 🔗 Linked Issue

Closes #2181

> **Maintainer note, stated up front:** #2181 has no `status:approved` label yet, so `Check Issue Has status:approved` and `Check PR Has type:* Label` will both be red. A contributor account cannot apply repository labels (HTTP 403, admin-only), so neither is fixable from this side and this PR discloses it rather than looking merge-ready. Related history is in #1982, which described the same symptom and was closed as completed; it is closed and unlabelled, so it cannot carry approval for a linked PR, which is why #2181 exists.

---

## 🏷️ PR Type

---

## 📝 Summary

- Adds the missing `default:` to the post-progress classification switch so an unclassified cause is no longer discarded.
- Deliberately does **not** change the classification itself: the 1861 unknown-outcome rule is preserved exactly.
- Scope is one `default:` arm plus one regression test.

---

## 📂 Changes

| `internal/cli/review_operation_contract.go` | `default:` arm on the `reviewFacadeOperationProgressError` switch, assigning `failure.Cause` from the native error. |
| `internal/cli/review_progress_cause_test.go` | New regression test: the 1861 classification is asserted unchanged, and the cause must survive. |

---

## 🧠 What was actually wrong

The `reviewFacadeOperationProgressError` branch sets `Phase = "native_committed"` and the rest of the post-progress classification, then enters a `switch` with four cases: `git_command_timeout`, `git_command_failed` twice, and `operation_timeout`. There was no `default`.

When the wrapped cause is none of those four, no case matches, so `Code` keeps its struct-literal default (`operation_outcome_unknown`) and the branch returns early — before the line that assigns `failure.Cause = reviewIntegrationFailureCause(runErr)`. The caller receives:

```text
Code:            "operation_outcome_unknown"
Phase:           "native_committed"
MutationOutcome: "unknown"
Cause:           ""
```

"The negotiated review operation failed without authoritative mutation evidence", with no reason attached, for a state whose native reason the tool was already holding. A defect report is written into the user's repository as well, so the user is asked to report a state the tool could have described.

The comment on the general default in the same function says `cause` exists to carry the real native reason "instead of discarding it". This branch discarded it.

---

## 🚧 What this deliberately does not change

The unknown-outcome classification is correct and stays. `TestLockContentionAfterACommittedTransitionStillReportsUnknown` carries an explicit instruction not to relax it: a failure arriving after a committed native transition must keep reporting an unknown outcome, because a caller that retries a maybe-committed mutation can double-apply it. That test passes unchanged here, and the new test asserts the same classification fields to keep it that way.

Consequently **this does not make `TestConcurrentFinalizeElectsExactlyOneWriter` deterministic.** That test requires every loser to be typed not-started and retry-safe, which is a different question: whether a loser can legitimately reach the post-progress branch at all. I have not verified that every loser reaching it provably published nothing, so I have not touched it. #2181 records that open question.

---

## 🧪 Test Plan

- [x] `go build ./...`
- [x] `go vet ./internal/cli`
- [x] `go test ./internal/cli` — no new failures
- [x] Failing-test-first: the new test fails on `main` with `unclassified post-progress failure carries no cause`, and its printed envelope is the same shape reported in #2181 including the empty `cause`. The classification assertions pass before that point, which is what localises the defect to the discarded reason rather than the classification.
- [x] Both 1861 guard tests pass unchanged: `TestLockContentionAfterACommittedTransitionStillReportsUnknown` and `TestFinalizeRevisionConflictIsRetryableNotUnknown`.
- [ ] Full Docker E2E suite (not applicable to this classification-only change)

---

## 🤖 Automated Checks

| Check PR Cognitive Load | ⏳ | Well inside the 400-line budget |
| Check Issue Reference | ⏳ | Body contains `Closes #2181` |
| Check Issue Has `status:approved` | ❌ | Blocked on maintainer; #2181 is unlabelled |
| Check PR Has `type:*` Label | ❌ | Blocked on maintainer; contributor cannot apply labels |
| Unit Tests | ⏳ | `go test ./...` |
| Go Format | ⏳ | `go run ./internal/gofmtcheck` |
| E2E Tests | ⏳ | `cd e2e && ./docker-test.sh` |

---

## ✅ Contributor Checklist

- [x] PR is linked to existing issue #2181.
- [ ] Issue has `status:approved` — blocked on maintainer.
- [ ] Exactly one `type:*` label is applied — blocked on maintainer.
- [x] Exactly one PR type is selected in the body.
- [x] Relevant unit tests, vet, and format checks pass.
- [x] Commit follows Conventional Commits.
- [x] Commit contains no `Co-Authored-By` trailer.

---

## 💬 Notes for Reviewers

One judgement call worth confirming: the `default:` keeps `Code` as `operation_outcome_unknown` rather than minting a new code. A dedicated code such as `progress_outcome_unclassified` would be more precise, but it is an additive contract value and I would rather you choose whether that belongs in the published enum than add one unprompted. Happy to switch if you prefer it.

Note also that `TestConcurrentFinalizeElectsExactlyOneWriter` is flaky on `main` independently of this change: roughly three failures per 25 iterations of `go test ./internal/cli -run TestConcurrent -count=25`, measured at the same rate with and without unrelated local commits. This PR does not change that rate; #2181 has the measurements.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting for certain operation failures by preserving the original error details in failure messages.
  * Retained existing failure classification, retry behavior, required review actions, and operation history.
  * Clarified diagnostics for failures occurring after an operation has been committed, helping prevent unsafe retries.

* **Tests**
  * Added coverage for post-commit failures, lock contention, error preservation, and existing retry and status behavior.


Alan-TheGentleman:
I am taking over the bounded deterministic test correction now while preserving the contributor's diagnostic fix.

Alan-TheGentleman (approved):
I approve the contributor's diagnostic fix. I verified the deterministic concurrent FINALIZE correction fails closed, accepting only a diagnosed non-retryable post-transition outcome.

PR #2198, opened 2026-08-01
fix(engram): sync downloaded file and propagate close/sync errors

### LINKED ISSUE #1998: Engram download discards the file Close error and never syncs, so a failed flush can yield success with a verifying digest
### Pre-flight Checklist

### 📝 Bug Description

While auditing the download paths (static analysis plus code reading; no runtime incident; verified against v2.2.2), I found an error-handling gap in the engram component download that can let a corrupted/incomplete archive pass checksum verification.

`engramDownloadToFile` (`internal/components/engram/download.go:499-529`):

- opens the target with `os.Create(outPath)` and closes it via a **deferred `Close` whose error is discarded** (`download.go:517-521`); there is no `Sync`;
- copies with `io.Copy(io.MultiWriter(f, h), resp.Body)` and returns `hex.EncodeToString(h.Sum(nil))` (`download.go:523-528`) — i.e. the digest is computed from the **in-memory stream**, not from what actually reached disk.

To be precise about the failure window: a short `Write` during the copy is NOT the problem — `os.File.Write` and `io.MultiWriter` both turn a short count into an error, so the copy would fail loudly. The unguarded window is **write-back failure surfacing at `Close`** (or after it, since nothing ever `Sync`s): the kernel accepts the bytes into the page cache, the deferred `Close` reports the deferred I/O error, and that error is discarded. In that case the function returns a nil error **and** a digest that matches the release manifest — the caller `DownloadLatestBinary` (`download.go:189-207`) compares the streaming digest against the expected checksum, which passes because both were computed from the same stream, and never re-reads the persisted file. The corruption is certified rather than caught, and only surfaces later (if at all) when extraction opens the truncated archive.

A close-error guard already exists in this codebase for the same job: `internal/update/upgrade/download.go:210-259` folds the deferred `Close` error into the named return error and removes the partial output on failure (`download.go:243-245`). (That sibling also skips `Sync`, so it guards the close-error half only — but the engram path guards neither.)

### 🔄 Steps to Reproduce

These are fault-window defects, so reproduction is via fault injection rather than a user recipe:

1. Make the file's `Close` report an error (deferred write-back failure — e.g. an injected `Close` error in a test seam; in the wild: NFS/quota or any delayed-write error surfacing at close time).
2. `engramDownloadToFile` returns nil error and a digest computed from the in-memory stream.
3. Checksum verification in `DownloadLatestBinary` passes, while the on-disk archive is incomplete — the failure only surfaces later (if at all) when extraction opens the truncated file.

### ✅ Expected Behavior

A download that reports success (and a verifying digest) has fully durable, complete bytes on disk. A close/flush failure surfaces as an error and the partial file is removed — the close-error half of this is exactly what `internal/update/upgrade/download.go:234-246` already does for binary downloads; adding `Sync` before the checked `Close` would cover the durability half.

### ❌ Actual Behavior

When write-back fails at close time, the download returns success plus a digest that verifies against the release manifest over a file that is incomplete on disk (silent corruption with a passing integrity check).

### Gentle AI Version

v2.2.2

### Operating System

Linux (Ubuntu/Debian)

### AI Agent / Client

Claude Code

### 📋 Affected Area

Installation Pipeline

### 💡 Logs / Error Output

```shell
Output of a standalone reproduction (not a runtime incident): a verbatim
mirror of engramDownloadToFile where the file handle simulates kernel
write-back failing at Close, plus a control using the guarded pattern from
internal/update/upgrade/download.go:234-246 against the same failure.

release content: 1048576 bytes, checksums.txt digest 4ba90e1ea8710178...

[replica of engramDownloadToFile - download.go:499-529]
  returned error         : <nil>
  returned digest        : 4ba90e1ea8710178...
  checksum verification  : match=true  (DownloadLatestBinary would ACCEPT)
  on-disk file           : 983040 of 1048576 bytes (65536 lost at close)
  on-disk digest         : 89e81084e130b794...  match=false

[control: guarded-close pattern from upgrade/download.go:234-246]
  returned error         : close guarded.tar.gz: input/output error (write-back failed, 65536 bytes lost)
  partial file removed   : true

REPRODUCED (silent success + verifying digest + short file): true
CONTROL (guarded pattern surfaces the same failure)        : true
```

### Additional Context

If approved, I'd like to fix this in a single small PR (`fix/engram-download-close-sync`, well under the 400-line budget): port the guarded-close pattern from `internal/update/upgrade/download.go:234-246` to `engramDownloadToFile`, add `Sync` before the checked `Close`, plus a regression test that injects a close failure and asserts the function returns an error and removes the partial file. If you'd like, the same PR (or a follow-up) can add the missing `Sync` to the upgrade downloader too — your call on scope.

About the log above: it comes from a standalone reproduction program that mirrors `engramDownloadToFile` line by line (same `io.Copy(io.MultiWriter(f, h), resp.Body)`, same discarded deferred `Close`, same stream-side digest return); the only substitution is the file handle, which models kernel write-back failing at `Close` — the injection is external because `os.Create` is hardcoded in the function, which is also why the fix includes a test seam. Happy to share the full program as a gist if useful.

**Acceptance criteria** (updated per review feedback):

- An injected `Sync` failure returns an identifiable error and the partial file is removed.
- An injected `Close` failure returns an identifiable error and the partial file is removed.
- Both errors are propagated to the caller rather than discarded.
- The success path proves that a successful `Sync` + `Close` preserve the expected bytes on disk and the digest matches the persisted content.

The regression tests will cover all four via an injectable file seam (needed anyway, since `os.Create` is hardcoded in `engramDownloadToFile`).

ardelperal:
## 🔗 Linked Issue
Closes #1998

## 🏷️ PR Type

## 📝 Summary
Fixes a silent-corruption bug in `engramDownloadToFile` (`internal/components/engram/download.go`) where a write-back failure at file `Close` time is discarded, allowing an incomplete on-disk archive to pass checksum verification because the digest was computed from the in-memory stream rather than from the persisted bytes.

The change:

- Adds `f.Sync()` after the `io.Copy` completes, surfacing kernel write-back failures as a typed error (`sync %s: %w`).
- Replaces the discarded `defer f.Close()` with a defer-func that folds the close error into the named return when no prior error exists, and removes the partial output when the run did not reach the successful `completed = true` marker.
- Introduces an unexported `writeCloseSyncer` interface and a package-level `engramCreateFile` seam so tests can inject controllable files.
- Adds four regression tests at `internal/components/engram/download_test.go` exercising the seam for Sync failure, Close failure, both-error ordering, and the success-path byte/digest invariant.

The function signature and the digest computation are unchanged. No new exported symbols.

## 📂 Changes
| File | Change |
|------|--------|
| `internal/components/engram/download.go` | Added `writeCloseSyncer` interface, `engramCreateFile` seam, `f.Sync()` call, and the close-fold + partial-remove defer pattern in `engramDownloadToFile`. Function signature unchanged. |
| `internal/components/engram/download_test.go` | Added `fakeFile` test fixture and four regression tests (`TestEngramDownloadToFile_SyncFailure…`, `…CloseFailure…`, `…BothErrors…`, `…SuccessPath…`). |

Two work-unit commits:

1. `e262fcc2` — `fix(engram): sync downloaded file and propagate close/sync errors`
2. `cf6fe40e` — `test(engram): add regression tests for download close/sync error handling`

Total: **204 insertions, 3 deletions** across 2 files (well under the 400-line cognitive-load budget).

## 🧪 Test Plan
- [x] Unit tests pass (`go test ./internal/components/engram/... -count=1`) — 9.069s, all 4 new tests green
- [x] Go format passes (`go run ./internal/gofmtcheck`) — clean
- [ ] E2E tests pass (`cd e2e && ./docker-test.sh`) — N/A for this fix (no Docker-touching changes); deferred to CI on `pr-check.yml`
- [x] Manually tested locally — covered by the new unit tests; no UI surface to exercise manually

```bash
$ go test ./internal/components/engram/... -run "TestEngramDownloadToFile" -v
=== RUN   TestEngramDownloadToFile_SyncFailureReturnsErrorAndRemovesPartial
--- PASS: ... (0.01s)
=== RUN   TestEngramDownloadToFile_CloseFailureReturnsErrorAndRemovesPartial
--- PASS: ... (0.01s)
=== RUN   TestEngramDownloadToFile_BothErrorsPropagatedNotDiscarded
=== RUN   TestEngramDownloadToFile_BothErrorsPropagatedNotDiscarded/Sync_error_shadows_Close_error
--- PASS: ... (0.00s)
=== RUN   TestEngramDownloadToFile_BothErrorsPropagatedNotDiscarded/Close_error_after_successful_Sync
--- PASS: ... (0.00s)
=== RUN   TestEngramDownloadToFile_SuccessPathPersistsBytesAndDigest
--- PASS: ... (0.03s)
PASS
ok  github.com/gentleman-programming/gentle-ai/v2/internal/components/engram  3.241s
```

### Pre-existing failures (not introduced by this PR, verified with `git stash` baseline on clean `main` @ `b6ecb3ef`)

The `gentle-ai-collab-perfect` skill names `internal/components/communitytool/pi_codegraph`, `internal/tui/sync`, and `internal/tui/sddmode` as pre-existing failures. Verified on current `main`:

- `pi_codegraph` lives **inside** the `communitytool` package (not a separate directory); the package passes in 51s on clean main. **Skill claim is stale.**
- `internal/tui/sync` and `internal/tui/sddmode` packages **do not exist** on main. **Skill claim is stale.**

The actual pre-existing failures on main (verified with `git stash push -- <files> && go test ./... && git stash pop`):

- `internal/update` — 9 failing tests (release/security/CI coverage, not related to this fix).
- `internal/update/upgrade` — `TestRunStrategyUsesCaskOwnershipAndMigrationGuidance`.
- `internal/reviewtransaction` — package-level timeout.
- `internal/sddstatus` — package-level timeout.

Additionally, `internal/cli.TestReviewCaptureResultConcurrentSelectedLenses` flakes under concurrent full-suite load but passes in isolation (12.3s). Verified identical flake on clean main.

## 🤖 Automated Checks
| Check | Status | Notes |
|-------|--------|-------|
| `Check PR Cognitive Load` | ⏳ | 204 +/− 3 = 207 lines, well under 400 |
| `Check Issue Reference` | ⏳ | `Closes #1998` |
| `Check Issue Has status:approved` | ⏳ | Issue has `status:approved` + `bug` labels |
| `Check PR Has type:* Label` | ⏳ | Awaiting maintainer action (see below) |

## ✅ Contributor Checklist
- [ ] I have added the appropriate `type:*` label to this PR — see Pending maintainer actions
- [x] Unit tests pass (`go test ./internal/components/engram/...`)
- [ ] E2E tests pass (`cd e2e && ./docker-test.sh`) — N/A for this fix; deferred to CI
- [ ] Benchmark validation completed, or this change is not applicable to the benchmark — N/A (no benchmark impact)
- [x] I have updated documentation if necessary — N/A (no behavior change requiring docs)
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format

## Pending maintainer actions

The following are **maintainer-applied per `pr-check.yml` and `CONTRIBUTING.md`** in this repo — not within contributor scope (verified empirically `gh pr edit --add-label` returns 403 `AddLabelsToLabelable` for `ardelperal`):

- [ ] `type:bug` label applied to this PR — pending Alan
- [ ] Fork workflow approval — first-push run(s) on `action_required` pending Alan

## 💬 Notes for Reviewers
- Two commits, work-unit split: fix first (`e262fcc2`), tests second (`cf6fe40e`). Each commit is independently reviewable; the fix commit alone keeps the package green.
- Reference pattern lifted from `internal/update/upgrade/download.go` lines 224-232 (the `completed` + defer-func + close-fold + partial-remove pattern), with one enhancement vs. the reference: `f.Sync()` is called explicitly before `Close` so kernel write-back failures surface as a typed error. The reference does not call `Sync`; the issue's "Expected Behavior" section calls for it.
- New test seam `engramCreateFile` and the `writeCloseSyncer` interface are unexported. No public API change.
- The four regression tests use `httptest.NewServer`, `t.TempDir()`, and `t.Cleanup`-restored package-level var swaps. Cross-platform (no symlinks, no OS-specific paths).
- CodeGraph flagged `engramDownloadToFile` as having no direct unit tests before this change (all coverage went through `DownloadLatestBinary` at the seam level). The new tests are the first direct unit coverage of the function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download reliability by synchronizing files before completion.
  * Download failures now correctly propagate errors and remove incomplete output files.
  * File closing errors are surfaced to provide clearer failure reporting.
  * Successful downloads continue to persist complete data with the expected checksum.

### DIFF
--- a/internal/components/engram/download.go
+++ b/internal/components/engram/download.go
@@ -75,6 +75,17 @@ var (
 	}
 )
 
+// writeCloseSyncer abstracts *os.File for test seam injection.
+type writeCloseSyncer interface {
+	io.Writer
+	Sync() error
+	Close() error
+}
+
+var engramCreateFile = func(path string) (writeCloseSyncer, error) {
+	return os.Create(path)
+}
+
 func goPrivateModuleEnv(base []string, modulePath string) []string {
 	values := map[string]string{
 		"GONOSUMDB": modulePath,
@@ -514,17 +525,28 @@ func engramDownloadToFile(ctx context.Context, url string, outPath string) (hexD
 	if err := os.MkdirAll(filepath.Dir(outPath), 0o755); err != nil {
 		return "", fmt.Errorf("create dir: %w", err)
 	}
-	f, err := os.Create(outPath)
+	f, err := engramCreateFile(outPath)
 	if err != nil {
 		return "", fmt.Errorf("create %s: %w", outPath, err)
 	}
-	defer f.Close()
+	completed := false
+	defer func() {
+		if closeErr := f.Close(); err == nil && closeErr != nil {
+			err = fmt.Errorf("close %s: %w", outPath, closeErr)
+		}
+		if !completed || err != nil {
+			_ = os.Remove(outPath)
+		}
+	}()
 
 	h := sha256.New()
 	if _, err := io.Copy(io.MultiWriter(f, h), resp.Body); err != nil {
 		return "", fmt.Errorf("write %s: %w", outPath, err)
 	}
-
+	if err := f.Sync(); err != nil {
+		return "", fmt.Errorf("sync %s: %w", outPath, err)
+	}
+	completed = true
 	return hex.EncodeToString(h.Sum(nil)), nil
 }
 
--- a/internal/components/engram/download_test.go
+++ b/internal/components/engram/download_test.go
@@ -1455,3 +1455,182 @@ func TestSHA256ChecksumContract(t *testing.T) {
 	t.Logf("SHA256 contract: Go produces %q format (64 lowercase hex chars)", goDigest)
 	t.Logf("PowerShell fallback must produce identical format using .NET SHA256")
 }
+
+// fakeFile satisfies writeCloseSyncer for regression testing of engramDownloadToFile.
+type fakeFile struct {
+	writeErr error // returned by Write
+	syncErr  error // returned by Sync
+	closeErr error // returned by Close
+	written  []byte
+}
+
+func (f *fakeFile) Write(p []byte) (int, error) {
+	if f.writeErr != nil {
+		return 0, f.writeErr
+	}
+	f.written = append(f.written, p...)
+	return len(p), nil
+}
+func (f *fakeFile) Sync() error  { return f.syncErr }
+func (f *fakeFile) Close() error { return f.closeErr }
+
+// --- TestEngramDownloadToFile_SyncFailureReturnsErrorAndRemovesPartial ---
+
+func TestEngramDownloadToFile_SyncFailureReturnsErrorAndRemovesPartial(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("test content"))
+	}))
+	defer server.Close()
+
+	origCreate := engramCreateFile
+	t.Cleanup(func() { engramCreateFile = origCreate })
+
+	tmpDir := t.TempDir()
+	outPath := filepath.Join(tmpDir, "out")
+
+	engramCreateFile = func(path string) (writeCloseSyncer, error) {
+		return &fakeFile{syncErr: errors.New("sync failed")}, nil
+	}
+
+	_, err := engramDownloadToFile(context.Background(), server.URL, outPath)
+	if err == nil {
+		t.Fatal("expected non-nil error, got nil")
+	}
+	if !strings.Contains(err.Error(), "sync failed") {
+		t.Errorf("error = %v, want to contain 'sync failed'", err)
+	}
+	if _, statErr := os.Stat(outPath); !os.IsNotExist(statErr) {
+		t.Errorf("outPath stat = %v, want os.IsNotExist", statErr)
+	}
+}
+
+// --- TestEngramDownloadToFile_CloseFailureReturnsErrorAndRemovesPartial ---
+
+func TestEngramDownloadToFile_CloseFailureReturnsErrorAndRemovesPartial(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("test content"))
+	}))
+	defer server.Close()
+
+	origCreate := engramCreateFile
+	t.Cleanup(func() { engramCreateFile = origCreate })
+
+	tmpDir := t.TempDir()
+	outPath := filepath.Join(tmpDir, "out")
+
+	engramCreateFile = func(path string) (writeCloseSyncer, error) {
+		return &fakeFile{closeErr: errors.New("close failed")}, nil
+	}
+
+	_, err := engramDownloadToFile(context.Background(), server.URL, outPath)
+	if err == nil {
+		t.Fatal("expected non-nil error, got nil")
+	}
+	if !strings.Contains(err.Error(), "close failed") {
+		t.Errorf("error = %v, want to contain 'close failed'", err)
+	}
+	if _, statErr := os.Stat(outPath); !os.IsNotExist(statErr) {
+		t.Errorf("outPath stat = %v, want os.IsNotExist", statErr)
+	}
+}
+
+// --- TestEngramDownloadToFile_BothErrorsPropagatedNotDiscarded ---
+
+func TestEngramDownloadToFile_BothErrorsPropagatedNotDiscarded(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("test content"))
+	}))
+	defer server.Close()
+
+	origCreate := engramCreateFile
+	t.Cleanup(func() { engramCreateFile = origCreate })
+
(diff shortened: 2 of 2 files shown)


Alan-TheGentleman:
Thanks for this, ardelperal. The diagnosis is exact and the bug is real. On current main `engramDownloadToFile` (`internal/components/engram/download.go:517-526`) still does `defer f.Close()` with the error thrown away, never calls `Sync`, and leaves the partial archive in place, so the checksum is verified against bytes that were never proven to reach disk.

#2184 closes the same issue (#1998) with the same fix and opened about 80 minutes earlier the same day: same Sync-before-digest ordering, same interface seam over `os.Create`, propagated close error, partial-file cleanup. On duplicates the earlier PR has priority, so #2184 is the one to carry forward.

Two things from yours worth checking against it before it merges: your both-fail error-ordering assertion, and whether its fault injection covers the same four cases yours does.

ardelperal:
Closing as duplicate of #2184 per Alan's comment 2026-08-04T18:30Z — same #1998 fix (Sync-before-digest ordering, interface seam over os.Create, propagated close error, partial-file cleanup), opened ~80 minutes earlier the same day. On duplicates the earlier PR has priority; #2184 is the canonical carrier. Thanks for the diff-pairing review — your both-fail error-ordering assertion and four-case fault injection are good test patterns worth porting to #2184 if it doesn't already cover them.

PR #1105, opened 2026-07-10
fix(sdd-apply): move Executor Override before Orchestrator Gate

chrisssp:
## 🔗 Linked Issue

Closes #721

---

## 🏷️ PR Type

What kind of change does this PR introduce?

- [x] `type:bug` — Bug fix (non-breaking change that fixes an issue)
- [ ] `type:feature` — New feature (non-breaking change that adds functionality)
- [ ] `type:docs` — Documentation only
- [ ] `type:refactor` — Code refactoring (no functional changes)
- [ ] `type:chore` — Build, CI, or tooling changes
- [ ] `type:breaking-change` — Breaking change (fix or feature that changes existing behavior)

---

## 📝 Summary

The `Executor Override` block in `skills/sdd-apply/SKILL.md` was placed **after** the `ORCHESTRATOR GATE`. LLMs are autoregressive — they process instructions sequentially. When the `sdd-apply` sub-agent reads the skill, the first imperative instruction it encounters is "STOP. Delegate to the dedicated sdd-apply sub-agent". By the time it reaches the Executor Override (which says the gate does NOT apply), it has already locked onto the delegate action.

This causes the sub-agent to attempt self-delegation, resulting in loops or no action taken.

The fix moves the `Executor Override` **above** the `ORCHESTRATOR GATE` so the sub-agent reads "you are the executor — execute" before encountering the gate instruction. The internal reference is also updated ("the gate above" → "the gate below").

---

## 📂 Changes

| File / Area | What Changed |
|-------------|-------------|
| `internal/assets/skills/sdd-apply/SKILL.md` | Moved `## Executor Override` block above `ORCHESTRATOR GATE`; updated "the gate above" → "the gate below" |

---

## 🧪 Test Plan

**Unit Tests**
```bash
go test ./...
```

**E2E Tests** (Docker required)
```bash
cd e2e && ./docker-test.sh
```

- [x] Unit tests pass (`go test ./...`)
- [x] E2E tests pass (`cd e2e && ./docker-test.sh`) — Ubuntu: 79 passed, 0 failed, 2 skipped
- [ ] Manually tested locally

---

## 🤖 Automated Checks

The following checks run automatically on this PR:

| Check | Status | Description |
|-------|--------|-------------|
| Check PR Cognitive Load | ⏳ | PR should stay within 400 changed lines (`additions + deletions`) or use `size:exception` |
| Check Issue Reference | ⏳ | PR body must contain `Closes/Fixes/Resolves #N` |
[...]

PR #2210, opened 2026-08-01
fix(sdd-apply): prevent executor self-delegation

### LINKED ISSUE #721: fix(sdd-apply): Executor Override is ignored because ORCHESTRATOR GATE precedes it — sub-agent still loops delegating to itself
### Pre-flight Checklist

### 📝 Bug Description

The fix from PR #665 (issue #636) added the `## Executor Override` block to tell sub-agents they are the executor and should not delegate. However, the block is placed **after** the `> **ORCHESTRATOR GATE**` block in `skills/sdd-apply/SKILL.md`.

The current order is:

```markdown
> **ORCHESTRATOR GATE**: If you loaded this skill via the `skill()` tool, you are
> the ORCHESTRATOR — STOP. Do NOT execute these instructions inline. Delegate to
> the dedicated `sdd-apply` sub-agent...
>
> ## Executor Override
>
> If you ARE the `sdd-apply` sub-agent (NOT the orchestrator), the gate above does
> NOT apply to you. Continue with the phase work below. Do NOT delegate.
```

LLMs are autoregressive — they process instructions sequentially. When the sub-agent reads the skill, the **first** imperative instruction it encounters is "Delegate to the dedicated sdd-apply sub-agent". By the time it reaches the `Executor Override` (which says the gate doesn't apply), it has already locked onto the delegate action.

This causes the sub-agent to get confused, state that it needs to delegate to the apply sub-agent (even though it IS the apply sub-agent), and ultimately do nothing or loop.

### 🔄 Steps to Reproduce

1. Have `gentle-ai` installed with SDD multi-agent mode (v1.33.2)
2. Run `/sdd-apply` (or any SDD phase)
3. Observe the orchestrator launches the `sdd-apply` sub-agent
4. The sub-agent reads SKILL.md, sees "Delegate to the dedicated sdd-apply sub-agent" first, and attempts to delegate back to itself or gets stuck
5. This repeats or fails

### ✅ Expected Behavior

The `Executor Override` should be placed **before** the `ORCHESTRATOR GATE` so the sub-agent first reads "you are the executor — execute" before encountering the gate instruction.

Moving the override above the gate ensures the autoregressive model processes the correct instruction first:

```markdown
## Executor Override

If you ARE the `sdd-apply` sub-agent (NOT the orchestrator), the gate below does
NOT apply to you. Continue with the phase work below. Do NOT delegate.

> **ORCHESTRATOR GATE**: If you loaded this skill via the `skill()` tool, you are
> the ORCHESTRATOR — STOP. Do NOT execute these instructions inline. Delegate to
> the dedicated `sdd-apply` sub-agent...
```

### ❌ Actual Behavior

The sub-agent gets confused. It explicitly states that it has to delegate the task to the `sdd-apply` sub-agent, but since it already is the `sdd-apply` sub-agent, it ends up doing nothing.

---
## 🖥️ Environment

**Gentle AI Version:** 1.33.2
**Operating System:** Linux (Fedora/RHEL)
**AI Agent / Client:** OpenCode
**Affected Area:** Other

### 💡 Logs / Error Output

```text
(Exact logs were lost in chat history, but the observed behavior was the agent getting confused: "I need to delegate this task to the sdd-apply sub-agent" and since it was already the apply sub-agent, it resulted in no action being taken).
```

Alan-TheGentleman:
Closes #721

## Summary

This is the technically necessary maintainer successor to #1105. The original contributor PR remains open, but its fork disables maintainer edits (`maintainerCanModify=false`), so the approved fix cannot be updated in place.

- Credits @chrisssp and preserves original commit `d3f086697a62752732f170c98c9f16ca098f0b01` as an ancestor.
- The contributor's four-line ordering fix places `Executor Override` before `ORCHESTRATOR GATE` in the installed `sdd-apply` skill.
- The maintainer addition is runtime proof, not a replacement for the contributor's fix: a real pinned OpenCode executor is installed and invoked through `gentle-orchestrator`, must not receive `task`, must produce a bash nonce, and must return its nonce-bound result to the orchestrator.

## Scope

Exactly 275 changed lines across 2 files: 262 insertions and 13 deletions.

| File | Change |
|---|---|
| `internal/assets/skills/sdd-apply/SKILL.md` | Preserved contributor ordering correction. |
| `e2e/organicruntime/organic_runtime_test.go` | Added installed runtime proof and an orchestrator-spoof rejection control. |

## Validation

- [x] `git diff --check eaa7ae2186c8a2d64079c6802b7309ceaaf47aa8...59c35af29937f9141b8dab89c0f292afc7aa51f0`
- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] Pinned `opencode --version` reported `1.18.4`.
- [x] `PATH="/tmp/opencode/pr-1105-opencode-1.18.4/bin:$PATH" GENTLE_AI_REAL_AGENT_E2E=1 go test -v ./e2e/organicruntime -run '^TestRealAgentInstalledSDDApplyExecutorDoesNotDelegate$' -count=2` passed twice: 16.89s and 11.04s.
- [x] `go test -v ./e2e/organicruntime -run '^TestInstalledSDDApplyExecutorProofRejectsOrchestratorSpoof$' -count=1`
- [x] `go test -race -v ./e2e/organicruntime -run '^TestInstalledSDDApplyExecutorProofRejectsOrchestratorSpoof$' -count=1`
- [x] `GOOS=windows GOARCH=amd64 go test -c -o /tmp/opencode/pr-1105-sdd-apply-organicruntime-windows-amd64.test.exe ./e2e/organicruntime`

## Benchmark

N/A. This is instruction ordering plus a deterministic runtime regression proof; it does not change a performance-sensitive code path.

## Rollback

Revert this PR. No persisted data, schema migration, or external state requires repair.

## Review Boundaries

No RDD approval is claimed. RDD remains disabled and outside this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SDD workflow execution so dedicated phase agents proceed directly without unintended delegation.
  * Strengthened completion verification to reject spoofed, incomplete, unauthenticated, or unrelated results.
  * Improved secure command execution and result handling.

* **Tests**
  * Added end-to-end coverage for executor installation, prompt behavior, available tools, authenticated execution, and invalid command output.
  * Added consistency checks for execution-role guidance across SDD workflow phases.


MarcosArispe:
This PR matches the exact failure shape blocking our consumer workflow.

A dedicated `sdd-apply` executor loaded the installed skill, applied the orchestrator gate to itself, and stopped before task work while requesting delegation or a gate waiver. The native attempt was preserved with:

- outcome: `interrupted`
- changed lines: `0`
- next action: `begin`
- no repository, deployment, commit, push, PR, or review mutation

The ordering fix plus installed-runtime proof in this PR directly addresses that behavior.

The currently available beta build `2.2.5-0...ece470d` does not include this PR, so we cannot validate the fix without running from a source checkout, which we intentionally avoid for consumer recovery.

After merge, could you publish a tagged prerelease that includes the refreshed installed `sdd-apply` asset? We can then re-enter through native attempt status and validate one bounded P20 launch against the released build.

Acceptance from the consumer side:

1. The dedicated executor performs its assigned work without receiving or requesting the `task` tool.
2. It does not request a delegation waiver.
3. The preserved attempt resumes from `next_action: begin`.
4. No source checkout or unmerged PR is involved.

Alan-TheGentleman:
Extended this PR rather than opening a competing one, because it already carries the two things that matter: @chrisssp's original commit as an ancestor, and the real OpenCode runtime proof. Both are untouched, and the push was a fast-forward with no rebase.

Added on top, in `b7205824`: the reordering alone is not sufficient. With `Executor Override` placed first, the orchestrator now reads "you are the sub-agent, continue with the phase work, do not delegate" before reaching its own gate, so the contradiction moves to the other reader instead of disappearing. The prose was an imperative followed by its own retraction; whoever reads it has to resolve that.

Both blocks are now one `## Execution Role` block in which every imperative is preceded, never followed, by the condition that scopes it. The same defect existed in eight more skills (`sdd-archive`, `sdd-design`, `sdd-explore`, `sdd-init`, `sdd-propose`, `sdd-spec`, `sdd-tasks`, `sdd-verify`), so the fix now covers all nine `delegate_only` assets. `sdd-onboard` is deliberately untouched: it is not `delegate_only` and runs inline. Frontmatter is unchanged everywhere, since `disable-model-invocation` and `delegate_only` are the real enforcement and they already work.

There is also a ratchet now, `TestDelegateOnlySkillsStateRoleWithoutRetraction`, which the earlier state lacked: the existing checks passed through a loose "do not delegate" substring, so nothing stopped a future edit from reintroducing an instruction plus a retraction. It derives the skill list from each asset's own frontmatter rather than a hardcoded list, and asserts the property, one role block naming both roles with no retraction phrase, instead of pinning wording. I verified it independently by reintroducing the old pair in `sdd-verify`, a different skill than the author tested: it failed with the retraction named, and passed again on revert.

Eight `sdd-init` goldens regenerated through the sanctioned `-update` path. Full suites green locally: components, assets, cli, bench.

Alan-TheGentleman:
Applying `size:exception`. The diff is 725 lines but only a small part is logic: 329 of the additions are the OpenCode runtime proof in `organic_runtime_test.go`, 8 files are goldens regenerated through the sanctioned path, and the nine skill edits are the same six-line block repeated with the skill's own name substituted. The reviewable surface is one block of prose plus one ratchet test. Splitting would mean separating the contributor's credited commit from the proof that the fix works, which is the one thing this PR exists to keep together.

Alan-TheGentleman:
Live confirmation from the field while this sits in review: #2697, filed today against the published rc.2 by @Pablo-GNU, is this exact defect in `sdd-init`. Their executor read the gate, concluded it was the orchestrator, looked for a delegation primitive it does not have as the executor, and blocked with no artifacts produced.

Worth noting for the scope question: they hit `sdd-init`, not `sdd-apply`. Had this PR stayed limited to the originally reported file, that report would still be open after the merge. The nine-skill extension is what covers it.

Issue #2172, opened 2026-08-01
fix(review): negotiated START freezes an empty candidate on a clean post-apply worktree


aparragithub:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

### 📝 Bug Description

On a clean feature branch whose SDD implementation is **already committed**, the negotiated review loop freezes a candidate that reviews nothing, approves it, and only fails three gates later with a message that describes the wrong problem.

`review status --contract gentle-ai.review-integration/v2 --next-transition` proposes `review.start` with `projection: workspace`. Because the worktree is clean, START freezes `changed_files: 0`, `changed_lines: 0`, `paths: []`, and `base_tree == candidate_tree`. The zero-lens low-risk review finalizes as `approved` and `pre-commit` returns `allow`. Delivery then fails at `pre-push`, and no base selector can repair it, because the reviewed scope is empty rather than merely mis-based.

This is a distinct mechanism from #2090 (closed by #2169), where the freeze was correct and the emptiness appeared later at validation after a strict-subset correction and a recovered successor. Here the freeze itself is born empty, with no correction and no recovery involved. It is also the START-side sibling of #1203, which fixed the `post-apply` variant of "clean worktree after the approved candidate was committed".

Root trigger: `internal/cli/review_next_transition.go` proposes `review.start` with `TargetCurrentChanges` whenever target applicability is `unrelated`, with no clean-workspace check. This is the standard STATUS-driven loop, not a misuse.

### 🔄 Steps to Reproduce

1. Create a feature branch with **three commits** ahead of `origin/main`, and leave the worktree clean (`git status` reports nothing pending).
2. Run negotiated STATUS and execute exactly the `review.start` transition it returns:
   ```
   gentle-ai review status --cwd <repo> --contract gentle-ai.review-integration/v2 --next-transition
   ```
3. Observe the frozen authority: `projection: workspace`, `changed_files: 0`, `changed_lines: 0`, `paths: []`, and equal base/candidate trees.
4. Finalize the zero-lens review and validate `pre-commit`. Observe `allow`.
5. Validate delivery without a base selector:
   ```
   gentle-ai review validate --cwd <repo> --lineage <lineage-id> --gate pre-push
   ```
[...]

PR #2239, opened 2026-08-02
fix(review): consult the kill switch before pre-PR receipt composition

### LINKED ISSUE #1878: fix(review): disabled pre-PR still enters receipt composition
### Pre-flight Checklist

- [x] I searched open and closed issues for disabled review mode, `pre-pr`, `receipt-composition`, `invalid-chain`, and the exact denial text. This is not a duplicate.
- [x] I understand that PRs require `status:approved` before implementation.

### Bug Description

Gentle AI `2.2.0-rc.1` Refresh 7 still blocks the `pre-pr` gate when review mode is explicitly disabled for the clone.

The effective mode is correctly reported as `off` with source `clone_local`. `pre-commit` and `pre-push` can route operational delivery through `disabled/unmanaged`, but `pre-pr` enters compact receipt composition first and exits non-zero with `invalid-chain`.

This contradicts the prerelease kill-switch contract: disabling review must never fabricate approval, but it must allow operational delivery to fall through to ordinary repository policy instead of requiring a receipt chain.

### Steps to Reproduce

1. Use the verified Refresh 7 Linux AMD64 release asset.
2. In a Git repository containing historical compact-v2 review authority but no receipt governing the current branch, disable review for the clone:

   ```bash
   gentle-ai review mode disable --cwd /path/to/repo --scope clone --json
   ```

3. Confirm:

   ```bash
   gentle-ai review mode status --cwd /path/to/repo --json
   ```

   returns `effective: off`, `source: clone_local`.
4. Commit and push the current branch under ordinary repository policy.
5. Run:

   ```bash
   gentle-ai review validate --gate pre-pr --cwd /path/to/repo
   ```

6. Observe receipt composition denial and exit code 1.

### Expected Behavior

With review mode effectively off and no receipt governing the current delivery, `pre-pr` should behave like the documented disabled gate path:

```json
{
  "allowed": false,
  "action": "repository-policy",
  "delivery": "disabled/unmanaged"
}
```

The command should exit 0, must not claim `allow`, must not create or mutate authority, and must leave PR policy to the repository.

Real authority corruption should continue to fail closed.

### Actual Behavior

The kill switch is not consulted before pre-PR composition:

```json
{
  "result": "invalidated",
  "allowed": false,
  "action": "explicit-maintainer-action",
  "reason": "compact pre-PR receipt composition denied: no receipt chain contains at least two authoritative approved members",
  "context": {
    "gate": "pre-pr",
    "denial": {
      "stage": "receipt-composition",
      "code": "invalid-chain"
    }
  }
}
```

The command exits 1 and operational PR creation remains blocked despite effective mode `off`.

### Gentle AI Version

```text
gentle-ai 2.2.0-rc.1
Refresh 7 Linux AMD64 SHA-256:
1923d7b33f647b25bd2a246095b7aca691351621f28ef4c3ac4dfa877054ac60
```

### Operating System

Linux (Ubuntu/Debian)

### AI Agent / Client

OpenCode

### Affected Area

CLI (commands, flags)

### Additional Context

- The reproduction was repeated after replacing an older same-tag asset and verifying the exact current Refresh 7 digest.
- `pre-commit` in the same clone returned exit 0 with `delivery: disabled/unmanaged` when called without an explicit lineage.
- `pre-push` also proceeded operationally under the disabled mode.
- Repository source and review authority were not modified during the failing `pre-pr` call.
- The real repository identity, branch names, customer data, receipt payloads, and full authority identifiers are omitted.

Related but not duplicate:

- #1830 concerns disabled `pre-push` after historical publication.
- #1782 concerns incorrect composition of an otherwise valid reviewed linear chain while review is active.
- This report concerns kill-switch ordering: when effective mode is off, `pre-pr` should not require receipt composition at all.

Suggested regression coverage:

```text
TestPrePRDisabledModeReturnsUnmanagedBeforeReceiptComposition
TestPrePRDisabledModeDoesNotFabricateAllow
TestPrePRDisabledModeStillFailsClosedForUnreadableModePolicy
```

aparragithub:
Closes #1878

## Problem

pre-PR was the one lifecycle gate that vetoed delivery while the user's kill switch was off. Selector-free compact chain composition ran *before* the disabled branch, so an `invalid-chain` denial returned through `emitFacadeGateEvaluationNegotiated` before `emitDisabledUnmanagedDelivery` could ever be reached. pre-commit and pre-push on the same repository already reported `disabled/unmanaged` and exited 0.

That contradicted the comment sitting directly above the disabled branch, which states the kill switch is consulted for every discovery outcome.

## Change

Both composition call sites in `internal/cli/review_facade.go` are now guarded by the existing `reviewDrivenDevelopmentDisabled` helper:

- `2908-2913`, the discovery-success path, reached when an exact receipt is discovered and denies with `receipt-binding` / `base-mismatch`. This one was not covered by the analysis in the issue, and gating only the other call site leaves the defect reachable.
- `2917-2922`, the discovery-failure path originally reported.

Composition traverses provider-owned historical authority, so it is receipt-driven development doing work. With the switch off that work must not happen at all, and control flow then falls through to the existing disabled branch unchanged.

This cannot relax anything: `reviewDrivenDevelopmentDisabled` fails closed to enabled on an unreadable or tampered mode record, and with reviews on every composition path is byte-identical.

## Tests

`internal/cli/review_disabled_prepr_composition_test.go` adds both halves of the decision over one shared fixture:

- **disabled** — reports `disabled/unmanaged`, exits 0, fabricates no approval, emits no `receipt-composition` denial, and stays replay stable.
- **enabled** — the identical repository still composes and still denies with `receipt-composition` / `invalid-chain`, with today's exact field set, so the change cannot silently widen into the enabled path.

The fixture builds two reviewed lineages plus one commit no receipt covers. Two is load-bearing: `internal/reviewtransaction/compact_chain.go:155` returns not-applicable below two compact stores, so a single-receipt fixture never reaches composition and would pass with or without this fix.

Verified the disabled test fails on `main` with exactly the reported envelope:

```
review lifecycle gate denied: invalidated: compact pre-PR receipt composition denied:
no exact compact receipt chain covers the selected base through HEAD
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled reviews now correctly report as unmanaged without running receipt validation or blocking delivery.
  * Prevented composition errors and pre-PR denials when receipt-driven development is disabled.
  * Preserved existing receipt validation and invalid-chain behavior when reviews are enabled.

### DIFF
--- a/internal/cli/review_disabled_prepr_composition_test.go
+++ b/internal/cli/review_disabled_prepr_composition_test.go
@@ -1,0 +1,148 @@
+package cli
+
+import (
+	"bytes"
+	"errors"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/gentleman-programming/gentle-ai/v2/internal/reviewtransaction"
+)
+
+// reviewPrePRHistoricalAuthorityRepo builds the exact shape issue-1878 reports:
+// a repository that already carries compact review authority from earlier work,
+// plus a later commit no receipt covers. Selector-free pre-PR reaches compact
+// chain composition on this shape, which is what makes it the useful fixture
+// for both halves of the kill-switch decision.
+func reviewPrePRHistoricalAuthorityRepo(t *testing.T) string {
+	t.Helper()
+	repo := initReviewCLIRepo(t)
+	branch := strings.TrimSpace(runReviewCLIGit(t, repo, "symbolic-ref", "--short", "HEAD"))
+	if err := os.WriteFile(filepath.Join(repo, "tracked.txt"), []byte("published base\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	runReviewCLIGit(t, repo, "add", "tracked.txt")
+	runReviewCLIGit(t, repo, "commit", "-qm", "published base")
+	// The remote is pinned here, so pre-PR derives its publication base from
+	// this commit.
+	configureCLIReviewPublicationRemote(t, repo, branch)
+
+	// Two reviewed lineages. Composition is only attempted from two compact
+	// stores upward (compact_chain.go: len(stores) < 2 is not applicable), so a
+	// single receipt would never exercise the path this issue is about.
+	for _, step := range []string{"first reviewed slice", "second reviewed slice"} {
+		base := strings.TrimSpace(runReviewCLIGit(t, repo, "rev-parse", "HEAD"))
+		if err := os.WriteFile(filepath.Join(repo, "tracked.txt"), []byte(step+"\n"), 0o644); err != nil {
+			t.Fatal(err)
+		}
+		runReviewCLIGit(t, repo, "add", "tracked.txt")
+		runReviewCLIGit(t, repo, "commit", "-qm", step)
+		finalizeFacadeReviewForRepo(t, repo, "--base-ref", base, "--committed-only")
+	}
+	return repo
+}
+
+// TestReviewValidateReportsDisabledUnmanagedDeliveryAtPrePRWithHistoricalAuthority
+// closes issue-1878. pre-PR was the one gate that vetoed delivery while the
+// user's kill switch was off: selector-free composition ran before the disabled
+// branch, so an `invalid-chain` denial returned before `disabled/unmanaged`
+// could ever be reported. pre-commit and pre-push on this same repository
+// already report; pre-PR must now agree with them.
+func TestReviewValidateReportsDisabledUnmanagedDeliveryAtPrePRWithHistoricalAuthority(t *testing.T) {
+	reviewModeHome(t)
+	repo := reviewPrePRHistoricalAuthorityRepo(t)
+
+	disableReviewForClone(t, repo)
+
+	// Work authored and committed while disabled. No receipt can cover it, so
+	// the composed chain cannot reach HEAD — the exact denial the user hits.
+	if err := os.WriteFile(filepath.Join(repo, "tracked.txt"), []byte("work authored while disabled\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	runReviewCLIGit(t, repo, "add", "tracked.txt")
+	runReviewCLIGit(t, repo, "commit", "-qm", "authored while disabled")
+
+	var output bytes.Buffer
+	err := RunReviewFacadeValidate([]string{"--cwd", repo, "--gate", string(reviewtransaction.GatePrePR)}, &output)
+	// The gate reports; it does not veto.
+	if err != nil {
+		t.Fatalf("disabled pre-PR vetoed delivery instead of reporting it: %v\n%s", err, output.String())
+	}
+	var result ReviewValidateResult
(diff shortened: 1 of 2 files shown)


aparragithub:
I don't have permission to apply labels on this repo, so the **Check PR Has type:\* Label** gate needs a maintainer to add `type:bug`. Everything else is in place: linked issue #1878 already carries `status:approved`, and the diff is 165 changed lines, well inside the 400-line budget.

Alan-TheGentleman:
The two-halves test design here is the right shape for a kill-switch fix: prove the disabled side reports, and prove the enabled side keeps composing and keeps denying exactly as before. That is what makes a fix like this auditable.

The ground moved underneath it. Wave 5 of #2462 deleted selector-free pre-PR chain composition outright. `EvaluateCompactPrePRChain` and both of its call sites in `runReviewFacadeValidate` are gone, and `internal/cli/review_pre_pr_composition_deletion_test.go` is now an AST scan that fails if any production file calls it again.

The outcome issue #1878 wanted arrived in that same wave by a different route: the kill switch is consulted exactly once, before any authority read, at `internal/cli/review_facade.go:3122`, so `pre-pr` reports `disabled/unmanaged` before composition would ever have been reached. The guard was correct. The traversal it guarded no longer exists, which also means #1878 can be closed as fixed.

aparragithub:
Closing this as obsolete. The defect it fixes is already gone from `main`, resolved structurally rather than by the guards proposed here.

I checked before assuming either way. Bringing only `review_disabled_prepr_composition_test.go` onto current `main` (`7aa971d7`), with none of this branch's production changes applied, `TestReviewValidateReportsDisabledUnmanagedDeliveryAtPrePRWithHistoricalAuthority` passes. That test builds historical authority, disables reviews for the clone, commits work while disabled, then asserts the pre-PR gate reports `disabled/unmanaged`, carries no `receipt-composition` denial, and is replay stable. It compiled against helpers already in `main`, so the scenario really ran.

What makes it interesting: `main` does not contain the guards this PR adds. `git grep reviewDrivenDevelopmentDisabled` over `main`'s `internal/cli/review_facade.go` returns nothing, and the kill switch is consulted only from `sdd_attempt.go` and `sdd_status.go`. The behavior was fixed by the RDD root simplification (#2462) and the `rdd-wave7` series, all landing after this branch's merge-base (`545485db`). A better fix than the targeted one proposed here, and it makes these guards redundant.

One leftover worth a maintainer's eye, filed nowhere yet because I am not sure it is a defect. The enabled-path companion test still fails on `main`, but for a changed error taxonomy rather than the original bug:

```
TestReviewValidatePrePRKeepsComposingHistoricalAuthorityWhileEnabled
  got:  receipt-discovery / receipt_ambiguous
  want: receipt-composition / invalid-chain
```

With reviews on, that repository shape now denies earlier, at discovery, as ambiguous. If that reordering is intended, nothing to do. If a historical-authority chain is meant to reach composition and be reported as an invalid chain, then something regressed and it deserves its own issue. Say the word and I will file it with a reproduction.

Thanks for the original review on #2172-era work. This branch has served its purpose.

PR #2288, opened 2026-08-02
fix(tui): warn when custom models lack tool calls

### LINKED ISSUE #934: Local Ollama models silently excluded from SDD phase picker — missing tool_call flag not documented
### Pre-flight Checklist

### 📝 Bug Description

When a local Ollama provider is correctly defined in `opencode.json`, its models do not appear in
  the SDD phase model picker. The TUI only lists `ollama-cloud` (a paid cloud service), with no
  indication that local models are supported or why they are not shown.

  Related to #375 (LM Studio, v1.23.0). Unlike that report, in v1.40.2 `opencode.json` IS already
  read via `LoadConfigProviders` + `MergeCustomProviders` — that part is fixed. The remaining
  friction is a different, undocumented requirement.

### 🔄 Steps to Reproduce

1. Configure a local Ollama provider in ~/.config/opencode/opencode.json:

    ```
 "provider": {
       "ollama": {
         "name": "Ollama (local)",
         "npm": "@ai-sdk/openai-compatible",
         "options": { "baseURL": "http://localhost:11434/v1" },
         "models": {
           "qwen3-coder:30b": { "name": "Qwen3 Coder 30B" }
         }
       } 
     }
```
     
  2. Run gentle-ai → OpenCode SDD Profiles → Create new profile
  3. Reach the provider selection screen
2. 
3. 

### ✅ Expected Behavior

The "ollama" provider appears in the picker with the configured local models available for assignment to SDD phases.

### ❌ Actual Behavior

"ollama" does not appear. Only "ollama-cloud" is shown (a separate cloud provider requiring OLLAMA_API_KEY). The local models are silently excluded with no error or warning.

### Gentle AI Version

1.40.2

### Operating System

Linux (Ubuntu/Debian)

### AI Agent / Client

OpenCode

### 📋 Affected Area

TUI (terminal UI)

### 💡 Logs / Error Output

```shell
No error output — the exclusion is silent.
```

### Additional Context

Root cause found in internal/opencode/models.go:

```
    func FilterModelsForSDD(provider Provider) []Model {
        var models []Model
        for _, m := range provider.Models {
            if m.ToolCall {   // false by default in Go → all custom models excluded
                models = append(models, m)
            }   
        }
        return models
    }   

    type ConfigModel struct {
        Name     string `json:"name"`
        ToolCall bool   `json:"tool_call"`   // not mentioned in any docs
    }
```
    
  The ConfigProvider struct only reads "name" and "tool_call" per model. Since "tool_call"
  defaults to false, any model in opencode.json without that field is invisible to the SDD picker.

---

**Workaround:** 
add "tool_call": true to each model entry in opencode.json that supports tool calling.
  This immediately fixes the issue — the provider and models appear in the picker on the next
  gentle-ai run.

 **Suggested fix options (in order of effort):**

  A. Docs: add a "Local Ollama / LM Studio" section to docs/opencode-profiles.md explaining the
     tool_call requirement with a minimal example.

  B. TUI warning: when a provider has models but none pass FilterModelsForSDD, show a hint
     instead of silently omitting the provider.

  C. Auto-detect: query GET /v1/models on local providers at startup and infer tool_call support
     from model family (Qwen, Llama, Gemma → true; embedding/OCR → false).

Matere413:
## 🔗 Linked Issue

Closes #934

---

## 🏷️ PR Type

- [ ] `type:feature` — New feature
- [ ] `type:refactor` — Code refactoring

---

## 📝 Summary

- Warns when a custom provider defines models but none declare `tool_call: true`.
- Keeps those models unavailable instead of weakening SDD capability filtering.
- Covers state and rendered warning behavior with focused tests.

---

## 📂 Changes

| `internal/tui/screens/model_picker.go` | Adds deterministic non-blocking warnings for affected custom providers |
| `internal/tui/screens/model_picker_test.go` | Verifies warning rendering and that unsupported models remain unavailable |

---

## 🧪 Test Plan

- [x] `git diff --check`
- [x] `go test ./internal/tui/screens -run "^Test(NewModelPickerState|RenderModelPickerShowsConfigWarning)$" -count=1`
- [x] `go test ./internal/tui/screens`
- [x] `go test ./internal/opencode`
- [x] `go test ./... -run "^$"`
- [ ] Full E2E suite, not applicable to this focused TUI warning and not run locally

Benchmark validation: N/A. This change does not touch review lifecycle, gates, recovery, delivery, or benchmark code.

Receipt-driven development was disabled for this clone after the installed OpenCode reviewer transport returned `unsupported-capability`; delivery is therefore `disabled/unmanaged`. Focused verification and repository compilation passed.

---

## ✅ Contributor Checklist

- [ ] Exactly one `type:*` label is applied (maintainer action required: `type:bug`)
- [x] Go format passes
- [x] Documentation is unchanged because `tool_call` is already documented
- [x] Commits contain no `Co-Authored-By` trailers

---

## 💬 Notes for Reviewers

The warning is intentionally informational. It does not auto-enable models, infer model capabilities, query providers, or modify `FilterModelsForSDD`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clearer configuration warnings when custom providers have models but none support the required tool-calling capability.
  * Providers without configured models or with at least one compatible model no longer generate unnecessary warnings.
  * Providers without compatible models remain listed but are marked unavailable.
  * Warnings are preserved alongside existing configuration notices and displayed consistently in a predictable order.
  * Clarified the LM Studio warning by specifying the required `"tool_call": true` setting and preventing duplicate notices.

### DIFF
--- a/internal/tui/screens/model_picker.go
+++ b/internal/tui/screens/model_picker.go
@@ -138,6 +138,7 @@ func NewModelPickerState(cachePath string, settingsPath string) ModelPickerState
 		customProviderIDs: customIDs,
 	}
 	state.refreshAvailableModels()
+	state.ConfigWarning = appendCustomProviderToolCallWarnings(state.ConfigWarning, providers, configProviders)
 	return state
 }
 
@@ -205,6 +206,42 @@ func (state *ModelPickerState) refreshAvailableModels() {
 	for _, id := range state.AvailableIDs {
 		state.SDDModels[id] = opencode.FilterModelsForSDD(state.Providers[id])
 	}
+}
+
+func appendCustomProviderToolCallWarnings(
+	existing string,
+	providers map[string]opencode.Provider,
+	configProviders map[string]opencode.ConfigProvider,
+) string {
+	ids := make([]string, 0, len(configProviders))
+	for id := range configProviders {
+		ids = append(ids, id)
+	}
+	sort.Strings(ids)
+
+	for _, id := range ids {
+		configProvider := configProviders[id]
+		if len(configProvider.Models) == 0 {
+			continue
+		}
+
+		provider, ok := providers[id]
+		if !ok || len(opencode.FilterModelsForSDD(provider)) > 0 {
+			continue
+		}
+
+		name := provider.Name
+		if name == "" {
+			name = id
+		}
+		existing = appendConfigWarning(existing, fmt.Sprintf(
+			`Custom provider %q has models, but none declare tool_call: true. Add tool_call: true to at least one model in provider.%s.models.`,
+			name,
+			id,
+		))
+	}
+
+	return existing
 }
 
 func appendConfigWarning(existing, warning string) string {
--- a/internal/tui/screens/model_picker.go
+++ b/internal/tui/screens/model_picker.go
@@ -194,7 +194,7 @@ func (state ModelPickerState) Update(msg tea.Msg) ModelPickerState {
 	state.customProviderIDs = append(state.customProviderIDs, "lmstudio")
 	state.refreshAvailableModels()
 	if len(discovery.Models) > 0 && len(opencode.FilterModelsForSDD(provider)) == 0 {
-		state.ConfigWarning = appendConfigWarning(state.ConfigWarning, "LM Studio models need tool_call: true in provider.lmstudio.models for SDD.")
+		state.ConfigWarning = appendConfigWarning(state.ConfigWarning, "LM Studio models need \"tool_call\": true in provider.lmstudio.models for SDD.")
 	}
 	return state
 }
@@ -235,7 +235,7 @@ func appendCustomProviderToolCallWarnings(
 			name = id
 		}
 		existing = appendConfigWarning(existing, fmt.Sprintf(
-			`Custom provider %q has models, but none declare tool_call: true. Add tool_call: true to at least one model in provider.%s.models.`,
+			`Custom provider %q has models, but none declare "tool_call": true. Add "tool_call": true to at least one model in provider.%s.models.`,
 			name,
 			id,
 		))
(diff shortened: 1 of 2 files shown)


Matere413:
Addressed the review findings in `83d1e18a`: the warning now uses valid JSON syntax, the table comment is accurate, and tests cover deterministic multi-provider ordering plus preservation/rendering of existing warnings. Focused packages, gofmtcheck, diff check, and repository compilation pass.

A maintainer still needs to apply the required `type:bug` label; contributor permissions do not allow me to add it.

Alan-TheGentleman (changes requested):
Thanks for preserving capability filtering and avoiding duplicate LM Studio warnings. One path remains: the generic warning skips `lmstudio`, but a discovery error returns after saying configured models are used even when every configured model is excluded for missing `"tool_call": true`. Emit the capability guidance on that error path without duplication and add the regression. The existing model-picker tests do not cover this branch.

Matere413:
Please re-review head `75ca5612`. The review feedback has been addressed, and all required checks are green.

Alan-TheGentleman (approved):
Approving. Merged current main into your head and ran both modules; root and `bench` are green.

The refactor is the right shape: the LM Studio warning becomes `appendLMStudioToolCallWarning` with a `strings.Contains` dedup so the two call sites cannot emit it twice, and the general case is a sibling function rather than a special case bolted onto the first.

Two details I want to name because they are easy to omit and you did not:

- `sort.Strings(ids)` before iterating `configProviders`. Map iteration order in Go is randomized, so without this the warning block would reorder between runs and any golden or snapshot over it would flake intermittently. That is the kind of flake that costs a day to chase.
- The `name == ""` fallback to the provider id, so a provider with no display name still produces an actionable message instead of `Custom provider ""`.

The message names the exact config path to edit (`provider[<id>].models`), which is what makes it a refusal someone can act on rather than a notice.

Issue #1809, opened 2026-07-24
bug(state): install and sync serialize the full InstallState without a lock, risking concurrent-write clobber


pablontiv:
## Objective

Prevent lost updates in the shared `~/.gentle-ai/state.json` document.

`state.Write` and `state.WriteReconciled` publish the complete `InstallState`. Atomic replacement prevents partial JSON, but it does not serialize a read-modify-write transaction. Every cooperative mutation must therefore:

1. derive the same canonical lock identity,
2. acquire that lock before reading the latest state,
3. mutate only the fields it owns,
4. publish while the lock is still held, and
5. preserve rollback, corruption, and fail-open or fail-closed behavior owned by that call path.

## Audit boundary

This body reflects `main` at `59a742a55653f45d314a728816646a0a8664a9f7`, collected at `2026-09-02T21:34:47Z` and reconciled at `2026-09-02T21:40:44Z`.

The original report was filed against v2.1.11 on macOS. It correctly distinguished atomic publication from transaction serialization and identified stale whole-document writes in install and sync. Later source inspection broadened the cluster to every install-state writer.

## Current status

The parent remains open. Four production read-modify-write owners are coordinated on the frozen revision, while seven production owners still call the whole-document sinks without the canonical lock.

### Delivered and proven

- [x] **CLI install final persistence** is coordinated in `internal/cli/run.go:308-327`. The lock covers the fresh read/merge through reconciled publication. Named tests: `TestPersistInstallStateMergesExplicitAgentsFromLatestState` and `TestPersistInstallStateFullInstallPreservesUnrelatedFields`.
- [x] **Final sync managed-state persistence** is coordinated in `internal/cli/sync.go:1678-1722`. Named test: `TestPersistSyncManagedAssetStateReReadsLatestStateAfterLockContention`.
- [x] **Global review-mode mutation** is coordinated in `internal/cli/review_mode.go:429-450`, delivered through PR #3775. Named test: `TestWriteGlobalRDDModeSerializesWithInstallStateAndPreservesFreshFields`.
- [x] **Update cooldown persistence** is coordinated in `internal/update/cooldown.go:93-108`, delivered by PR #4083 at this frozen revision and scoped through child issue #4086. Named tests: `TestCheckAllWithCooldown_ConcurrentReviewModeDisablePreservesMode` and `TestCheckAllWithCooldown_ContendedPersistenceRejectsReviewModeDisable`.

[...]

PR #2299, opened 2026-08-03
fix(state): serialize install-state writes behind an advisory lock

### LINKED ISSUE #2352: feat(state): add a locked read-modify-write primitive for install state
## Parent initiative

Part of #1809.

Slice 2 of 3. The parent issue #1809 remains open until the call-site migration lands,
which is the slice that actually fixes the reported bug.

---

### Pre-flight Checklist

### 🔍 Affected Area

Installation Pipeline

### 💡 Problem Statement

`internal/state` offers no way to mutate persisted install state safely. `state.Write` marshals a whole `InstallState` and atomically replaces `state.json`, with no lock, no revision, and no compare-and-swap. Callers therefore read, modify in memory, and write back, with an unguarded window between the read and the write, and the last writer wins with whatever snapshot it held.

That window is what #1809 reports. Fixing #1809 means migrating every persistence call site off read-then-`Write`, and there is nothing for them to migrate *onto*.

### 📦 Proposed Solution

Add `state.Update(homeDir, mutate)`:

- takes the exclusive advisory lock from the platform primitive slice,
- re-reads persisted state **inside** the lock,
- applies the caller's mutation to that fresh copy,
- writes while still holding the lock.

`state.Write` stays available so nothing breaks while call sites migrate.

Acceptance criteria for this issue, separate from #1809:

- `Update` serializes concurrent mutations across processes, verified under `-race`.
- The lost-update sequence from #1809 fails before the change and passes after: actor A reads state, actor B writes `rdd_mode=off`, actor A writes `persona` from its stale snapshot and silently restores `rdd_mode=on`.
- The lock contract is documented, including that it is **not** reentrant. Acquisition opens a new descriptor per call and both `flock` and `LockFileEx` are per-handle, so a `mutate` closure that calls back into `Update` would deadlock.
- `Write` and `Update` interacting on the same state file is covered by a test.

**Behaviour change to call out.** `state.Write` ends up taking the same lock and can return a lock-timeout error it never returned before. That is a change to an existing exported function, not just new surface, so callers that ignore its error or assume it cannot block deserve a second look during review.

This issue does **not** cover the platform lock helpers, which are their own slice, nor migrating any call site, which stays on #1809.

### 🔄 Alternatives Considered

**Make `Write` itself locking instead of adding `Update`.** Rejected, and this is the load-bearing one: a write-only lock does not close the race. The stale snapshot is captured at read time, so serializing only the write still lets a stale full-object write land. That is exactly the failure mode #1809 reports, and it is why the primitive has to own the read as well.

**Lock at each call site instead of adding a primitive.** Rejected: eight independent lock implementations is the shape that produced the three disagreeing `_shared` file lists tracked in #1806.

**Bundle this with the platform lock helpers and the call-site migration in one PR under `size:exception`.** Rejected: combined it is roughly 660 changed lines, and the reviewer asked for the incremental slices to stay inside the budget rather than for an exception.

### 📎 Additional Context

Slice 2 of 3 under #1809. Implementation is ready and verified: 389 changed lines against the platform-lock slice, with `go build`, `go vet`, `gofmtcheck`, `go test -race` and a Windows build all passing.

Following the pattern already merged here, where each slice closes its own approved child issue and the parent stays open. See #1225, which closes #1776 while remaining "Part of #1197", and the #2046 / #2061 pair.

The chain: platform lock primitive, then this, then #1809 for the call-site migration, which is the slice that actually closes the bug.

I have no triage rights on this repository. Maintainer-owed: `status:approved` before PR #2299 can pass its CI gates. It is currently the direct cause of that PR's failing `Check Issue Has status:approved` run.

pablontiv:
## 🔗 Linked Issue

Closes #2352

## 🏷️ PR Type

---

## 📝 Summary

Slice A of a two-PR chain for #1809. Adds the primitive; slice B migrates the call sites.

`state.Write` marshalled the whole `InstallState` and atomically replaced `state.json` with no lock, no revision, and no compare-and-swap, so the last writer won with whatever snapshot it had read. Atomic replacement kept the file valid JSON while silently accepting a stale full-object write.

This PR adds `state.Update(homeDir, mutate)`, which takes an exclusive advisory lock on a sidecar lock file, re-reads the persisted state **inside** the lock, applies the caller's mutation to that fresh copy, and writes while still holding the lock. `state.Write` is unchanged so nothing breaks; call sites move over in slice B.

The regression test reproduces the sequence from the issue: actor A reads state, actor B writes `rdd_mode=off`, actor A then writes `persona` from its stale snapshot. Before this change the test fails by silently restoring `rdd_mode=on`.

Locking is split by platform so both are exercised: `lock_unix.go` and `lock_windows.go`.

---

## 📂 Changes

|-------------|-------------|
| `internal/state/state.go` | `Update` primitive: lock, re-read, mutate, write under lock |
| `internal/state/lock_unix.go` | Advisory lock acquisition on Unix |
| `internal/state/lock_windows.go` | Advisory lock acquisition on Windows |
| `internal/state/state_lock_test.go` | Lost-update regression plus concurrency coverage |

---

## 🧪 Test Plan

```bash
go test ./internal/state/... -race -count=1
go run ./internal/gofmtcheck
go build ./...
go vet ./...
```

- [x] Unit tests pass for `internal/state`, including `-race`
- [x] Go format passes
- [x] `go build ./...` and `go vet ./...` clean
- [ ] Full `go test ./...` — deferring to CI. Locally `internal/cli` needs `-timeout 30m` and fails one unrelated `uv` test that reproduces identically on the base tree.
- [ ] E2E tests — deferring to CI
- [x] Manually verified the lost-update sequence fails before the change and passes after

Benchmark validation: N/A. This touches install-state persistence only, not review-lifecycle, gates, recovery, delivery, or benchmark surfaces.

- [x] PR stays within 400 changed lines (391 additions + 9 deletions = 400)

---

## 💬 Notes for Reviewers

**This slice does not fix the user-visible bug on its own.** It adds the primitive and pins the regression. The eight read-modify-write call sites migrate in the stacked follow-up, which is what actually closes the race in `install` and `sync`.

Split because the combined change is 528 changed lines, past the 400 budget. Each slice sits inside it, so no `size:exception` is requested.

Chain: this PR, then `pablontiv/sdd-1809-state-lock-callsites`.

I have no triage rights on this repository. Maintainer-owed label: `type:bug`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple processes update installation state concurrently.
  * Prevented failed, interrupted, or invalid updates from overwriting existing state.
  * Added timeout handling when installation state remains locked.
  * Ensured locks are released after errors or interruptions across supported platforms.
  * Preserved existing state when updates fail or state data is invalid.
  * Safely initialized missing installation state when needed.
  * Coordinated all state writes to prevent updates from being lost.

### DIFF
--- a/.deadcode-baseline.txt
+++ b/.deadcode-baseline.txt
@@ -219,8 +219,7 @@ internal/sddstatus/status.go	reportLineHasPassSignal
 internal/sddstatus/status.go	reportTextIsClearlyPassing
 internal/sddstatus/status.go	reportValueIsBenign
 internal/sddstatus/status.go	stripMarkdownSignal
-internal/state/lock_unix.go	tryLockFile
-internal/state/lock_unix.go	unlockFile
+internal/state/state.go	Update
 internal/storage/bytes.go	FormatBytes
 internal/system/install_deps.go	InstallCommandsForDep
 internal/system/install_deps.go	installCommandsBrew
--- a/.deadcode-baseline.txt
+++ b/.deadcode-baseline.txt
@@ -219,6 +219,8 @@ internal/sddstatus/status.go	reportLineHasPassSignal
 internal/sddstatus/status.go	reportTextIsClearlyPassing
 internal/sddstatus/status.go	reportValueIsBenign
 internal/sddstatus/status.go	stripMarkdownSignal
+internal/state/lock_unix.go	tryLockFile
+internal/state/lock_unix.go	unlockFile
 internal/storage/bytes.go	FormatBytes
 internal/system/install_deps.go	InstallCommandsForDep
 internal/system/install_deps.go	installCommandsBrew
--- a/internal/state/lock_test.go
+++ b/internal/state/lock_test.go
@@ -1,0 +1,51 @@
+package state
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestTryLockFileReportsBusyForSecondHandle covers the platform primitive
+// directly: a second open handle on an already-locked file reports busy rather
+// than returning an error or blocking.
+func TestTryLockFileReportsBusyForSecondHandle(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "lock")
+
+	first, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, 0o644)
+	if err != nil {
+		t.Fatalf("open first handle: %v", err)
+	}
+	defer first.Close()
+	locked, err := tryLockFile(first)
+	if err != nil {
+		t.Fatalf("lock first handle: %v", err)
+	}
+	if !locked {
+		t.Fatal("first handle could not take a free lock")
+	}
+
+	second, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, 0o644)
+	if err != nil {
+		t.Fatalf("open second handle: %v", err)
+	}
+	defer second.Close()
+	locked, err = tryLockFile(second)
+	if err != nil {
+		t.Fatalf("lock second handle: %v", err)
+	}
+	if locked {
+		t.Error("second handle took a held lock; the lock does not exclude")
+	}
+
+	if err := unlockFile(first); err != nil {
+		t.Fatalf("unlock first handle: %v", err)
+	}
+	locked, err = tryLockFile(second)
+	if err != nil {
+		t.Fatalf("relock second handle: %v", err)
+	}
+	if !locked {
+		t.Error("second handle could not take the lock after release")
+	}
+}
--- a/internal/state/lock_test.go
+++ b/internal/state/lock_test.go
@@ -1,3 +1,5 @@
+//go:build darwin || dragonfly || freebsd || illumos || linux || netbsd || openbsd || windows
+
 package state
 
 import (
--- a/internal/state/lock_unix.go
+++ b/internal/state/lock_unix.go
@@ -1,0 +1,33 @@
+// The `unix` pseudo-tag also matches aix and solaris, neither of which defines
+// syscall.Flock, and hurd, which has no port in this toolchain. Listing the
+// verified targets keeps the constraint honest and fails closed on new ports.
+// android satisfies the linux tag and ios satisfies darwin, so both are covered.
+//go:build darwin || dragonfly || freebsd || illumos || linux || netbsd || openbsd
+
+package state
+
+import (
+	"errors"
+	"os"
+	"syscall"
+)
+
+// tryLockFile attempts to take an exclusive advisory lock without blocking.
+// It returns (true, nil) when the lock was taken, (false, nil) when another
+// holder has it, and (false, err) on any other failure.
+// Adapted from internal/reviewtransaction/store_lock_unix.go.
+func tryLockFile(file *os.File) (bool, error) {
+	err := syscall.Flock(int(file.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
+	if err == nil {
+		return true, nil
+	}
+	if errors.Is(err, syscall.EWOULDBLOCK) || errors.Is(err, syscall.EAGAIN) {
+		return false, nil
+	}
+	return false, err
+}
+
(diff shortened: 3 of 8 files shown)


dnlrsls (changes requested):
This is a useful foundation, but it does not fix #1809 on its own, as the PR body correctly states. Please do not merge it with `Closes #1809`, because that would close the issue before the production call sites are migrated. Please update the linkage and resolve the remaining locking-contract gaps, especially unsupported Unix build targets, non-reentrancy documentation, and direct `Write`/`Update` interaction coverage, before this becomes the chain foundation.

dnlrsls:
Chain plan confirmed: stacked to `main`. Please treat this as the foundation slice, remove the closing keyword for #1809, and address the requested changes. Once this slice is ready and merged, #2300 can be rebased onto the updated `main` so its incremental diff is reviewed independently.

pablontiv:
Rebased onto current `main` and pushed the remaining requested change.

**Resolved in this push**

`003a2d5c` adds `internal/state/lock_unsupported.go`. The `unix` pseudo-tag matches aix and solaris, which do not define `syscall.Flock`, and hurd, which has no port in this toolchain, so the allowlist in `lock_unix.go` deliberately excludes them — but that left `tryLockFile`/`unlockFile` undefined there and the package failed to build. The new file covers the exact complement of the two real allowlists and fails closed: both functions return `ErrLockUnsupported`, so `acquireStateLock` stops rather than writing state unlocked. `lock_test.go` now carries the union of the two real allowlists, and `lock_unsupported_test.go` type-checks the fallback under cross-compilation. Verified with `GOOS=aix GOARCH=ppc64 go build|vet|test -c ./internal/state/`.

**Addressed on 2026-08-03, still awaiting re-review**

- `Closes #1809` removed — the body links `Closes #2352`; #1809 closes with the call-site migration in #2300.
- Non-reentrancy documented in the doc comments of both `Update` and `Write`: every acquisition opens a fresh file description, so a nested call self-blocks until `lockTimeout` and returns `ErrLockTimeout`.
- Direct `Write`/`Update` interaction coverage: `TestWriteWaitsForUpdateLock` asserts a whole-document `Write` cannot land inside an `Update` read-modify-write window.

**Also in this push**

- `09e60760` — `acquireStateLock` checked the deadline *after* calling `tryLockFile`, so a lock released during `time.Sleep` could be acquired after `lockTimeout` had elapsed, and a held lock overshot by up to one `lockRetryDelay`. It now checks before each attempt and caps the sleep to the remaining time. Measured: a 20 ms timeout took 101 ms before, ~20 ms after.
- `4a7660b0` — a test failing inside an `Update` window could report while still holding the lock, timing out every later test in the package. The window is now opened through a helper that registers its release with `t.Cleanup`.
- `944bd064` — cosmetic loop rewrite, no behavior change.

**Verification**

`go build ./...` and `go vet ./...` clean; `go test ./internal/state/... -race` ok; deadcode ratchet reports no new unreachable functions. Cross-compiled linux/amd64, darwin/arm64, windows/amd64, and aix/ppc64. Full `go test ./...` fails in 7 packages, all of which fail identically on a clean `main` at the same commit — a test-name diff shows no test failing here that does not also fail there.

`state.Update` still has no production caller, so `.deadcode-baseline.txt` keeps its single entry until #2300 lands.

decode2:
Closing this PR without merge because the implementation has been superseded by the current state-persistence architecture.

The original diagnosis was valuable, and the historical state race tests still pass. However, this branch is now conflicting and 590 commits behind, exceeds the review budget, closes an unapproved child issue, and introduces `Update` without any production caller while whole-document `Write` can still publish stale state. Current `main` now coordinates through the canonical `state.json.lock`, hardened file coordination, current schema preservation, and reconciled writes, so rebasing or cherry-picking this implementation would recreate competing ownership contracts.

We will preserve this PR as design and test provenance for #1809. Any replacement should start from a deterministic current-main cross-process reproduction and use the current canonical lock and persistence contracts. Thank you for surfacing the underlying lost-update problem and doing the early platform work.

PR #2300, opened 2026-08-03
refactor(state): migrate install-state call sites onto the advisory lock

pablontiv:
## 🔗 Linked Issue

Closes #1809

---

## 🏷️ PR Type

- [x] `type:refactor` — Code refactoring (no functional changes)

---

## 📝 Summary

Slice B of the two-PR chain for #1809. Stacked on the slice that adds `state.Update`.

Migrates the eight remaining read-modify-write call sites off the read-then-`Write` pattern and onto `state.Update`, so each mutation re-reads the persisted state inside the lock instead of writing back a snapshot captured earlier. This is the slice that actually closes the race described in the issue: `sync` read `persistedState` before building its pipeline and could write that same in-memory snapshot afterwards.

No behaviour changes beyond the serialization itself.

---

## 📂 Changes

| File / Area | What Changed |
|-------------|-------------|
| `internal/app/app.go` | State mutations moved onto `Update` |
| `internal/app/selfupdate.go` | Same |
| `internal/cli/review_mode.go` | Same |
| `internal/cli/sync.go` | Closes the read-then-write window described in the issue |
| `internal/components/uninstall/service.go` | Same |
| `internal/tui/model.go` | Same |
| `internal/update/cooldown.go` | Same |

---

## 🧪 Test Plan

```bash
go test ./internal/app/... ./internal/components/uninstall/... ./internal/tui/... ./internal/update/... ./internal/state/...
go run ./internal/gofmtcheck
go build ./...
go vet ./...
```

- [x] Unit tests pass for every touched package
- [x] Go format passes
- [x] `go build ./...` and `go vet ./...` clean
- [ ] Full `go test ./...` — deferring to CI. Locally `internal/cli` needs `-timeout 30m` and fails one unrelated `uv` test that reproduces identically on the base tree.
- [ ] E2E tests — deferring to CI
- [x] Manually reviewed each converted call site for mutation equivalence

Benchmark validation: N/A. This touches install-state persistence call sites only.

- [x] Commits follow Conventional Commits
- [x] Commits do not include `Co-Authored-By` trailers

---

## 💬 Notes for Reviewers

**Stacked PR — size will drop when the base lands.** GitHub measures this branch against `main`, which does not yet contain the slice-A commit, so the diff reads 528 changed lines. This slice itself is **137 insertions / 130 deletions = 267 lines**, verifiable with `git diff --stat pablontiv/sdd-1809-state-lock..pablontiv/sdd-1809-state-lock-callsites`. No `size:exception` is requested.

[...]

PR #2385, opened 2026-08-03
feat(state): add the platform advisory file-lock primitive

### LINKED ISSUE #2384: feat(state): add the platform advisory file-lock primitive
## Parent initiative

Part of #1809.

Slice 1 of 3. The parent issue #1809 remains open until the call-site migration lands,
which is the slice that actually fixes the reported bug.

---

### Pre-flight Checklist

### 🔍 Affected Area

Installation Pipeline

### 💡 Problem Statement

`internal/state` has no way to take a cross-process file lock. Serializing install-state writes (#1809) needs one, and the package currently has nothing to build on.

`internal/reviewtransaction` already carries a working pair (`store_lock_unix.go`, `store_lock_windows.go`), but it is unexported and package-private, so `internal/state` cannot reach it. The choice is to duplicate that pattern deliberately in `internal/state` or to extract a shared package. This issue proposes the former, and the Alternatives section explains why.

Splitting this out is not cosmetic. Bundled with the `Update` primitive it produced a 495-line change against the 400-line review budget. Separated, the lock primitive is 117 changed lines and the primitive that uses it is 389, both reviewable on their own.

### 📦 Proposed Solution

Add the platform advisory file-lock helpers to `internal/state`:

- `tryLockFile` takes an exclusive lock without blocking and reports whether another holder has it, distinguishing contention from real failure.
- `unlockFile` releases it.
- `flock` on Unix, `LockFileEx` on Windows, so both are exercised rather than one emulating the other.
- A contention test proving a second handle sees the lock held while the first holds it.

Acceptance criteria for this issue, separate from #1809:

- A second handle observes contention rather than acquiring the lock or erroring.
- Build constraints cover only GOOS values that genuinely provide the underlying syscall. `aix` and `solaris` lack `syscall.Flock`; `hurd` has no port. Note that `unix && !solaris` would silently drop `illumos`, which does have it, so the constraint needs an explicit allowlist rather than a subtraction.
- Cross-compilation succeeds for every allowlisted GOOS.

This issue does **not** cover `state.Update` or migrating any call site. Those are #2352 and #1809.

### 🔄 Alternatives Considered

**Extract a shared internal package from `internal/reviewtransaction`.** Rejected for now: it changes an area with its own review lifecycle and would grow the diff well past the budget this split exists to respect. Worth doing as its own refactor if a third consumer appears, and I would rather that decision be made deliberately than as a side effect of a bug fix.

**Reuse `internal/reviewtransaction`'s helpers directly.** Not possible: they are unexported.

**Keep the lock inline inside `Update`.** Rejected: it puts platform-specific syscall handling inside the state-mutation primitive, and the two have different reasons to change.

### 📎 Additional Context

Slice 1 of 3 under #1809. Implementation is ready and verified locally: 117 changed lines, `go build`, `go vet`, `gofmtcheck`, `go test -race`, a Windows build and cross-compilation for every allowlisted GOOS all pass.

Following the pattern already merged in this repository, where each slice closes its own approved child issue and the parent stays open. See #1225, which closes #1776 while remaining "Part of #1197", and the #2046 / #2061 pair.

The chain: this issue, then #2352 (`state.Update` on top of this), then #1809 (call-site migration, the slice that actually closes the bug).

One note for whoever reviews the follow-up: `state.Write` ends up taking this lock and can return a timeout error it never returned before. That is a behaviour change to an existing exported function and it is called out in #2352 rather than buried here.

I have no triage rights on this repository. Maintainer-owed: `status:approved` before the PR can pass its CI gates, plus whatever `type:*` label you prefer over the auto-applied `enhancement`.

pablontiv:
## 🔗 Linked Issue

Closes #2384

---

## 🏷️ PR Type

---

## 📝 Summary

Slice 1 of 3 under #1809. Adds the platform advisory file-lock helpers to `internal/state`, and nothing else.

`internal/reviewtransaction` already has a working equivalent, but it is unexported, so `internal/state` cannot reach it. The rationale for duplicating the pattern here rather than extracting a shared package is in #2384.

`tryLockFile` distinguishes contention from real failure, so a caller can tell "someone else holds it" from "the syscall failed". `unlockFile` releases it. `flock` on Unix, `LockFileEx` on Windows.

The build constraint is an explicit allowlist rather than `unix` minus exclusions. `aix` and `solaris` do not provide `syscall.Flock` and `hurd` has no port, but `unix && !solaris` would silently drop `illumos`, which does provide it. The allowlist was verified by cross-compiling every listed GOOS.

---

## 📂 Changes

|-------------|-------------|
| `internal/state/lock_unix.go` | `flock`-based acquisition and release, allowlisted build constraint |
| `internal/state/lock_windows.go` | `LockFileEx` equivalent |
| `internal/state/state_lock_test.go` | Contention test: a second handle observes the lock held |
| `.deadcode-baseline.txt` | Temporary entry, see below |

---

## 🧪 Test Plan

```bash
go test ./internal/state/... -race -count=1
go run ./internal/gofmtcheck
go build ./... && go vet ./...
GOOS=windows go build ./internal/state/
./scripts/deadcode-ratchet.sh
```

- [x] Unit tests pass for `internal/state`, including `-race`
- [x] Go format passes
- [x] `go build ./...` and `go vet ./...` clean
- [x] Windows build and cross-compilation for every allowlisted GOOS pass
- [x] Deadcode ratchet passes
- [ ] Full `go test ./...` — deferring to CI. Locally `internal/cli` needs `-timeout 30m` and fails one unrelated `uv` test that reproduces identically on the base tree.
- [ ] E2E tests — deferring to CI

Benchmark validation: N/A. This adds file-locking helpers to install-state persistence, not a review-lifecycle, gate, recovery, delivery or benchmark surface.

- [x] PR stays within 400 changed lines (117)

---

## 💬 Notes for Reviewers

**The deadcode baseline entry is deliberate and temporary.** These helpers have no production caller in this slice by construction; the caller is `state.Update` in the next one. `scripts/deadcode-ratchet.sh` therefore fails without a baseline entry. This is not the escape hatch's intended case, which is why the commit message says so plainly, and the follow-up slice removes both entries and replaces them with `state.Update` becoming reachable. If you would rather see the primitive and its caller land together, that is a ~506-line PR and I will do it that way instead.

Chain: this PR, then #2352 (`state.Update`), then #1809 (call-site migration, the slice that closes the bug). Linkage follows the pattern in #1225.

I have no triage rights on this repository. Maintainer-owed: `status:approved` on #2384, plus a `type:*` label here.

## Summary by CodeRabbit

* **New Features**
  * Added cross-platform file locking support for Unix-like systems and Windows.
  * Lock attempts now return immediately when a file is already locked, allowing applications to handle contention without blocking.
  * Locks can be released reliably after use.

* **Tests**
  * Added coverage verifying lock acquisition, contention handling, and subsequent lock release.

### DIFF
--- a/.deadcode-baseline.txt
+++ b/.deadcode-baseline.txt
@@ -210,6 +210,8 @@ internal/sddstatus/status.go	reportLineHasPassSignal
 internal/sddstatus/status.go	reportTextIsClearlyPassing
 internal/sddstatus/status.go	reportValueIsBenign
 internal/sddstatus/status.go	stripMarkdownSignal
+internal/state/lock_unix.go	tryLockFile
+internal/state/lock_unix.go	unlockFile
 internal/storage/bytes.go	FormatBytes
 internal/system/install_deps.go	InstallCommandsForDep
 internal/system/install_deps.go	installCommandsBrew
--- a/internal/state/lock_test.go
+++ b/internal/state/lock_test.go
@@ -1,0 +1,51 @@
+package state
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestTryLockFileReportsBusyForSecondHandle covers the platform primitive
+// directly: a second open handle on an already-locked file reports busy rather
+// than returning an error or blocking.
+func TestTryLockFileReportsBusyForSecondHandle(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "lock")
+
+	first, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, 0o644)
+	if err != nil {
+		t.Fatalf("open first handle: %v", err)
+	}
+	defer first.Close()
+	locked, err := tryLockFile(first)
+	if err != nil {
+		t.Fatalf("lock first handle: %v", err)
+	}
+	if !locked {
+		t.Fatal("first handle could not take a free lock")
+	}
+
+	second, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, 0o644)
+	if err != nil {
+		t.Fatalf("open second handle: %v", err)
+	}
+	defer second.Close()
+	locked, err = tryLockFile(second)
+	if err != nil {
+		t.Fatalf("lock second handle: %v", err)
+	}
+	if locked {
+		t.Error("second handle took a held lock; the lock does not exclude")
+	}
+
+	if err := unlockFile(first); err != nil {
+		t.Fatalf("unlock first handle: %v", err)
+	}
+	locked, err = tryLockFile(second)
+	if err != nil {
+		t.Fatalf("relock second handle: %v", err)
+	}
+	if !locked {
+		t.Error("second handle could not take the lock after release")
+	}
+}
--- a/internal/state/lock_unix.go
+++ b/internal/state/lock_unix.go
@@ -1,0 +1,33 @@
+// The `unix` pseudo-tag also matches aix and solaris, neither of which defines
+// syscall.Flock, and hurd, which has no port in this toolchain. Listing the
+// verified targets keeps the constraint honest and fails closed on new ports.
+// android satisfies the linux tag and ios satisfies darwin, so both are covered.
+//go:build darwin || dragonfly || freebsd || illumos || linux || netbsd || openbsd
+
+package state
+
+import (
+	"errors"
+	"os"
+	"syscall"
+)
+
+// tryLockFile attempts to take an exclusive advisory lock without blocking.
+// It returns (true, nil) when the lock was taken, (false, nil) when another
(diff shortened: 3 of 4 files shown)


pablontiv:
Closing as redundant with #2299.

This PR's `lock_unix.go` and `lock_windows.go` are byte-identical to the ones already on the #2299 branch, and `lock_test.go` differs only by the build constraint added there. Both PRs target `main`, which makes the diffs mutually dependent rather than independent slices: merging either one empties or shrinks the other.

Per the chain plan confirmed in #2299, that PR is the foundation slice and carries the primitive; #2300 rebases onto `main` after it merges. The unsupported-platform handling this PR's review asked for is resolved in #2299 by `003a2d5c`, including the `lock_test.go` build constraint.

Issue #2423, opened 2026-08-04
refactor(review): delete orphaned review authority after the SDD clean break


Alan-TheGentleman:
## Pre-flight Checklist

- [x] The full #3417 and #3564 cleanup cluster was audited against current main.
- [x] Existing issues and pull requests were searched before reusing this approved tracker.
- [x] This issue is the non-SDD follow-up. #3564 owns the SDD clean break.

## Outcome

Delete every review/RDD production surface outside SDD that becomes orphaned after #3564 removes its final consumer. Retain only code with a named live path in the current atomic compact-v2 lifecycle.

This is deletion, not migration. Historical formats, versions, bindings, receipts, recovery shells, switches, and compatibility readers carry no lifecycle authority and are not recognized by the replacement implementation.

## Dependency

This work begins after #3564 lands its complete SDD cleanup. The post-#3564 source tree is the only valid reachability baseline. Pre-#3564 references do not justify retaining a provider whose final consumer is scheduled for deletion.

### D1. No backward compatibility

Old review formats and authority representations are unsupported. A replaced version fails closed with one actionable instruction to start a fresh atomic review. There is no dual reader, compatibility translator, migration, record rewrite, or legacy fallback.

### D2. Preserve only the live atomic lifecycle

The retained review lifecycle is:

```text
candidate consent → freeze → admitted lenses → at most one bounded correction → targeted validation → approval and burn
```

Fresh STATUS/FINALIZE routing, exact consent, immutable candidate binding, reviewer admission, bounded correction, validator evidence, and terminal burn remain only when they are reachable from this lifecycle.

Approval leaves no terminal receipt, lineage, witness, tombstone, mirror, gate authority, or durable review history. Git remains the durable project history.

### D3. Delete by causal closure

When a consumer is deleted, delete its newly unreachable providers, direct tests, fixtures, schemas, help text, refusal continuations, ratchet entries, manifests, docs, and compile references in the same causal work unit. Do not baseline newly dead code and do not preserve a helper solely because a test calls it.

### D4. Evidence required for retention

Every retained review symbol must have all three:

1. a non-test production caller;
2. a named current compact-v2 scenario;
3. a focused test proving that scenario.

Anything missing one of these is deletion scope unless a separate approved issue owns a concrete current use.

## Scope

After #3564 merges, inventory and delete orphaned review/RDD code outside `internal/sddstatus`, including as applicable:

[...]

PR #2444, opened 2026-08-04
test(bench): run the damaged-store axis in CI and cover the new lifecycle flows

### LINKED ISSUE #2443: test(bench): the damaged-store axis runs in no CI path and new lifecycle flows lack journeys
Two coverage gaps that must close before the release candidate, both found during the wave 6 and wave 7 verifications.

## The damaged-store axis never runs in CI

`ci.yml` runs the portable core plus `--axis source-coupled --only j57`, so ds01 through ds12 execute on no CI path at all. Those journeys are the exit evidence for the disposition and repair contract across two shipped waves: closure ordering, resume convergence, authorization on resume, over-collection. A regression in any of them passes CI silently today, and the risk grew as waves 2 and 6 added more proofs resting on them.

Six of the crash-position journeys additionally require a `-tags bench_fixture` build. A plain binary reports them `unsupported` rather than `failed`, which must never be read as a pass. Neither the axis Properties declaration nor `bench/README.md` disclosed the tag requirement, so contributors got unexplained results.

## New lifecycle flows have no journey coverage

Package tests, e2e and goldens cover the flows waves 3 through 6 introduced, but the bench corpus is the layer that exercises a real user end to end, and it lagged. This is not theoretical: a journey caught a genuine production defect in wave 6 that unit tests structurally could not see, because the defect lived in a public entrypoint wrapper the unit tests bypassed.

Flows to cover: post-verify offer accepted through to a receipt, kill switch off end to end asserting structural absence rather than a no-op, targeted re-verify after a correction, and a per-gate deny to allow round trip. Flows already covered elsewhere should be confirmed and left alone rather than duplicated.

Alan-TheGentleman:
## 🔗 Linked Issue

Closes #2443

## 🏷️ PR Type

- [x] `type:feature`

## 📝 Summary

**CI gate.** All 17 damaged-store journeys now run on every PR, in both binary shapes: the plain build asserts the 11 non-crash journeys complete and the 6 crash-position ones report unsupported, then the `-tags bench_fixture` build asserts all 17 complete. A length guard fails if a typo ever shrinks the selection. Measured cost is about 28 seconds, against roughly 25 for the entire existing portable core, so no schedule-only split was needed. The build-tag requirement is now disclosed in both the axis Properties and the README, which previously documented only ds01 through ds05.

**Journey gate.** Four new journeys cover flows that had none: post-verify offer accepted through to a discovered receipt, kill switch off across one continuous flow asserting structural absence at every stage in the same run rather than one isolated stage per journey, targeted re-verify after a real correction including running the printed continuation and proving archive returns to ready, and one gate held through a full deny to allow round trip. Two further candidate flows were found already covered and deliberately not duplicated.

The portable core goes from 59 to 63 journeys.

## 🧪 Test Plan

- [x] Every new journey run against a freshly built binary and shown completing
- [x] Every new journey shown genuinely RED against a deliberately broken product, then restored: the offer allow branch, the disabled-mode routing guard, the re-verify classifier, and the gate allow assignment
- [x] The CI gate itself proven RED by breaking `RepairAuthorityDisposition`
- [x] Full `go test ./... -count=1` root and bench, gofmt, vet, deadcode ratchet green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,79 @@ jobs:
             completed_once("j57-sdd-authority-drift-during-discovery-fails-closed")
           ' "$RUNNER_TEMP/bench-source-coupled.json"
 
+      # The damaged-store axis (ds01-ds12, 17 journeys — ds11 alone expands to
+      # six crash-position journeys) is exit evidence for the disposition and
+      # repair contract shipped across two waves: closure ordering, resume
+      # convergence, authorization-on-resume, over-collection. It runs on
+      # every PR, not only push/schedule: on this machine the whole axis costs
+      # ~12s untagged plus ~16s tagged (~28s total), against ~25s for the
+      # entire portable core — small next to the job's existing docker/E2E
+      # cost, so nothing is dropped to a schedule-only lane.
+      #
+      # Six of the seventeen (the ds11 crash-position journeys) require a
+      # `-tags bench_fixture` binary to genuinely interrupt the product
+      # mid-transaction; run first against the plain binary already built
+      # above, where those six MUST report `unsupported`, never `completed`
+      # or a silent skip — an unsupported result must never be mistaken for a
+      # pass. Then run the identical selection against the tagged binary,
+      # where all seventeen MUST complete. The explicit `length == 17` checks
+      # are the guard against a typo in this id list quietly shrinking the
+      # axis's own CI coverage back down.
+      - name: Run damaged-store axis evidence
+        run: |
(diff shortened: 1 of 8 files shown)


Alan-TheGentleman:
Thanks for building real journey coverage and CI evidence. The `ReviewGate`, `ReVerify`, receipt-discovery, and gate-round-trip flows in `bench/journeys_gate2.go` were replaced by the atomic approval and burn lifecycle in `b43e0928a2bf`, while #3564 and #2423 retire the remaining authority. The path forward is to port only a demonstrably missing current atomic-lifecycle journey under #2423 after cleanup, rather than preserving these retired journeys.

PR #2448, opened 2026-08-04
feat(review): give OpenCode genuine immutable receipt-review through its shell-less channel

### LINKED ISSUE #2207: fix(review): gate unsupported immutable review transport before lifecycle mutation
### Pre-flight Checklist

- [x] I searched open and closed issues, including transport, capability, lifecycle, Codex, OpenCode, and immutable-evidence terms. This is not an equivalent issue.
- [x] This issue is maintainer-approved. Implementation PRs must link it and preserve the stated boundaries.

### Problem

Negotiated receipt review currently models the lifecycle independently from whether the active runtime can actually transport immutable review evidence and capture a bound result. A runtime without a native transport can therefore reach `review.start` and `collect` for a selected lens even though it cannot launch that lens with a trusted binding, deliver immutable `artifact_subject` / `base_tree` / `candidate_tree` / ordered manifest evidence, or return `AS/B/C/M` through native capture.

The reviewer correctly returns incomplete rather than inspecting mutable state, but negotiated STATUS can reoffer the same bound slot. That creates an impossible loop: delivery remains blocked while the runtime has no permitted path to produce an admissible receipt. The failure should be a typed unsupported-capability result before authority, budget, or collection state mutates.

### User Impact

Users on unsupported or incompletely integrated runtimes can start a medium or high-risk review that can never complete. They see repeated collection work rather than one actionable explanation that immutable receipt review is unavailable in that runtime. This also makes normal SDD completion and receipt-gated delivery disagree.

### Reproduction

1. Use a registered runtime that receives shared SDD/reviewer assets but has neither the managed OpenCode transport nor Claude's prompt-carried transport.
2. Start negotiated review with `gentle-ai.review-integration/v2` for a medium or high-risk candidate.
3. Complete the returned consent transition.
4. Observe STATUS return `collect` and one or more bound reviewer-result inputs.
5. Launch the selected lens. It cannot receive a trusted immutable binding/context or invoke an allowed immutable-inspection transport, so it returns incomplete.
6. Query STATUS again. The same bound collection slot can be reoffered although no supported runtime path can make it complete.

### Current Behavior

- `internal/components/sdd/boundedreview.go` renders a generic reviewer path that explicitly says other runtimes without an enforceable transport return incomplete.
- `internal/components/sdd/inject.go` installs the managed review-result plugin only for OpenCode and Kilo Code.
- Claude has a separate prompt-carried immutable context and pinned runtime proof.
- `internal/agents/capabilitymanifest/manifest.go` provides static adapter capabilities but no immutable receipt-review transport declaration or runtime identity that negotiated lifecycle can consume.
- The lifecycle can still proceed to review start/collection before this absence is represented as a typed result.

### Expected Behavior

Negotiated STATUS and START must know whether the active runtime has a complete immutable receipt-review transport. If not, they must return a typed unsupported-capability outcome before any review authority, selected-lens budget, or collection mutation. A completed immutable transport remains eligible only when it can bind the exact artifact subject and trees, deliver the ordered manifest through an enforceable least-privilege boundary, and capture the result under the same binding.

### Scope

- Add one immutable receipt-review transport capability to the canonical adapter/capability manifest.
- Carry the active runtime identity into negotiated STATUS and START, rather than inferring support from shared prompt distribution.
- Preflight the active runtime before review authority creation, lens selection/budget consumption, or collection output.
- Return a stable typed unsupported-capability result with actionable runtime/capability identity when the transport is absent or incomplete.
- Preserve the existing OpenCode/Kilo managed provider path and Claude prompt-carried path.
- Add behavior tests across every registered adapter, including unsupported and incomplete cases.

### Non-Goals

- Do not treat generic SDD, skills, MCP, file subagents, or static prompt presence as proof of review transport.
- Do not weaken immutable inspection, allow live worktree substitution, broaden reviewer permissions, or admit incomplete results.
- Do not implement Codex transport in this issue. That capability is tracked separately.
- Do not change normal SDD behavior when receipt review is not selected.

### Evidence

Audited on `main` at `eaa7ae2186c8a2d64079c6802b7309ceaaf47aa8`:

- The review ledger contract requires routing only from negotiated transitions and marks access/provider failure or incomplete inspection as non-completion.
- OpenCode receives provider-owned context and bounded commands through `internal/assets/opencode/plugins/review-result-artifacts.ts`.
- Claude receives prompt-carried immutable evidence and an opt-in pinned runtime test in `internal/components/sdd/claude_review_network_none_e2e_test.go`.
- No corresponding runtime transport exists for Codex, and the manifest has no immutable-review-transport capability.

Related, not duplicates:

- #2003 documents the Claude defect and its runtime-specific remediation.
- #1967 and #1977 are closed OpenCode transport incidents resolved by the v2 provider path.
- #1033 is a partial Antigravity dynamic-subagent parity discussion.
- #2032 is Codex documentation-only work.
- #1867 covers unreadable/truncated evidence admission and recovery after review begins, not preflight runtime capability gating.
- #1873 defines broader native capability certification and explicitly excludes Codex from its first tuple; it does not make review lifecycle routing capability-aware.

### Acceptance Criteria

- [ ] The canonical adapter/capability manifest declares immutable receipt-review transport as an explicit immutable capability, distinct from skills, SDD, generic subagents, and MCP.
- [ ] Negotiated STATUS and START receive the active runtime identity and resolve its declared transport before lifecycle mutation.
- [ ] Unsupported or incomplete runtimes return a typed unsupported-capability outcome before authority creation, lens/budget consumption, or `collect` output.
- [ ] The typed outcome names the missing immutable receipt-review transport without exposing private paths, candidate content, or environment values.
- [ ] Unsupported paths cannot reoffer the same collection slot or create an impossible review loop.
- [ ] OpenCode and Kilo Code retain the managed provider transport, exact binding preflight, least-privilege inspection boundary, and native result capture.
- [ ] Claude retains the prompt-carried immutable context and its existing bounded no-shell transport behavior.
- [ ] Normal SDD remains unchanged when receipt review is not selected.
- [ ] Behavior coverage enumerates every registered adapter and proves supported, unsupported, and incomplete transport outcomes.
- [ ] A runtime claiming support has a native integration/E2E proof before it is eligible for negotiated immutable receipt review.

### Rollout

1. Land the manifest and lifecycle preflight with all unsupported/incomplete runtimes fail-closed.
2. Preserve the proven OpenCode/Kilo and Claude paths behind their explicit declarations.
3. Add a new runtime only together with its exact immutable transport, capture boundary, negative tests, and pinned native runtime evidence.
4. Surface the typed unavailable state in diagnostics and release notes so users do not enter a non-converging review.

`skill_resolution: issue-creation + maintainer-voice`

### LINKED ISSUE #2413: fix(review): unsupported-capability reviewer refusal names no exit, so agents infer 'install another runtime'
## Observed

On OpenCode, a bounded review reaches the reviewer stage and the lens agents refuse with `unsupported-capability`. The refusal text (`internal/components/sdd/boundedreview.go:167`) reads:

> Immutable OpenCode candidate inspection is unsupported-capability: OpenCode cannot securely bind provider-injected dynamic values to this child session's Bash permission. Do not run Bash, native review commands, another provider, or a live worktree. Return incomplete inspection with empty paths/findings and evidence that secure immutable inspection is unavailable, then stop.

The consuming agent, left with no named continuation, concluded on its own that the remedy was to install the Claude Code binary and offered that to its human. The SDD flow stayed parked in `resolve-blockers`.

## Why the refusal itself is correct

This is fail-closed by design. A runtime that cannot enforce a per-command shell boundary cannot be trusted to keep reviewer inspection confined to the frozen trees, so returning incomplete is the right outcome. That part should not change.

## The defect

The refusal names no executable exit. Every stop in this system is supposed to name one, and this one leaves the consumer with a dead end, so an agent will invent a remedy. Here it invented the most drastic one available: install a second agent runtime. A user who only wanted to commit is told to change their toolchain.

## Expected

The refusal should name both real continuations, in this order:

1. `gentle-ai review mode disable` to proceed under ordinary repository policy with delivery reported as `disabled/unmanaged`. No install, no receipt, honest record.
2. Running the review on a runtime that supports secure immutable inspection, if a receipt is actually wanted. Naming the supported runtimes makes installing an informed choice rather than an inferred obligation.

## Scope

Message and exit-naming only. No machinery, no change to the admission decision. Related: #2209 (supported-runtime matrix), and the blocking-budget policy in `docs/architecture/rdd-freeze-expansion-policy.md` which states that every stop must name its reason and its exit.

### LINKED ISSUE #2417: feat(review): restore OpenCode immutable inspection through the shell-less native channel
## Decision

OpenCode is the most-used runtime in this community and must ship with working immutable receipt-review. #2209 already states this ("OpenCode | Native/bundled supported. Retain the proven transport"), and current main contradicts it.

## What is actually blocking today

`internal/components/sdd/inject.go:944` injects an unsupported-capability prompt for OpenCode with `bash: deny`, so every lens returns incomplete and no receipt can ever be produced. The stated cause is that provider-injected dynamic values cannot bind to the child session's Bash permission.

That cause is real but the conclusion does not follow. OpenCode loads `opencode.json` and its plugins at process startup, while the provider's values only exist after `review.start` inside an already-running process, so exact-value allow patterns can never be written in time. Per-session permission generation is structurally impossible.

## Why no upstream capability is needed

The plugin already owns a shell-less channel. `runNative` in `internal/assets/opencode/plugins/review-result-artifacts.ts` is `spawn("gentle-ai", args)` with exact argv and no `shell: true`, so it is provider host code and never subject to the agent's Bash permission. It is already used three times.

`injectReviewerContext` exists with **zero call sites**: commit `86ea91b9` replaced the before-hook body with an unconditional throw. Restoring and extending it with `inspect-candidate` calls removes the problem entirely, along with the whole Windows Bash-permission defect class.

## Target shape

Ship the lens with **no shell and no read tool**. The provider materializes discovery and selective diffs and injects them; the injected prompt becomes provably the lens's only byte source. That is strictly stronger than what the other runtimes get today, where the guarantee is prompt-only.

## Slices

1. Plugin evidence materialization through the restored hook.
2. Go generator: `bash:false`, `read:false`, delete `openCodeReviewerPermission()`.
3. Capability flip and contract-truth rewrite.
4. Proofs: closed-world mutation suite (read:true, bash allow, fall back to caller prose, truncate instead of fail closed) plus a poisoned-worktree organic e2e with the Windows-only skip removed.

## Must measure before shipping

Whether OpenCode truncates a very large `task` `args.prompt`. The byte budget must sit below that limit and fail closed, because silent truncation is the one mode that could fabricate a false clean result.

## Related

#2207 (reopened), #2413, #2278 (still open, the actual Windows migration defect that was misgeneralized into a permanent capability claim).

Alan-TheGentleman:
## 🔗 Linked Issue

Closes #2417
Closes #2413
Closes #2207

## 🏷️ PR Type

- [x] `type:feature`

## 📝 Summary

OpenCode gets genuine immutable receipt-review. It is the most-used runtime in this community and until now its lenses were told the capability was unsupported, so a medium or high risk candidate could freeze consent, tier, lenses and budget and then never reach a receipt.

No upstream capability was needed. The plugin already owned a shell-less channel: `runNative` spawns `gentle-ai` with exact argv and no shell, so it is provider host code and never subject to the agent's Bash permission. `injectReviewerContext` existed with zero call sites, disabled two days earlier. Restoring and extending it removes the problem, and the whole Windows Bash-permission defect class with it.

**The shape.** The lens holds no shell, no read tool, no MCP tool, no plugin tool, no skill loader, and cannot spawn a nested agent, verified against OpenCode 1.18.10's own resolution rather than inferred from config. The provider materializes discovery and per-path patches and injects them; caller-authored prose is discarded wholesale.

**What adversarial verification found and what it cost.** The tool gate alone was not enough. OpenCode concatenates live project instruction files, `instructions` globs, the skill catalog and an environment block into every session's system prompt regardless of tools, so post-freeze bytes reached the lens above the injected evidence. Two isolation variables close most of it and the plugin now refuses to launch without them. A remote `instructions` URL is fetched regardless of those variables and from any layer, so the plugin separately reads the effective configuration through its own OpenCode client and refuses when one is present, naming the entry. An unreadable config also refuses, because an unverifiable configuration cannot be ruled safe.

Also closed: the budget refusal now names a real continuation and sits at 4 MiB, matching the existing frozen-candidate cap rather than an arbitrary figure; an empty patch for a content-changing path refuses instead of injecting empty evidence; and the plugin test harness now fails rather than skipping when the plugin breaks at parse time.

## 🧪 Test Plan

- [x] Truncation measured empirically against real OpenCode 1.18.10 with a fixture server: no ceiling found through 16 MiB, so the budget refuses rather than truncates
- [x] Poisoned-worktree e2e against a real `opencode` process proving the lens cannot see live state, with the Windows-only skip removed
- [x] A second real e2e proving the remote-instructions refusal is a separate gate, run with both isolation variables correctly set
- [x] Every mutant confirmed RED then reverted byte-identically: read tool enabled, a bash allow, falling back to caller prose, truncating instead of refusing, and disabling each new gate
- [x] Full `go test ./... -count=1` root and bench, gofmt, vet, both ratchets, goldens without `-update`

## Note on the shipped claim

The contract text states exactly what is enforced and no more. The one remaining unstated residual is the environment block, which carries working directory, platform and date: structural metadata, not attacker-controllable.


Alan-TheGentleman:
Closing as superseded: #2479 (branch pr-2448-unblock) rebuilt this on current main and merged, closing #2417 and #2413. #2207 remains open pending verification that the merged transport-capability path fully covers it against current main; if it does, it closes with that evidence, and if not, the gap is the next slice. All future OpenCode transport work continues from what #2479 landed.

Issue #2418, opened 2026-08-04
feat(review): give Codex the reviewer lens assets it needs to honor its supported status


Alan-TheGentleman:
## Gap

#2209 lists Codex among the three runtimes supported for immutable receipt-review. `internal/assets/codex/` contains exactly one file, `sdd-orchestrator.md`. There are no reviewer lens assets, so there is nothing to launch when a review selects a lens. `internal/assets/skills/_shared/review-ledger-contract.md:17` already acknowledges this.

Compare `internal/assets/claude/agents/`, which ships five lens agents (risk, readability, reliability, resilience, refuter).

## Why it matters

A declared capability that the runtime cannot honor is the exact defect pattern being fixed across #2207, #2413 and #2417. Declaring Codex supported while it has no lens path either strands its users after state has been frozen, or forces an early refusal that contradicts the approved support decision.

## Work

Build the Codex lens assets so the declaration becomes true. Follow the shape #2417 establishes for OpenCode rather than Claude Code's current one: the provider materializes the inspection evidence and the lens receives no live-state access, so the guarantee is structural rather than prompt-only.

Until the assets exist, the capability matrix must report Codex honestly rather than advertise a path that is not there.

## Related
#2209 (the support decision), #2417 (OpenCode's shell-less shape, the reference implementation), #2207.

Alan-TheGentleman:
## Probe verdict: hook input rewriting works, but not on the models we actually ship

We ran the probe on Codex 0.144.0, 0.145.0 and 0.146.0. A `PreToolUse` hook does genuinely rewrite tool input through `hookSpecificOutput.updatedInput`, including the subagent spawn call. The parent authored `message: "GENTLE_ORIGINAL_MESSAGE"`, the hook replaced it, and the child echoed the hook's string. Identical on all three versions.

The qualification is the actual finding. There are two spawn tools, and **the model picks which one**, not the CLI version and not any client flag:

| Model | spawn tool | `message` field | rewritable |
|---|---|---|---|
| `gpt-5.6-sol`, `gpt-5.6-terra` | `collaborationspawn_agent` | encrypted blob | no |
| `gpt-5.6-luna`, `gpt-5.5`, `gpt-5.4` | `spawn_agent` | plaintext | yes |

[...]

PR #2479, opened 2026-08-04
feat(review): give OpenCode genuine immutable receipt-review

### LINKED ISSUE #2207: fix(review): gate unsupported immutable review transport before lifecycle mutation

### LINKED ISSUE #2413: fix(review): unsupported-capability reviewer refusal names no exit, so agents infer 'install another runtime'

### LINKED ISSUE #2417: feat(review): restore OpenCode immutable inspection through the shell-less native channel

Alan-TheGentleman:
OpenCode gets genuine immutable receipt-review through its shell-less channel. Until this lands, every OpenCode user is locked out of negotiated review on the published release candidate, and OpenCode is the most-used runtime.

This is the contributor's work in PR #2448, rebased onto main with all eight commits preserved, plus the changes the migration made necessary.

## The redundancy, and the deletion it enabled

`gentle-ai review lens-context`, merged as #2453 after this PR was written, had already absorbed the PR's central mechanism into Go: the same 4 MiB budget, the same refuse-never-truncate rule, the same empty-patch guard, the same section headers. It also already shipped `ReviewerContextLevelRuntimeInterception`, described as "a runtime adapter replaced whatever the caller produced", which was provisioned for exactly this plugin.

So rather than merge two mechanisms doing one job, the duplicate is deleted. Gone from the plugin: `materializeReviewEvidence`, `inspectionArgs`, `REVIEW_CONTEXT_BYTE_BUDGET`, and the now-unreachable `preflightCapture` / `validManifest` / `validManifestEntry` chain with its three interfaces. `injectReviewerContext` now makes one call and injects the bytes verbatim. **The plugin shrank by roughly 200 lines net.**

Two checks stay plugin-side because the provider cannot make them. The block must end with `GENTLE_AI_REVIEW_CONTEXT_END`, and the provider's binding must match the lineage, target, revision, order and handle the task claimed. That second one matters: a caller relaying a handle for one lineage while claiming another now fails loudly instead of capturing into that other lineage after the reviewer is already spent.

Kept unchanged because they are genuinely novel with no native equivalent: the `OPENCODE_DISABLE_PROJECT_CONFIG` and `OPENCODE_DISABLE_EXTERNAL_SKILLS` gate, and the remote-`instructions` refusal via `client.config.get()`. Both close OpenCode's system-prompt assembly, which is invisible to any native command.

## Verification

Fresh `git init`, binaries built from the exact commits.

Before: `immutable_review_transport_unsupported`, `next_action: stop`, `cause: the active runtime lacks immutable receipt-review transport`.

After: `operation_failed`, `next_action: retry`. That residual is the empty repository, not the runtime, confirmed because `--agent claude-code` returns the identical code on the same repository. On a repository with one commit the branch returns `status/v4`, `action: start`, `reason_code: fresh_target_ready`, and a command carrying `--agent=opencode`.

Codex still refuses, as do Pi and Kilocode. The #2418 decision is untouched.

A real negotiated review was driven through to collect: START, consent envelope, `--consent granted`, medium risk with the reliability lens, then `review lens-context --delivery runtime_interception` emitting 6372 bytes in exactly the order the plugin consumes, with a seven-field binding matching the transition tokens exactly.

Root `go test ./... -count=1` clean, bench module clean, gofmt and vet clean, deadcode ratchet reports no new unreachable functions. Docker E2E on ubuntu with `RUN_FULL_E2E=1`: 488 passed, 0 failed, 1 skipped.

## A real cost regression the guard caught

The OpenCode lens prompt grew, because it now describes the injected block instead of carrying one refusal sentence. Standard tier went from 19,276 to 20,157 characters, leaving 9.2 percent headroom under the 22,200 ceiling against a 15 percent requirement. The ceiling was raised to 23,300, following the precedent already recorded in that file. The full four-lens tier already cleared it.

## Known, and it needs a follow-up

`internal/assets/skills/_shared/review-ledger-contract.md` and `docs/review-integration.md` are held at main deliberately. Both are under concurrent edit on two other branches, and carrying them here would have turned one golden into thirteen.

**Both still say OpenCode is eligible but transport-disabled, which this change makes false.** That is prose rather than behavior, but whichever branch lands last owes the correction.

Also pre-existing and now reachable: `reviewNextTransitionRefreshCommandV21` hardcodes `--agent claude-code` at `internal/cli/review_capabilities.go:54`, so every lens-context refusal tells an OpenCode operator to refresh as claude-code. Sibling branches are aimed at that.

`e2e/organicruntime/opencode_reviewer_unsupported_e2e_test.go` is now misnamed, since it contains the support proof. Not renamed here because another branch edits the same file and a rename-versus-edit conflict would be worse.

Closes #2417
Closes #2413
Closes #2207

Part of #2471

## Summary by CodeRabbit

* **New Features**
  * OpenCode now supports immutable review workflows using provider-supplied review context.
  * Reviews include complete, ordered evidence tied to the correct candidate.
* **Bug Fixes**
  * Reviewers can no longer access live worktree changes, secrets, or repository instruction files during isolated reviews.
  * Reviews now refuse to proceed when isolation or configuration checks fail.
* **Tests**
  * Expanded coverage for context validation, refusal handling, recovery, and reviewer isolation.


rbedani:
Consumer follow-up (same occurrence as my report in #2132 — lineage `review-5d525633161476e5`, OpenCode, Linux aarch64). Monitoring this PR for the fix; the merge is currently blocked on two actionable items:

1. **Merge conflict**: `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING` — `pr-2448-unblock` needs a rebase against `main` after the #2448 merge.

2. **"Unit Tests" fails in the `Run benchmark evidence` step** (run 30943064780, job 92108734805), not in the unit tests themselves:
   - Claim `j43-recovery-guard-rails-as-an-operator-meets-them` — "invalidate the healthy approved authority" — fails: *"the product named a runnable continuation, and mechanical evidence outranks the declaration"*.
   - The mechanical evidence is the derived-invalidation error emitted under that journey (`invalidated is now a derived verdict, never a write; see it instead: gentle-ai review validate …`), so the journey's by-design declaration for j43 appears stale against the current invalidate semantics.

Happy to re-verify the post-correction continuation flow once this lands and a release is cut.

PR #2591, opened 2026-08-05
fix(filemerge): support Windows junction parent directories

### LINKED ISSUE #2164: sync aborts on Windows directory junction when writing the gga PowerShell shim
### Pre-flight Checklist

### 📝 Bug Description

`gentle-ai sync` aborts unconditionally when the parent directory of the gga PowerShell shim is a Windows directory junction. The shim writer classifies the junction as "not a directory" and the whole sync pipeline fails, so no component is synced at all.

I do not think this is the same defect as #306 / PR #1077. Those cover symlinked managed *files* in `internal/components/filemerge`, resolving a link before an atomic write. This one is a *parent directory* classification in the gga shim writer, and PR #1077 (still open) does not appear to touch that path. If you judge them the same family, closing this as a duplicate is reasonable.

### 🔄 Steps to Reproduce

1. On Windows, make `%USERPROFILE%\bin` a directory junction pointing at a directory on another volume:

   mklink /J C:\Users\<user>\bin D:\<dir>\bin

   In my case the junction was created on 2026-05-23 and contains a `gga` binary.

2. Run `gentle-ai sync`.

3. Sync aborts before applying any step.

The junction is a real, resolvable directory. Two independent checks agree:

- Python: os.path.isdir -> True, os.path.islink -> False
- PowerShell: Get-Item -Force -> Mode `d----l`, LinkType `Junction`, Target `D:\<dir>\bin`

### ✅ Expected Behavior

sync resolves the junction and writes `gga.ps1` into the target directory, exactly as it would for a plain directory. A reparse point that resolves to an existing directory should be treated as a directory.

### ❌ Actual Behavior

Sync aborts immediately with a parent-path classification error and applies nothing. Because `sync` has no component filter, there is no supported way to skip the gga component and sync the rest.

The impact goes beyond the manual command: the Windows auto-update scheduled task ends with `gentle-ai sync`, so every daily run now dies at that step. From my auto-update log:

- 74 runs started
- 70 reached the sync step
- 65 completed
- last completed run: 2026-07-27

The 5 runs from 2026-07-28 onward log "Running gentle-ai sync..." and never log a completion line. The scheduled task reports result 267014 (task terminated) and the per-run config backup directories it creates are empty.

This looks like a regression rather than a bad local setup. The junction predates the breakage by about two months and sync completed 65 times with it already in place. The failure starts somewhere between 2026-07-20 and 2026-07-27.

### Gentle AI Version

2.2.4

### Operating System

Windows

### AI Agent / Client

Claude Code

### 📋 Affected Area

Installation Pipeline

### 💡 Logs / Error Output

```shell
$ gentle-ai sync
Error: execute sync pipeline: ensure gga powershell shim: write gga runtime file "C:\Users\<user>\bin\gga.ps1": parent path "C:\Users\<user>\bin" for "C:\Users\<user>\bin\gga.ps1" is not a directory

Same error with `-strict-tdd -sdd-mode multi`, so it is not flag specific.

Auto-update log tail from the 2026-08-01 run, which upgraded fine and then stopped at sync:

  [ok] gentle-ai     0.0.0-...+dirty -> 2.2.4
  Config backup: upgrade-20260801T010011Z
  1 succeeded, 0 failed, 0 skipped.
  Updated to v2.2.4 - restart gentle-ai to continue.
  [2026-08-01 03:00:12] Running gentle-ai sync (rebases skills/agents to ~/.claude)...

(no further lines - the run never reaches its completion marker)
```

### Additional Context

Workaround that unblocks everything except one flag:

  gentle-ai install --agent claude-code --component skills

That resolves to engram,sdd,skills, never touches the gga shim, and completes normally. The cost is that `-strict-tdd` exists only on `sync` and is not exposed by `install`, so strict TDD mode is unreachable while this is broken.

Possibly related Windows reparse-point issues: #1682, #1818, #1604, #2121, #1918. Related but distinct link handling: #306 and PR #1077.

One unverified hypothesis, offered only as a starting point and not as a diagnosis: Go's os.Lstat on a Windows junction reports a reparse point rather than a directory, so an IsDir() check on the un-followed stat result would return false. I have not read the source.

I am happy to implement the fix and open a PR if you approve this issue.

Matere413:
## 🔗 Linked Issue

Closes #2164

---

## 🏷️ PR Type

- [ ] `type:feature` — New feature
- [ ] `type:refactor` — Code refactoring

---

## 📝 Summary

- Accept Windows directory junctions when their resolved target is an existing directory.
- Preserve rejection for broken junctions, loops, unsupported reparse points, and non-directory targets.
- Add Windows-specific regression coverage without changing non-Windows behavior.

---

## 📂 Changes

|-------------|-------------|
| `internal/components/filemerge/writer.go` | Delegates irregular parent classification to a platform-specific resolver. |
| `internal/components/filemerge/atomic_parent_windows.go` | Resolves valid Windows directory junction parents. |
| `internal/components/filemerge/atomic_parent_other.go` | Keeps non-Windows behavior unchanged. |
| `internal/components/filemerge/writer_junction_windows_test.go` | Covers valid, broken, looping, and non-directory junction targets. |

---

## 🧪 Test Plan

- [x] `go test ./internal/components/filemerge -count=1`
- [x] `go test ./internal/components/... -count=1`
- [x] `GOOS=windows GOARCH=amd64 go test -c ./internal/components/filemerge`
- [ ] Runtime execution on Windows/NTFS. The Windows suite cross-compiles successfully, but a real junction runtime was unavailable from macOS.
- E2E: N/A, this is a focused filesystem boundary covered by package tests.
- Benchmark validation: N/A, this does not change review lifecycle or benchmark behavior.

---

## ✅ Contributor Checklist

- [x] Appropriate `type:*` label requested
- [x] Relevant unit tests pass
- [x] Go format passes
- [x] E2E applicability explained
- [x] Benchmark applicability explained
- [x] Documentation is not required for this internal compatibility fix

---

## 💬 Notes for Reviewers

Please pay particular attention to the Windows reparse-point boundary. The resolver only accepts a junction when `os.Readlink` and `os.Stat` prove that its target exists and is a directory.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved atomic file writing on Windows when files are stored inside directory junctions.
  - Added validation to reject broken junctions, junctions targeting non-directories, and junction loops.
  - Preserved consistent behavior on non-Windows platforms.

- **Tests**
  - Added Windows coverage for valid directory junctions and unsafe junction targets.
  - Added automated verification that all junction-related tests are discovered and executed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,6 +199,23 @@ jobs:
           cache: true
           cache-dependency-path: go.sum
 
+      - name: Run native Windows junction parent tests
+        shell: pwsh
+        timeout-minutes: 5
+        run: |
+          $selector = '^(TestWriteFileAtomicFollowsDirectoryJunction|TestEnsureAtomicParentDirRejectsUnsafeDirectoryJunctionTargets|TestWriteFileAtomicRejectsJunctionLoop)$'
+          $expected = @(
+            'TestWriteFileAtomicFollowsDirectoryJunction'
+            'TestEnsureAtomicParentDirRejectsUnsafeDirectoryJunctionTargets'
+            'TestWriteFileAtomicRejectsJunctionLoop'
+          )
+          $selected = @(go test ./internal/components/filemerge -list $selector | Where-Object { $_ -match '^Test' })
+          if ($LASTEXITCODE -ne 0) { throw 'Could not list native Windows junction parent tests' }
+          $missing = @($expected | Where-Object { $_ -notin $selected })
+          if ($missing.Count -gt 0) { throw "Missing native Windows junction parent tests: $($missing -join ', ')" }
+          go test -v ./internal/components/filemerge -count=1 -run $selector
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
       - name: Build native test binary
         shell: pwsh
         run: go build -trimpath -ldflags="-s -w -X main.version=ci" -o "$env:RUNNER_TEMP\gentle-ai.exe" ./cmd/gentle-ai
--- a/internal/components/filemerge/atomic_parent_other.go
+++ b/internal/components/filemerge/atomic_parent_other.go
@@ -1,0 +1,9 @@
+//go:build !windows
+
+package filemerge
+
+import "os"
+
+func resolveAtomicParentJunction(string, os.FileInfo) (string, os.FileInfo, bool, error) {
+	return "", nil, false, nil
+}
--- a/internal/components/filemerge/atomic_parent_windows.go
+++ b/internal/components/filemerge/atomic_parent_windows.go
@@ -1,0 +1,31 @@
+//go:build windows
+
+package filemerge
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func resolveAtomicParentJunction(dir string, info os.FileInfo) (string, os.FileInfo, bool, error) {
+	if info.Mode()&os.ModeIrregular == 0 {
+		return "", nil, false, nil
+	}
+
+	// os.Readlink supports Windows symlinks and mount-point junctions. Other
+	// reparse-point tags return an error and remain rejected.
+	target, err := os.Readlink(dir)
+	if err != nil {
+		return "", nil, false, fmt.Errorf("resolving directory junction parent %q: %w", dir, err)
+	}
+	resolved, err := filepath.Abs(target)
+	if err != nil {
+		return "", nil, false, fmt.Errorf("resolve directory junction target %q: %w", target, err)
+	}
+	targetInfo, err := os.Stat(dir)
+	if err != nil {
+		return "", nil, false, fmt.Errorf("stat directory junction target %q: %w", resolved, err)
+	}
+	return resolved, targetInfo, true, nil
+}
--- a/internal/components/filemerge/atomic_parent_windows.go
+++ b/internal/components/filemerge/atomic_parent_windows.go
@@ -15,15 +15,14 @@ func resolveAtomicParentJunction(dir string, info os.FileInfo) (string, os.FileI
 
 	// os.Readlink supports Windows symlinks and mount-point junctions. Other
 	// reparse-point tags return an error and remain rejected.
-	target, err := os.Readlink(dir)
+	if _, err := os.Readlink(dir); err != nil {
+		return "", nil, false, fmt.Errorf("resolving directory junction parent %q: %w", dir, err)
+	}
+	resolved, err := filepath.EvalSymlinks(dir)
 	if err != nil {
 		return "", nil, false, fmt.Errorf("resolving directory junction parent %q: %w", dir, err)
 	}
-	resolved, err := filepath.Abs(target)
-	if err != nil {
-		return "", nil, false, fmt.Errorf("resolve directory junction target %q: %w", target, err)
-	}
-	targetInfo, err := os.Stat(dir)
+	targetInfo, err := os.Stat(resolved)
 	if err != nil {
 		return "", nil, false, fmt.Errorf("stat directory junction target %q: %w", resolved, err)
 	}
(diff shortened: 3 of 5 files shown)


Matere413:
Maintainer action needed: please apply the `type:bug` label.

The PR template and linked approved issue both classify this as a bug fix, but my fork contributor permissions cannot add labels to the upstream PR.

Alan-TheGentleman (changes requested):
The change is correctly placed at the shared atomic-parent boundary, and I added `type:bug`. The remaining blocker is native evidence: the current Windows Runtime workflow does not execute `internal/components/filemerge`, so its green result and local cross-compilation never create a junction. Refresh onto current main and add a native Windows CI selector covering valid, broken, non-directory, and looped junction parents. Cross-volume coverage can remain conditional on a second writable volume.

Matere413:
The three E2E jobs are failing before the tests run, during Docker Buildx bootstrap, with `HTTP 500 Internal Server Error` from the BuildKit daemon (`moby/buildkit:buildx-stable-1`) on ubuntu, arch, and fedora.

Could a maintainer please rerun these jobs? As the contributor from a fork, I do not have the required admin permissions to rerun them.

Alan-TheGentleman (changes requested):
Hey @Matere413! The approach is right and the CI step is genuinely good, so let me start there before the one thing I want changed.

**The CI step is the best part of this PR.** You could not run these on Windows from macOS, so instead of shipping untested code you added a runner step that *lists* the tests and fails if any expected one is missing, then runs them. That guards the real hazard with build-tagged files: they can silently drop out of selection and the suite stays green while covering nothing. Please keep it.

**The failing E2E jobs are not yours.** All three died in `Set up Docker Buildx`, which is infrastructure, and `main` is green on the same jobs. They need a re-run, not a fix.

## The change I need

`resolveAtomicParentJunction` diverges from the symlink branch sitting right above it in `ensureAtomicParentDir`, and the divergence matters:

```go
// existing symlink branch          // new junction branch
EvalSymlinks(dir)                   Readlink(dir) → filepath.Abs(target)
os.Stat(resolved)                   os.Stat(dir)
```

**`filepath.Abs` resolves against the process working directory, not against the junction's parent.** If the target is relative, `resolved` points somewhere unrelated, `writer.go` then does `dir = resolved`, and the atomic write lands in the wrong directory. Silently.

Directory junctions normally store absolute targets, so you may reasonably think this is unreachable. But your own comment says `os.Readlink` here also covers Windows symlinks, and directory symlinks *can* be relative. That is the reachable path.

**And `os.Stat(dir)` stats the junction while the error message names `resolved`.** `Stat` follows reparse points so the value usually matches, but the mismatch is a trap for the next reader and the message is already lying about what it looked at.

## Suggested fix

Use the same resolver the symlink branch already trusts:

```go
resolved, err := filepath.EvalSymlinks(dir)
if err != nil {
    return "", nil, false, fmt.Errorf("resolving directory junction parent %q: %w", dir, err)
}
targetInfo, err := os.Stat(resolved)
if err != nil {
    return "", nil, false, fmt.Errorf("stat directory junction target %q: %w", resolved, err)
}
return resolved, targetInfo, true, nil
```

`EvalSymlinks` handles relative targets, resolves chains, and fails on broken and looping targets, so your four rejection cases should still pass unchanged. Worth confirming the loop case specifically, since `EvalSymlinks` reports it differently than a `Readlink` + `Stat` pair would.

## Everything else

Verified locally: merges onto current `main` with no conflicts, `GOOS=windows go build ./...` and `go vet` clean, `go test ./internal/components/filemerge` green. The non-Windows stub keeps other platforms byte-for-byte unchanged, which is the right way to do this.

Fix that resolver and re-run the E2E and this is good to go.

Alan-TheGentleman (approved):
Addressed, and verified at the head (`5d48dece`) rather than from the PR diff view.

```go
if _, err := os.Readlink(dir); err != nil { ... }   // reparse-point probe only, target discarded
resolved, err := filepath.EvalSymlinks(dir)          // same resolver the symlink branch trusts
targetInfo, err := os.Stat(resolved)                 // stats what the message names
```

Both halves of what I asked for are in: `filepath.Abs` is gone, so a relative target can no longer resolve against the process working directory, and the `Stat` now matches the path its error names.

You also went past the request and added `TestWriteFileAtomicFollowsRelativeDirectoryJunction`, which stages exactly the case I argued was the reachable one. That is the part that makes this verifiable instead of merely correct-looking, and it is why I am approving rather than asking for coverage.

Keeping the CI listing step was the right call too.

Approving. 16/16 green.

Issue #2659, opened 2026-08-06
refactor(review): purify Snapshot.Identity to content-addressed form


Alan-TheGentleman:
## Problem

`Snapshot.Identity` is not purely content-addressed. `snapshotIdentityForProjection` (`internal/reviewtransaction/snapshot.go:1434-1457`) folds the intended-untracked path list and its `untrackedProof` into the identity hash, so two candidates with byte-identical `baseTree`/`candidateTree` can carry different identities when only their representation differs. The gate comparison itself is already content-based (`validateDerivedGate`, `internal/reviewtransaction/receipt.go:278-320`), and `FrozenCandidateContext` is already purely content-addressed; the identity hash is the one remaining place where representation leaks into authority recognition.

This is root 21 of #2471 (revised diagnosis in its thread with verified anchors). Known symptoms consistent with this impurity: #2091 (workspace target identity not re-derivable for identical content) and #1951 (staged-equivalent receipt shape at pre-commit). #1633 (closed) documents an earlier instance of the same class.

## Requested outcome

Purify the identity hash and retire the representation residue:

- Drop `intended[]` and `untrackedProof` from the identity domain in `snapshotIdentityForProjection`.
- Keep the `kind` and `projection` domain-separation tags. They prevent a `current-changes` receipt from conflating with a `base-workspace-overlay` review of identical bytes and are load-bearing.
- The untracked replay proof survives as side-band evidence only, for the recovery/replay flows that still consume it (`BuildStagedWorkspaceOverlayRecovery`, `BuildCorrectedCandidate`). It no longer participates in identity (maintainer decision D1, recorded in #2471).
- Version the identity domain tag so lineages frozen under the impure hash keep validating during migration. Historical identities and receipts remain parseable forever; stored authority is never rewritten.

## Expected behavior

A representation transition of identical content (untracked declared vs. staged) recomputes to the same identity, so `ValidateEvidence` (`snapshot.go:262-286`) keeps passing instead of failing closed on a candidate nobody changed.

## Actual behavior

Identity churns on representation facts, producing "snapshot paths, digests, or identity do not match Git tree evidence" failures and unrecognizable-authority shapes for content-identical candidates.

## Evidence

[...]

PR #2670, opened 2026-08-06
feat(review): forecast field in negotiated STATUS — kill the try/error/fix/retry loop

### LINKED ISSUE #2664: feat(review): forecast field in negotiated STATUS — kill the try/error/fix/retry loop
## Problem

Users experience every unmet precondition as one loop: try → typed error → apply the named fix → try again → next error. Each refusal is individually excellent (the refusal ratchet guarantees every one names its resolution or declares why it cannot), but their composition makes users traverse our state machine one rung at a time, discovering the ladder — or the wall at its end — by trial. A user heading into a dead-end state invests real work before learning no exit exists; finding the current list of such walls cost a manual 65-issue audit.

## Requested outcome

One diagnostic pass, compiler-style: evaluate ALL preconditions for the requested operation at once and return the complete ordered resolution plan alongside the single sanctioned next step.

## Design (agreed with maintainer, recorded here as the implementation contract)

- **Unified, not parallel**: an additive `forecast` field in the existing negotiated STATUS envelope, beside `next_transition`. NO separate diagnose verb — two precondition evaluators would drift, recreating the two-owners problem the waves deleted.
- **Go computes, exclusively**: the same native evaluator that derives `next_transition` produces the forecast in the same pass, from real state (Git trees, authority store, policy, capability manifest). No model participation in computing it — model-authored authority-adjacent data was #2442's root disease; the advisory boundary applies to information too.
- **Head executable, tail descriptive**: only `next_transition` carries exact runnable tokens. Forecast entries carry refusal codes and human descriptions, deliberately no invocations, so no orchestrator can batch-execute the tail without re-querying (the one-step negotiated contract exists because the world changes between steps).
- **Forecast, not promise**: recomputed on every STATUS query; later steps depend on post-step state, so partial horizons are reported honestly ("known so far, re-query after step N"). Docs must never present it as a guarantee.
- **Agents relay losslessly**: same contract as the consent envelope — translate to the user's language, never summarize, reorder, or invent steps. Without an agent, the CLI projects the same forecast human-readably. One source, two audiences.
- **Dead-ends self-report**: a forecast terminating in "no resolution exists" is a typed, countable signal naming the missing capability — the class of unrecoverable states becomes production-measured instead of backlog archaeology. The forecast does not unblock anything (root fixes do); it eliminates the discovery cost and generates the next roots table from data.
- **Deletion angle**: refusal sites become data rows consumed by one evaluator instead of N scattered early-returns each discovered by trial.

## Scope discipline

Before implementation: a read-only mechanism map of the current precondition evaluation sites with file:line anchors, per the `issue-root-resolution` skill. The map outranks this issue's assumptions.

## Origin

#2471 follow-up (the try→error→fix→retry observation), designed 2026-08-06.

decode2:
Closes #2664

### Summary of Changes

- Adds an additive `forecast` field to `ReviewTargetStatusResult` in `internal/cli/review_status_contract.go` and `contracts/review-integration/v1/schemas/status-v2.schema.json`.
- Computes the forecast exclusively in Go via `newReviewForecast` in `internal/cli/review_next_transition.go`:
  - **Head**: Step 1 guaranteed to match `next_transition.reason_code` and `kind`.
  - **Tail**: Steps > 1 carry only `step`, `kind`, `reason_code`, and human `description`—deliberately omitting runnable commands/arguments to avoid unverified batch execution.
  - **Horizon**: Set to `terminal` for stop transitions, `partial` for multi-step horizons.
- Updated `status-v2.fixture.json` and added unit tests in `internal/cli/review_next_transition_test.go`.

## Summary by CodeRabbit

* **New Features**
  * Review status responses now include an optional forecast of upcoming workflow steps.
  * Forecasts show the expected horizon, ordered actions, reason codes, and clear descriptions.
  * Forecasts support complete, partial, and terminal workflows, including immediate stop conditions.

* **Bug Fixes**
  * Improved visibility into next steps for multi-step reviews and stopped or invalid review states.
  * Forecasts now consistently align with the reported next transition.
  * Added validation to ensure forecast steps are complete, ordered, and internally consistent.


decode2:
@coderabbitai review

decode2:
@coderabbitai review

Alan-TheGentleman (changes requested):
Thanks for picking this up so fast, and the head/tail/horizon split is the right skeleton. I am requesting changes: three blockers, and one design question that decides how much of the tail survives.

**Blockers**

1. **The v2 STATUS payload no longer validates against its own published schema.** `review_facade.go:1014` sets `Forecast` for both contracts, but only `contracts/review-integration/v1/schemas/status-v2.schema.json` gained the property. `v2/schemas/status-v4.schema.json` and `status-v5.schema.json` both declare `"additionalProperties": false` and have no `forecast`. Verified by emitting a real payload: negotiated STATUS returns `"schema": "gentle-ai.review-integration.status/v5"` carrying a `forecast` object, so any consumer validating the envelope (gentle-pi, the OpenCode plugin, CI) rejects every negotiated STATUS. The existing test does not catch it because `validateAgainstPublishedStatusNextTransitionSchema` compiles only the `$defs.next_transition` subschema, never the whole document. Whatever else changes, that whole-document validation gap needs a test.

2. **`contracts/review-integration/v1` is byte-frozen and this edits it.** `FREEZE.md` in that directory states every file must stay byte-unchanged, and the two pinned SHA-256 drift guards in `review_provider_artifact_contract_test.go` exist to catch exactly this. Updating those pins to make the guard pass is the one move that is never available. The six sibling `status-v2-*.fixture.json` files are also left stale against the new output.

3. **`captured_verification_evidence_ready` does not exist.** It appears only in the three lines this PR adds. The real StateValidating execute reason is `captured_verification_evidence_passed` (`review_next_transition.go:260`), alongside `captured_verification_failed` and `captured_verification_tooling_failed`. A consumer keying on forecast reason codes matches nothing at that step.

**The design question, which matters more than the three above**

Four of the six tail projections do not match what the product actually does next:

- `captured_results_ready` to `delivery_gate_required` is unreachable: `delivery_gate_required` only comes from `StateApproved`, and finalizing from `StateReviewing` lands on Escalated, CorrectionRequired, or Validating (`compact.go:1060-1066`).
- `fresh_target_ready` to `reviewer_results_required` is wrong for every low-risk candidate, which freezes zero lenses and goes straight to `captured_results_ready`. `newReviewForecast` already receives `selectedLenses` and the tier and branches on neither.
- `correction_plan_required` to `targeted_validation_required` skips `correction_repository_verification_required`.
- `horizon: "complete"` is the default and only downgrades for seven reason codes, so collect transitions that obviously continue (`lineage_selection_required`, `empty_candidate_base_ref_required`, `final_verification_retry_authorization_required`, the recovery collections) get labelled complete. And since `Validate()` rejects multi-step forecasts with `complete`, that label can only ever mean "exactly one step", which is not what the enum reads as.

The whole point of the field is that a user stops discovering the ladder one error at a time. A tail that is confidently wrong at step 2 is worse than no tail: it replaces an honest unknown with a plausible lie, and an agent relaying it losslessly will relay the lie faithfully. So the ask is either derive the tail from real state (branch on tier, lens count, and the actual post-finalize states), or ship head plus `horizon` only and let the tail land when the derivation is real. The second option is a perfectly good first slice and I would merge it happily.

**One non-blocking note**: `newReviewForecast` recomputes `newReviewNextTransition` with identical arguments to the call the facade already made at `review_facade.go:1011`. It is pure today, so head agreement holds, but passing the computed head in removes both the duplicate classification pass and a load-bearing purity assumption.

Alan-TheGentleman:
Answering the `contracts/review-integration/v1` question specifically, because it turns out to resolve two of my blockers at once and it also requires me to correct something on our side.

**The v1 change should not exist at all.** `FREEZE.md` in that directory states the rule the forecast can follow instead of fighting: *"No new-lineage (v3) behavior consumes this contract; v3 uses `contracts/review-integration/v2/**` exclusively."* The forecast is new-lineage behavior, and #2664 asked for it in the negotiated STATUS envelope. Negotiated means the v2 contract. So the fix is to set `Forecast` only when the contract is not v1, and to add the property to `status-v4.schema.json` and `status-v5.schema.json` instead. That leaves v1 byte-untouched, leaves both pinned hashes alone, and simultaneously closes my first blocker, since the schemas that actually need the property are the two that currently reject the payload.

**Where I owe you a correction.** I flagged the freeze as an absolute byte rule, and our own merge last night did not meet that standard either: `fbb55080` (issue #2659) rewrote three fixtures under that same frozen directory, because purifying the identity hash legitimately changed values those fixtures record. So the line I should have drawn is not "never touch a byte", it is: **a fixture is a recording of what the product emits, and when the emitted value legitimately changes the recording follows it with a dated justification; a schema is the contract itself, and adding a property to a frozen contract's schema is a contract change.** Your PR is on the second side of that line and ours was on the first, but I was enforcing a distinction that is not written down anywhere. I am filing that clarification against the freeze document so the next contributor does not have to infer it from a review.

None of this changes the other blockers: the four tail projections that do not match real state, and `captured_verification_evidence_ready` not existing. Those are still the ones that decide whether the tail ships at all.

Alan-TheGentleman (approved):
The requested correction is now complete. The forecast is v2-only, one-step, and mirrors the already-computed `next_transition`; speculative tails and the invalid reason code are gone; v1 is byte-restored; v4/v5 schemas carry the property; CLI narration and lossless agent relay are covered.

The whole-document ratchet is `TestNativeStatusSchemasValidateWholeForecastEnvelope`: it compiles the actual v4/v5 schema roots, validates the full document, and proves unknown root properties are rejected. Full local, driven bench, and CI suites pass.

Issue #2174, opened 2026-08-01
bug(review): status offers review.start while effective global mode rejects it as disabled


Victor0451:
### Pre-flight Checklist

- [x] I searched open and closed issues for `rdd_disabled`, default review mode, disabled start, and inconsistent status/start behavior. I found related disabled-mode issues, but no duplicate of this status/start contradiction.
- [x] I understand that PRs require `status:approved` before implementation.

### Bug Description

On Gentle AI 2.2.4, the read-only review mode status and the negotiated review lifecycle disagree about the effective mode for the same repository and unchanged candidate.

`review mode status` reports receipt-driven development as on, decided by default, with both global and clone-local sources unset. Negotiated `review status --next-transition` then returns an executable `review.start` transition for the current workspace target. Executing that exact provider-returned transition immediately fails with typed code `rdd_disabled`, claiming the global mode source keeps review off.

This leaves the consumer unable to follow the authoritative `next_transition`, while the read-only mode surface provides no indication that review is disabled.

Observed evidence only, no root cause is asserted.

### Steps to Reproduce

1. Use a Git repository with a non-empty workspace candidate and no active review lineage for that target.
2. Run the read-only review mode status operation.
3. Observe that effective receipt-driven development is reported as on, decided by default, with global and clone-local unset.
4. Request negotiated review status using contract `gentle-ai.review-integration/v2` and `next-transition`.
5. Observe `next_transition.kind: execute`, `operation: review.start`, `reason_code: fresh_target_ready`.
6. Execute the exact returned operation and ordered arguments once.
7. Observe typed failure `rdd_disabled` before native mutation.

### Expected Behavior

The mode status, negotiated lifecycle status, and `review.start` preflight should resolve the same effective mode.

If review is disabled, the read-only surfaces should report that disposition and negotiated status should stop instead of offering an impossible start. If review is enabled by default, the exact returned start transition should not be rejected as globally disabled.

### Actual Behavior

Mode status:

```text
receipt-driven development: on (decided by default)
global: unset
clone-local: unset
```

Negotiated status:

```json
{
  "next_transition": {
    "kind": "execute",
    "reason_code": "fresh_target_ready",
    "execute": { "operation": "review.start" }
  },
  "target_identity": "sha256:5c42dda82cb9a0ff86e278e934c2395eb3d7d09ac50c346753c4fd36784073c5"
}
```

Exact returned start operation:

```json
{
[...]

PR #2705, opened 2026-08-07
fix(sdd): stop naming a refused operation in the maintainer_decision exit

### LINKED ISSUE #2530: fix(sdd): allow rescope after passed attempt exceeds line budget
### Pre-flight Checklist

- [x] Searched open and closed issues; no equivalent report was found.
- [x] This automated provider-defect report contains observed evidence only and does not assert an unconfirmed root cause.
- [x] Privacy review removed project names, paths, users, hosts, credentials, raw arguments, source, diffs, and environment values.

### 📝 Bug Description

A passed SDD runtime attempt that retains more changed lines than its immutable ceiling enters `maintainer_decision`. The settlement guidance tells the consumer to ask a maintainer to rescope or reset the objective. After the maintainer explicitly approves increasing only that ceiling, the advertised rescope operation refuses the state because the objective is already decision-required. Status names reset as the only continuation.

This leaves no way to apply the narrower maintainer decision that settlement itself requested without performing another exceptional reset.

### 🔄 Steps to Reproduce

1. Acquire one bounded SDD apply attempt with a 32-line ceiling.
2. Complete the objective successfully, retaining 35 changed lines including persisted SDD evidence.
3. Settle the attempt as passed with bounded test, runtime, cleanup, and process evidence.
4. Observe a blocked result requiring a maintainer decision and guidance that names rescope or reset.
5. Query status and observe `decision_required: true` with `next_action: reset`.
6. Obtain explicit maintainer approval to increase only the objective ceiling from 32 to 35 lines.
7. Invoke the audited rescope operation with exact current revision, unchanged work unit/evidence goal, one attempt, the approved 35-line ceiling, actor, and reason.
8. Observe a non-mutating refusal because rescope requires an objective that is not decision-required.

### ✅ Expected Behavior

The provider-owned continuation should be internally executable. Either the explicitly authorized rescope should accept this terminal over-budget state while preserving immutable attempt history and lifetime charges, or settlement/status should not advertise rescope and should return the one supported audited continuation with its exact consequence.

### ❌ Actual Behavior

```text
settle:
  state: blocked
  reason: maintainer_decision
  changed_lines: 35
  max_changed_lines: 32
  continuation guidance: ask a maintainer to rescope or reset

status:
  decision_required: true
  complete: false
  next_action: reset

rescope:
  refused because objective rescope requires an existing objective that is not decision-required or complete
```

The failed rescope does not mutate runtime state. The candidate, SDD artifacts, terminal attempt evidence, lifetime accounting, and authority revision remain preserved.

### Gentle AI Version

`2.3.0-rc.1` official published Linux AMD64 artifact.

### Operating System

Linux x86_64.

### AI Agent / Client

OpenCode.

### 📋 Affected Area

CLI SDD runtime attempt lifecycle.

### Bounded Evidence

- Objective budget: 1 attempt, 32 changed lines.
- Terminal outcome: passed.
- Native retained accounting: 35 changed lines.
- Maintainer decision: approve only a 35-line ceiling for the same work unit and evidence goal.
- Rescope outcome: non-mutating refusal.
- Candidate mutation after refusal: none.
- Additional actor, retry, reset, review, commit, push, or authority-file edit after refusal: none.

### Related Reports Reviewed

- #2298 is not equivalent: it covers interrupted work reverted to zero retained lines before rescope.
- #1974 is not equivalent: it covers failed verification with remaining attempt budget.

This report covers a passed, retained, over-budget attempt whose own settlement guidance offers a rescope operation that rejects the resulting decision-required state.

aparragithub:
Closes #2530.

> **Note on process:** #2530 is not yet labelled `status:approved`. This PR is opened early so the proposed shape is concrete and reviewable — happy to hold, rework, or close it if the maintainers want the approval gate respected first, or if they prefer the alternative shape described below.

## The defect

`compactBlockedExitText`'s `CompactBlockMaintainerDecision` arm named `rescope or reset the objective`.

That arm is reached from exactly one call site — `compactTerminalState`'s `DecisionRequired` arm (`runtime_compact.go:163`) — and `runtimeObjectiveRescopeStructurallyPermitted` (`runtime_ledger.go:1260`) opens with:

```go
if status.DecisionRequired || status.Complete {
	return false
}
```

So the exit named `rescope` in the one state where `rescope` is structurally guaranteed to be refused. A consumer that followed the advertised exit hit a non-mutating refusal and had no continuation left, which is the dead end #2530 reported.

## Why the text could not have been right

`compactBlockedExitText(reason CompactBlockReason, token string)` never receives the `RuntimeStatus`. It has no way to distinguish which of the two operations is admissible for the objective in hand, so it named both unconditionally.

This is the same class as #2174 — a surface offering an operation the effective state rejects.

## The fix

Defer to the field that already carries the answer. `Status` publishes `next_action`, and the `sdd-attempt status` command was already being named in this same exit for the accounting, so the caller is sent to a place that resolves the block rather than to a guessed operation. This is the convention `runtimeLedgerStatusPointer` already applies elsewhere in the file: *name the status, let `next_action` name the operation.*

No lifecycle behavior changes — this is guidance text plus its regression guard.

## Why not the other option

#2530's "Expected Behavior" offered two exits: let `rescope` accept this state, or stop advertising it. This PR takes the second.

The first would reopen the budget-laundering path the surrounding code explicitly guards. `runtimeObjectiveResetAdmissible`'s comment records that reset is refused one layer deeper for a non-drifted candidate precisely so an elective early reset cannot launder the per-objective budget, and `Rescope`'s own doc comment scopes it to a *terminal, non-complete, zero-drift* objective. Admitting `rescope` into the decision-required state would be that same laundering under a different name.

## Alternative shape considered

Widening `compactBlockedExitText` to take the `RuntimeStatus` and name only the admissible operation — mirroring what `runtimeObjectiveChangeRefusal` (`runtime_ledger.go:1077`) already does — would let this arm print a complete command. It was not taken here because it changes the signature for all 20 call sites to serve one arm, and `next_action` already exists as the published answer. Glad to switch if maintainers prefer the stronger form.

## Test

`TestMaintainerDecisionExitNamesNoRefusedOperation` drives a real store to a decision-required objective and asserts against the admissibility predicates rather than a fixed string:

- `status.DecisionRequired` is true and `next_action` is `reset` (the premise holds)
- `runtimeObjectiveRescopeStructurallyPermitted` is **false** for that same status
- `runtimeObjectiveResetAdmissible` is **true** (the state does have a continuation)
- the exit does not name `rescope`, and still points at `sdd-attempt status` / `next_action`

The precondition assertions mean the test fails loudly if the underlying lifecycle rules ever change, instead of silently guarding a string.

Verified RED before the fix:

```
--- FAIL: TestMaintainerDecisionExitNamesNoRefusedOperation
    maintainer_decision exit names rescope, which
    runtimeObjectiveRescopeStructurallyPermitted refuses for every
    decision-required objective
```

## Verification

- `go build ./...` — clean
- `gofmt -l ./internal/ ./cmd/` — clean
- `go vet ./internal/sddstatus/` — clean
- focused: `TestMaintainerDecisionExitNamesNoRefusedOperation` and `TestSDDStatusStillStopsForAMaintainerDecision` both pass
- `go test ./...` — all packages pass except three in `internal/sddstatus` that also fail on unmodified `origin/main` in my environment (`TestReviewOfferDeclineNeverBlocksArchiveAtTheProjectionLevel`, `TestReviewOfferPresentAndArchiveReadyForAGenuinelyMissingReceipt`, `TestRuntimeDisabledUnmanagedRemediationConsumesTheOnlyRemainingAttempt`). All three concern `ReviewOfferBlock.Available` and appear to depend on machine-global review mode, which is disabled here. Confirmed pre-existing by stashing this branch's changes and re-running them on clean `origin/main`. Not touched by this PR.

## Rollback

Revert the single commit — the change is one exit string plus one new test file, with no persisted state, schema, or lifecycle surface involved.

Changed lines: 102 additions + 2 deletions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the maintainer-decision blocked message to direct callers to the audited status and `next_action` continuation.
  * Removed guidance to use a rejected rescope or reset operation.

* **Tests**
  * Added regression coverage to verify the decision-required state and ensure the compact exit provides the correct continuation guidance.

### DIFF
--- a/internal/sddstatus/runtime_compact.go
+++ b/internal/sddstatus/runtime_compact.go
@@ -486,9 +486,26 @@ func compactBlockedExitText(reason CompactBlockReason, token string) string {
 			"`gentle-ai sdd-attempt status --cwd <repo> --change <change>` to see the live attempt and its " +
 			"current revision, then reissue this call against that state"
 	case CompactBlockMaintainerDecision:
+		// #2530: this arm used to name "rescope or reset the objective". It is
+		// reached from exactly one call site -- compactTerminalState's
+		// DecisionRequired arm -- and runtimeObjectiveRescopeStructurallyPermitted
+		// refuses every decision-required objective, so half of that pair was an
+		// operation the runtime is guaranteed to reject in the only state that
+		// can print it. Naming a refused operation is the dead end this file's
+		// exit-naming audit exists to prevent.
+		//
+		// compactBlockedExitText is a pure function of (reason, token) and never
+		// receives the RuntimeStatus, so it cannot tell which continuation is
+		// admissible here. Rather than widen the signature for one arm, it defers
+		// to the field that already carries the answer: Status publishes
+		// next_action, and the same status command was already being named for
+		// the accounting. This is the convention runtimeLedgerStatusPointer
+		// applies everywhere else -- name the status, let next_action name the
+		// operation.
 		return "this work unit's attempt or changed-line budget needs a maintainer decision; run " +
-			"`gentle-ai sdd-attempt status --cwd <repo> --change <change>` for the accounting, then ask a " +
-			"maintainer to rescope or reset the objective, or run " + compactBlockedReviewModeDisableExit
+			"`gentle-ai sdd-attempt status --cwd <repo> --change <change>` for the accounting and for " +
+			"next_action, which names the one audited continuation this objective admits, then ask a " +
+			"maintainer to authorize that continuation, or run " + compactBlockedReviewModeDisableExit
 	case CompactBlockActiveAttempt:
 		// Adversarial finding F2: the bare `sdd-attempt acquire --token <t>`
 		// / `settle --token <t>` forms are not complete commands -- each
--- a/internal/sddstatus/runtime_compact_maintainer_exit_test.go
+++ b/internal/sddstatus/runtime_compact_maintainer_exit_test.go
@@ -1,0 +1,83 @@
+package sddstatus
+
+import (
+	"context"
+	"strings"
+	"testing"
+)
+
+// TestMaintainerDecisionExitNamesNoRefusedOperation is #2530's regression guard.
+//
+// The maintainer_decision exit used to name "rescope or reset the objective".
+// That text is emitted from exactly one call site — the DecisionRequired arm of
+// compactTerminalState — and runtimeObjectiveRescopeStructurallyPermitted
+// refuses every decision-required objective. So the advertised exit named an
+// operation the runtime is guaranteed to reject in the only state that can
+// produce it, which is the dead end #2530 reported.
+//
+// The assertion is deliberately tied to the admissibility predicates rather
+// than to a fixed string: the exit may name an operation only when that
+// operation would actually be accepted for this same status.
+func TestMaintainerDecisionExitNamesNoRefusedOperation(t *testing.T) {
+	repo := initRuntimeLedgerRepo(t)
+	seedReadyChange(t, repo, "maintainer-exit", "- [ ] 1.1 Work\n")
+	store := mustRuntimeStore(t, repo, "maintainer-exit")
+
+	active, err := store.Begin(context.Background(), BeginAttemptRequest{
+		RequestID: "begin-maintainer-exit", WorkUnit: "apply-auth",
+		EvidenceGoal: "prove the auth runtime", MaxAttempts: 1, MaxChangedLines: 20,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, err := store.Finish(context.Background(), FinishAttemptRequest{
+		ExpectedRevision: active.Revision, RequestID: "finish-maintainer-exit", Outcome: AttemptFailed,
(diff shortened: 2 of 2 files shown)


aparragithub:
Closing this as superseded by #2770, which landed the same fix earlier today.

I checked the code rather than assuming the closed issue meant the work was done. #2770 rewrites the same `case CompactBlockMaintainerDecision` arm in `internal/sddstatus/runtime_compact.go`, for the same reason this PR gave: `runtimeObjectiveRescopeStructurallyPermitted` refuses every decision-required objective, so naming rescope pointed at an operation the runtime is guaranteed to reject in the only state that can print it. On current `main` (`3d1e6735`) the arm reads "ask a maintainer to reset the objective", so the refused operation is gone and there is no live defect left here.

The one difference is presentation, not behavior. This branch deferred to `next_action` so the text stays correct if the set of admitted continuations ever grows; #2770 names `reset` directly, which is more concrete for the reader today. That is a judgment call, and the merged version is a reasonable one, so it is not worth a competing PR. If the admitted set does widen later, the `next_action` indirection is worth revisiting then.

No action needed from anyone. Thanks for the review time on it.

PR #2715, opened 2026-08-07
fix(update): swap Windows binary when destination is held (#2319 swap-half)

### LINKED ISSUE #2319: fix(tui): upgrade reports success on Windows but binary is not replaced (atomic rename silently fails + stale registry channel)
## Pre-flight Checklist

## 📝 Bug Description

On Windows, when the user accepts the "Update available" prompt inside the gentle-ai TUI, the operation **reports success but the binary does not change**. The staged replacement files (`gentle-ai.<hash>.new.exe` and `gentle-ai.<hash>.vcs.exe`) are left stranded in the install directory, **smaller AND older** than the binary already running. The TUI confirms the update was applied and exits cleanly with no error.

The defect has two coupled root causes that must both be addressed to ship a fix:

1. **Atomic replace is silently failing on Windows.** The two-stage updater at `internal/update/upgrade/strategy.go:660` (`downloadAndReplace` -> `Download` -> `atomicReplace`) returns success based on the *download* stage; the subsequent `os.Rename(.new.exe, gentle-ai.exe)` fails on Windows when the destination is held open, when ACLs differ, or when the rename is gated by antivirus / Windows Search / indexer. The repo''s own test at `internal/update/upgrade/download_test.go:202-205` explicitly skips the Windows path with the comment *"atomic replace uses rename — Windows behavior is different"*. That is the only piece of unverified behavior in the updater, and it ships with the platform.

2. **The TUI does not surface the swap failure.** Because the staged file landed and the process exit code is 0, the success envelope is sent even though the atomic rename was the operation the user actually asked for. The TUI should distinguish "staged" from "activated" and refuse to report success on partial state.

Additionally, the candidate offered in this case (`v2.1.11+statusfix.e424f5d8`, dated 2026-07-26) is older than the binary already running (`v2.2.3`, 2026-07-31). The registry-driven channel that surfaced this candidate is stale; the TUI''s "Update available" line is showing a downgrade against the user''s local state. This is a third observable symptom that should be addressed alongside the silent-no-op: the registry channel refresh path needs to be reconciled before the TUI offers an upgrade.

## 🔄 Steps to Reproduce

1. Install gentle-ai on Windows via the official PowerShell installer (`irm https://gentle-ai.com/install.ps1 | iex`). Confirm the binary path is `C:\Users\<user>\AppData\Local\gentle-ai\bin\gentle-ai.exe`.
2. Run `gentle-ai.exe --version`. Note the current version.
3. Open the TUI (`gentle-ai.exe` with no args, or via the configured shell entry point).
4. Wait for the TUI to show `Update available: <old> -> <new>` and accept the upgrade.
5. The TUI reports `Update applied` (or equivalent success copy) and exits to the menu without an error code.
6. Re-run `gentle-ai.exe --version`. The version is unchanged from step 2.
7. Inspect the install directory and observe stranded `.new.exe` and `.vcs.exe` files, smaller and older than the active `gentle-ai.exe`.

## ✅ Expected Behavior

After accepting the TUI''s update prompt:

- The active `gentle-ai.exe` reports the new version (matching `gentle-ai --version` output to what the TUI offered).
- No stranded `.new.exe` / `.vcs.exe` files remain in the install directory on success.
- If the atomic replace cannot complete (file held, ACL mismatch, AV lock, etc.), the TUI reports the failure explicitly with the specific cause, **does not** claim success, and leaves the prior binary in place.
- The candidate offered by the TUI is never an older or smaller binary than what the user already has on disk.

## ❌ Actual Behavior

- The TUI prints success and exits with code 0.
- The active binary keeps reporting the prior version.
- `gentle-ai.<hash>.new.exe` and `gentle-ai.<hash>.vcs.exe` are left behind, dated before the active binary.
- The candidate offered for upgrade was v2.1.11+statusfix.<hash> (~16 MB, 26/07/2026) while the running binary was v2.2.3 (~23 MB, 31/07/2026). The TUI was offering a downgrade and did not flag it.

## 🖥️ Environment

Gentle AI Version: v2.2.3 (active); staged candidate was v2.1.11+statusfix.<hash> (older, smaller)

Operating System: Windows

AI Agent / Client: Other

Affected Area: TUI (terminal UI)

## 💡 Logs / Error Output

```powershell
PS> Get-ChildItem "$env:LOCALAPPDATA\gentle-ai\bin" | Format-Table Name,Length,LastWriteTime -AutoSize

Name                                    Length LastWriteTime
----                                    ------ -------------
gentle-ai.<hash>.new.exe                ~16 MB  26/07/2026 09:14
gentle-ai.<hash>.vcs.exe                ~16 MB  26/07/2026 09:15
gentle-ai.exe                           ~23 MB  31/07/2026 17:24
gentle-ai.exe.bak.2.1.11.20260726-101416 ~16 MB  26/07/2026 09:16
gentle-ai.exe.pre-refresh-20260727..... ~17 MB  26/07/2026 10:14
gentle-ai.exe.pre-v2.2.0-20260727193135  ~17 MB  27/07/2026 17:00
```

## Additional Context

Adjacent issues for context (none covers this exact drift as `status:approved` today):

- **#724** "Decoupled auto-update pipeline for Windows" — open, `status:needs-design`. Broader redesign (shared `gentle-update` shim). Relevant; the silent-no-op fix is a precondition for the redesign to be useful.
- **#2216** "Binary installs rename without fsync" — open, `status:needs-design`, `priority:high`. Covers crash corruption but not silent no-op.
- **#2117** "bug(sdd): tasks executor returns empty result on v2.2.4" — closed, was approved. Different surface (sdd-executor).
- **#951** "doctor: PATH-shadow detection misses Windows executables" — closed, was approved. Adjacent (PATH-shadow) but different defect.
- **#1975** "(Installer) gentle-ai upgrade fails to update active binary due to PATH priority conflict" — closed. PATH-priority variant of upgrade failure; the silent-no-op surface is a *separate* upstream defect.

Workaround applied locally: `go install github.com/gentleman-programming/gentle-ai/v2@<tag>` from an elevated PowerShell replaces the binary correctly without involving the TUI. The repo does not currently publish Windows release assets, so `go install` from source is the only currently-working Windows distribution channel regardless of this drift. A PowerShell helper that automates detection, cleanup, `go install`, and verification is available on request.

## 🖥️ Environment (re-verified 2026-08-22)

- **Gentle AI Version:** `gentle-ai 2.4.1-0.20260821223000-ec3b1db7aa6d` (main @ ec3b1db7)
- **Status:** Still applies — verified by code inspection against current main

ardelperal:
## Linked Issue

Fixes #2319

Refs #3509 — the registry-channel half of the umbrella issue. This PR ships the swap-half (atomic-rename path on Windows) only; the registry-channel half remains an open follow-up.

---

## PR Type

- [x] `type:fix` — Bug fix (non-breaking change that fixes an issue)

---

## Summary

Per #2319, on Windows when the user's destination binary is held (e.g. by another process or a permission lock), the update flow fails to swap. This PR moves the destination aside with an OS-level atomic rename when the destination already exists and is in use, falling back to copy + remove when a same-volume rename is unavailable. The change is scoped to the `filemerge` package plus the `upgrade/download` step that consumes it; no new contracts, no new surfaces.

Only the bugfix requested by #2319 (swap-half). The registry-channel half is tracked in #3509 and out of scope here. No architecture decisions, no new modules, no new abstractions.

---

## Changes
| File | Change |
|------|--------|
| `internal/components/filemerge/move.go` | Entry point that picks the right OS-specific move path. |
| `internal/components/filemerge/move_unix.go` | Unix implementation: `rename(2)` (same-volume atomic) with copy+remove fallback. |
| `internal/components/filemerge/move_windows.go` | Windows implementation: `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING`; same-volume rename first, copy+remove fallback. |
| `internal/components/filemerge/move_windows_test.go` | Tests for the Windows path. |
| `internal/components/filemerge/writer.go` | Wire the new move into the filemerge writer. |

### DIFF
--- a/internal/components/filemerge/move.go
+++ b/internal/components/filemerge/move.go
@@ -1,0 +1,22 @@
+package filemerge
+
+// MoveFileReplace publishes src over dst, replacing whatever dst holds.
+//
+// On Unix this is os.Rename and nothing else: a same-directory rename is already
+// atomic and already replaces an executing binary, because the running process
+// keeps its descriptor to the old inode while the directory entry moves.
+//
+// On Windows os.Rename is only the first attempt. A destination held by
+// antivirus, the search indexer, or the running image itself makes MoveFileEx
+// fail with ERROR_ACCESS_DENIED, ERROR_SHARING_VIOLATION, or
+// ERROR_LOCK_VIOLATION, and those are the shapes #2319 reported as "upgrade
+// succeeded, binary unchanged". moveFileReplace on that platform walks a
+// fallback chain instead of surrendering; see move_windows.go for the order and
+// for why MOVEFILE_DELAY_UNTIL_REBOOT is not part of it.
+//
+// This function is the mechanism, never the guarantee. The caller still has to
+// read the destination back — a move that returns nil has not necessarily taken
+// effect, and replaceDurably is where that is proven.
+func MoveFileReplace(src, dst string) error {
+	return moveFileReplace(src, dst)
+}
--- a/internal/components/filemerge/move.go
+++ b/internal/components/filemerge/move.go
@@ -1,5 +1,22 @@
 package filemerge
 
+// DisplacedSuffix is the deterministic filename the Windows move path uses
+// when a held destination cannot be overwritten in place. The previous binary
+// is renamed to <dst><DisplacedSuffix> so the new binary can take dst's
+// place; the suffix is deterministic (not random) so a later run knows
+// exactly which file is the recovery path.
+//
+// The cleanup of any leftover DisplacedSuffix file is the caller's
+// responsibility — after read-back verification, never before. Decoded as:
+//
+//	<recovery-state> = <dst><DisplacedSuffix>  (always, on Windows displacement)
+//	                  ∅                       (when the move succeeded without
+//	                                          hitting the displacement rung)
+//
+// and the caller reads dst, proves the swap, then os.Remove(<dst><DisplacedSuffix>)
+// which is a no-op when no displacement happened.
+const DisplacedSuffix = ".gentle-ai-displaced"
+
 // MoveFileReplace publishes src over dst, replacing whatever dst holds.
 //
 // On Unix this is os.Rename and nothing else: a same-directory rename is already
@@ -13,6 +30,11 @@ package filemerge
 // succeeded, binary unchanged". moveFileReplace on that platform walks a
 // fallback chain instead of surrendering; see move_windows.go for the order and
 // for why MOVEFILE_DELAY_UNTIL_REBOOT is not part of it.
+//
+// On the Windows displacement rung the prior binary is left at
+// <dst><DisplacedSuffix> until the caller proves the swap via read-back. This
+// is the contract that decode2 (2026-08-18 PR #2715) review required: the
+// cleanup of the displaced file is the caller's privilege, never the move's.
 //
 // This function is the mechanism, never the guarantee. The caller still has to
 // read the destination back — a move that returns nil has not necessarily taken
--- a/internal/components/filemerge/move_unix.go
+++ b/internal/components/filemerge/move_unix.go
@@ -1,0 +1,13 @@
+//go:build !windows
+
+package filemerge
+
+import "os"
+
+// moveFileReplace is a plain rename everywhere except Windows. Nothing is added
+// here on purpose: POSIX rename(2) already replaces the destination atomically,
+// including a destination that is currently being executed, so a fallback chain
+// would only add ways to be wrong.
+func moveFileReplace(src, dst string) error {
+	return os.Rename(src, dst)
+}
--- a/internal/components/filemerge/move_windows.go
+++ b/internal/components/filemerge/move_windows.go
@@ -12,10 +12,11 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-// displacedSuffix names the destination that was moved out of the way so a held
-// file could be replaced. It is deterministic rather than random so a later run
-// can clear a leftover instead of accumulating one per attempt.
-const displacedSuffix = ".gentle-ai-displaced"
+// displacedSuffixLocal is the package-private alias for filemerge.DisplacedSuffix.
+// The suffix value lives in move.go so non-Windows callers (and tests) can
+// reference it without dragging in <golang.org/x/sys/windows>; the build-tagged
+// file here just reads from the public constant.
+const displacedSuffixLocal = DisplacedSuffix
 
 // moveFileReplace publishes src over dst on Windows.
 //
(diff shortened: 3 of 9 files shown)


decode2 (changes requested):
Requesting changes for rollback safety. `displaceAndMove` can leave the prior binary only at `<dst>.gentle-ai-displaced` after a failed source move and failed restoration, but the next attempt unconditionally deletes that recovery path. It also removes the backup before the caller verifies the destination digest.

Please preserve invocation-owned recovery state through read-back verification and add native Windows fault-injection coverage for failed move, failed restore, retry, and cleanup. Current `binaryUpgrade` also rejects the relevant Windows update path before `Download` runs, so this PR cannot independently close #2319 without its enabling companion and distribution work.

ardelperal:
decode2 review 2026-08-18 addressed. AI-slop discipline applied after the fact: every line traces to one of decode2's three bullet points.

**Preserve displacement through read-back.** `moveFileReplace` (Windows displacement rung) used to remove `<dst>.gentle-ai-displaced` BEFORE returning success. The cleanup now lives in `atomicReplace` in `internal/update/upgrade/download.go`, gated on the read-back actually matching the staged bytes. A failed read-back preserves the displaced file as a recovery path; a successful one deletes it.

**DisplacedSuffix exported** from `internal/components/filemerge/move.go` so the `atomicReplace` cleanup references a single source of truth instead of recomputing the suffix string at the call site.

**Tests added** in `internal/update/upgrade/download_durability_test.go`:

- `TestAtomicReplacePreservesDisplacedBackupOnReadbackFailure` — pre-stages the displaced file with recovery content, runs `atomicReplace` with renameFn no-op + src/dst content mismatch, asserts the digest-mismatch error fires AND the displaced file is intact.
- `TestAtomicReplaceRemovesDisplacedBackupAfterReadbackSuccess` — pre-stages the displaced file with junk content, success path, asserts the file is gone after `atomicReplace` returns.
- `TestAtomicReplaceNoDisplacedBeforeNoCleanup` — regression guard for the clean-rename case so the cleanup branch cannot throw on a non-existent path.

Existing `move_windows_test.go` already covers the Windows fault-injection for failed-move (running-image case in `TestMoveFileReplaceReplacesARunningExecutable`) and failed-restore (`TestMoveFileReplaceRefusesToLoseAnUndisplaceableDestination`). Retry coverage cannot be added from Linux without MoveFileEx fault injection; flagged for a Windows CI contributor.

**PR body**: `Closes #2319` -> `Refs #2319` per decode2's note that `binaryUpgrade` rejects the relevant Windows update path before `Download` runs, so this PR cannot independently close #2319 without its enabling companion slice + distribution work.

ardelperal:
Tested against main @ ec3b1db7aa6d (gentle-ai 2.4.1-0.20260821223000). Windows swap fix is unique — no supersession.

ardelperal:
Closing this PR as superseded/foreign-fix. v2.5.0-rc.1 (released 2026-08-26, tag sha `7afe50d1`) closes the linked #2319 with the opposite approach from this PR's slice. Instead of fixing the atomic Windows swap, rc.1 bypasses it entirely:

- `internal/update/upgrade/strategy.go:binaryUpgrade` returns `ManualFallbackError` with `gentleAIWindowsSourceInstallHint` for Windows self-upgrades — surfaces as `UpgradeSkipped`, not `UpgradeFailed`.
- `internal/update/upgrade/download.go:Download` early-returns on Windows with `"upgrade %q on Windows requires manual update — %s"`.
- `internal/update/upgrade/strategy.go:goInstallUpgrade` adds `preflightWindowsGentleAIGoInstallWithDestination` that verifies Go owns the active executable before writing, or skips to manual recovery.

Four code comments in rc.1 explicitly reference #2319 (`renameFn` testability, `atomicReplace` Windows-hold, `binaryUpgrade` strategy table, `preflightWindowsGentleAIGoInstallWithDestination`). The swap-fix surface this PR implemented no longer exists in shipped behavior — Windows self-upgrade is now an explicit manual step (download from GitHub releases / `go install`), not an automated atomic rename. The PR's CI is green but merging would land code against a retired surface. Same pattern as #2994 / #2761 / #2928 closed 2026-08-22/23.

Issue #2685, opened 2026-08-07
fix(install): nothing detects installed assets older than the running binary


Alan-TheGentleman:
## Problem

Upgrading the gentle-ai binary does not refresh installed managed runtime assets, and nothing detects the resulting skew. The user then hits failures produced by a previous version's assets while running a current binary, with error text that no longer exists anywhere in the product.

Confirmed occurrence: #2442's rc.2 report. The reporter ran the rc.2 binary and received `review task binding does not match the selected lens`, a message that shipped in rc.1's OpenCode plugin, was replaced by the per-field parser in `ea6ed3f8` three hours after rc.1 was tagged, and was deleted entirely in `4a65096e` for rc.2. Their plugin was rc.1's; their binary was rc.2's. Nothing anywhere told them.

This is worse than an ordinary stale file, because the failure is attributed to the current release. It costs the reporter their time and us a false defect signal against a version that cannot produce the behavior.

## Contributing factor we own

The rc.2 release notes instruct downloading the binary, `chmod`, and `mv` into place. That is a complete binary upgrade and a silent no-op for every managed asset, so following our own published instructions produces exactly this skew.

## Requested outcome

- Record the binary version that wrote each managed asset set, and detect skew against the running binary.
- On skew, refuse or warn at the point of use with the actionable next step (`gentle-ai sync`), never a message from the older vintage. Fail-closed for the review lifecycle specifically, since a stale reviewer transport can misreport a review's outcome.
- Update the install and upgrade documentation, including release notes, so a binary upgrade always names the asset refresh step.

## Related

- #2442 (the occurrence and its evidence)
- Root 22 in #2471 (installed assets promising what the binary does not ship). This is the version-skew facet of the same class: there the asset named a capability that never existed, here it names one that no longer does.

PatricioPastor:
Additional privacy-scrubbed occurrence on Gentle AI 2.3.0-rc.3. This appears equivalent to the installed-binary/managed-asset skew tracked here, so I am adding evidence rather than opening a duplicate. This reports observed behavior only; no root cause is asserted.

## Environment

- Gentle AI release candidate: 2.3.0-rc.3 (SHA-256 verified by the reporter)
- Client: OpenCode Task tool
- OS/architecture: Windows (win32), PowerShell 5.1
- Review contract: gentle-ai.review-integration/v2
- Projection: workspace

## Operation shape and bounded outcome

[...]

Issue #2492, opened 2026-08-04
review: the stop-reason invariant test cross-checks the wrong table, and 11 of 16 rows disagree


Alan-TheGentleman:
The stop-reason continuation table exists in two places with the same 16 codes but materially different text, and the fail-closed guard checks the copy rather than the contract.

## The two tables

- `docs/review-integration.md` -- the explanatory copy.
- `internal/assets/skills/_shared/review-ledger-contract.md` -- **the shipped asset consumers actually receive**, and the one the orchestrator contract tells them to read.

Both contain all 16 codes. Neither has a row the other lacks.

## The disagreement

Compared exhaustively, not by sampling. **11 of 16 rows differ on at least one load-bearing signal**: 10 disagree on whether the row is terminal, 10 disagree on whether a runnable escape is named, 9 disagree on both.

`missing_authority_binding`

- docs: `Terminal -- internal invariant violation: applicability was current_target but no authority binding resolved. File a defect; there is no caller-side retry.`
- asset: `Internal invariant violation with no caller-side retry. File a defect with the lineage id, or run gentle-ai review mode disable --scope clone --cwd <repo> to deliver under ordinary policy instead.`

`manual_intervention_required`

- docs: `Terminal -- the authority state is not one of this negotiated protocol's known states. Requires maintainer review of the lineage.`
- asset: `Authority state this protocol does not recognize. Ask a maintainer to review the lineage, or run gentle-ai review mode disable --scope clone --cwd <repo> to deliver under ordinary policy instead.`

`native_stop_required`

- docs: `Terminal -- the authority state accepts no automated action from this negotiation. Requires maintainer review of the lineage before any further command.`
- asset: `Escalated lineage not yet eligible for automated action. Ask a maintainer to review it, or run gentle-ai review mode disable --scope clone --cwd <repo> to deliver under ordinary policy instead.`

This is not a length difference. The docs say no continuation exists; the asset names one, and it runs.

Mismatched: `captured_artifacts_unverifiable`, `captured_result_selection_unavailable`, `captured_verification_evidence_invalid`, `corrupted_or_unverifiable_authority`, `final_verification_retry_unavailable`, `manual_intervention_required`, `missing_authority_binding`, `native_stop_required`, `unchanged_or_unverified_authority`, plus `staged_workspace_overlay_recovery_unavailable` (terminal only) and `recovery_scope_unchanged` (escape only).

[...]

PR #2728, opened 2026-08-07
fix(install): detect installed assets older than running binary (#2685)

### LINKED ISSUE #3147: bug(doctor): nothing surfaces managed assets older than the running binary before a review hits the skew
Successor to #2685, which was closed when the review path gained its own guard. Scoped to what remains: the user-facing detection.

## What exists today

`review.start` preflight refuses on skew (`managed_assets_outdated`, continuation: run `gentle-ai sync`). Verified working in the field — an occurrence commented on #2685 on 2026-08-12 shows exactly that refusal doing its job.

## What is missing

The user finds out **mid-review**, at START, after consent. `gentle-ai doctor` — the surface whose whole purpose is answering "is this installation healthy" — reports nothing about assets written by an older binary. The rc.7 occurrence on #2492 is the cost: an upgraded binary consulted orchestrator instructions from an rc.6-era sync, and nothing anywhere said so until a review dead-ended.

## Expected

- `sync` stamps the binary version that performed it into the install state.
- `doctor` compares that stamp against the running binary and warns with the exact continuation (`gentle-ai sync`) when they diverge, passes quietly when they match, and skips honestly when no stamp exists (legacy state).

PR #2728 implements exactly this shape.

Cobies:
## Summary of Changes

This PR resolves approved issue #2685 (`fix(install): nothing detects installed assets older than the running binary`):

- **`internal/state/state.go`**: Added `InstalledVersion string \`json:"installed_version,omitempty"\`` field to `InstallState` struct to track the `gentle-ai` binary version that installed or last synced managed runtime assets.
- **`internal/cli/run.go` & `internal/cli/sync.go`**: Updated `RunInstall` and `RunSync` to record `cli.AppVersion` in `InstalledVersion` upon successful completion and write to `state.json`.
- **`internal/app/app.go`**: Added automatic version skew detection in `RunArgs`. When `InstalledVersion < Version` (and `Version != "dev"`), flags `PendingSync = true` to trigger transparent deferred sync.
- **`internal/doctor/doctor.go` & `internal/cli/doctor.go`**: Added `installed:version` diagnostic check (`CheckInstalledVersion`). Emits warning status when installed assets are older than the running binary and suggests `RemedySync` (`gentle-ai sync`).

## Test Coverage & TDD Verification

- Unit test suites executed and passed 100% clean across `internal/state`, `internal/cli`, `internal/doctor`, and `internal/app`.

## Summary by CodeRabbit

- **New Features**
  - Added an installation version check to diagnostics.
  - Diagnostics now identify mismatches between the installed asset and running application versions, with guidance to run `gentle-ai sync`.
  - Matching versions are reported as healthy.

- **Improvements**
  - Installation and synchronization operations now retain version information for more reliable validation.
  - Existing version information is preserved when installation state is merged.

Closes #3147

(#2685 was closed by the review-path guard; this PR delivers the doctor-side half tracked there.)


Alan-TheGentleman:
Heads up before you spend more time on this: #2685 shipped in #2759 (`e2e1674d`), and it overlaps yours in `state.go`, `run.go` and `sync.go`.

The shipped fix records `managed_asset_digest`, a content digest of the assets the binary embeds, and refuses review work when an installed digest disagrees. It deliberately does not key on a version or build identity, because those carry `vcs.revision` and ldflags, so byte-identical assets got declared outdated after any rebuild. Your `InstalledBinaryVersion` field has that same property, which is why I did not take it as-is.

**Your doctor check is still worth having, and I would like it.** The shipped refusal tells you at the moment you are blocked; a doctor warning tells you before you get there, which is strictly better for the user. Rebase it on the field that now exists: read `state.ManagedAssetDigest`, compare it to `assets.ManagedDigest()`, and warn on a mismatch. That drops your `state.go` and `run.go` and `sync.go` changes entirely, so what is left is just `doctor.go`, `internal/doctor/doctor.go` and the tests, which is a much smaller and cleaner PR than this one.

Please also link an approved issue when you re-push. #2685 is closed now, so this needs its own: "doctor should warn on managed asset skew" is clearly in scope and I will approve it.

Sorry the timing worked out this way. The overlap was not visible when you opened this.

Alan-TheGentleman (approved):
Approving and merging. The branch was 365 commits behind, so I did the reconciliation myself rather than round-tripping: both state fields coexist (`InstalledBinaryVersion` from this PR, `ManagedAssetDigest` + runtime provenance from main), and the version stamp now lives inside `persistSyncManagedAssetState` so the state file has one writer under one lock instead of two persist paths racing.

The check itself is the missing half of what shipped for #2685: the review path already refuses on skew at START preflight — verified working in the field — but the user finds out mid-review, after consent. `doctor` saying `installed:asset_version ... run 'gentle-ai sync'` up front is the difference between a health check answering the question and a dead end teaching it. The rc.7 occurrence on #2492 two days ago was exactly this: an upgraded binary reading rc.6-era instructions, and nothing anywhere said so.

Two test moves worth recording:

- The mocked doctor integration expectation gains the new check line and its 8th summary row — that was this PR's own CI failure, its check working while its fixture predated it.
- `TestRunSyncPreservesCompletePersistedState` keeps its byte-identical guarantee for every field except the one sync is now deliberately allowed to write, and it **asserts the stamp happened** rather than merely tolerating the drift — a preservation guard that silently accepts a new write is a guard that has stopped guarding.

Verified on the merged composition: root suite, bench with a cleared cache, `go vet` on linux/windows/darwin, and the deadcode ratchet. Thanks @Cobies — the check shape (skip honestly on legacy state, warn with the exact continuation) was right from your first push; everything I touched was drift, not design.

Issue #2609, opened 2026-08-05
bug(opencode): general worker completes with empty task result


fer8614:
### Pre-flight Checklist

- [x] I searched open and closed issues and found no tracker for this general-worker case.
- [x] I understand that PRs require maintainer approval on the linked issue.

### 📝 Bug Description

A general-purpose read-only worker launched through OpenCode can report `completed` while returning an empty `<task_result>`. Resuming the same worker once to request the missing report can again report `completed` with an empty result.

This is the same observable boundary as the SDD reports in #2117, but maintainers explicitly scoped that issue to SDD phases and requested a separate tracker for general workers.

### 🔄 Steps to Reproduce

1. Launch a `general` worker through OpenCode for read-only verification of a disposable local prototype, with an explicit structured textual return contract.
2. Let the worker reach the completed state.
3. Observe an empty `<task_result>` with no report or typed failure.
4. Resume the same worker once and explicitly request the missing report.
5. Observe another completed wrapper with an empty `<task_result>`.

### ✅ Expected Behavior

A completed general worker returns the requested non-empty verification report, or a typed incomplete/failure envelope describing why it could not report.

### ❌ Actual Behavior

Both bounded attempts completed with empty results. No verification status, command evidence, findings, mutation evidence, or typed failure reached the parent workflow.

### Gentle AI Version

2.2.4

### Operating System

Linux (Other), x86_64

### AI Agent / Client

OpenCode 1.18.13

### 📋 Affected Area

Other

### 💡 Logs / Error Output

```text
Attempt 1: state=completed, task_result=<empty>
Attempt 2 (same resumed worker): state=completed, task_result=<empty>
```

### Additional Context

**Operation shape**

The delegated actor was a read-only independent verifier for a disposable prototype. It was instructed not to edit the prototype or consumer repository and to return a structured PASS/FAIL report with command evidence. The second attempt reused the same worker context and requested only the missing report.

**Mutation and preserved-state outcome**

- The consumer repository remained clean before and after both attempts.
- No source mutation, staging, commit, push, pull request, review, or downstream workflow was attributed to the silent worker.
- The disposable prototype remains preserved.
- The consumer stopped fail-closed at the independent-verification boundary and did not infer PASS.

**Minimal reproduction**

1. Create any disposable local artifact that can be inspected read-only.
2. Launch a `general` worker with a structured textual result contract.
3. Observe `completed` with an empty result.
4. Resume that exact worker once to request the report.
[...]

PR #2750, opened 2026-08-08
fix(sdd): carry the provider cause in the task failure envelope

### LINKED ISSUE #2677: bug(sdd): sdd-spec task can return an empty provider result
### Observed behavior

An OpenCode SDD workflow reached the `sdd-spec` phase with a valid OpenSpec proposal, then the provider task returned no valid result. Gentle AI failed closed with `sdd_task_result_empty`.

This report records observed evidence only; it does not claim a root cause.

### Environment

- Gentle AI: `2.3.0-rc.2`
- OpenCode: `1.18.13`
- OS/architecture: `Linux x86_64`
- Artifact mode: OpenSpec
- Execution mode: interactive

### Operation shape and bounded outcome

- Precondition: one named change with `proposal: done` and `specs: missing`
- Operation: launch the dedicated `sdd-spec` phase through the OpenCode task provider
- Attempts: 1 phase launch; no automatic retry
- Provider outcome: no valid task result
- Mutation outcome: no spec artifacts were created; the existing proposal remained available
- Continuation: the provider-supplied status command was run exactly once
- Preserved state after continuation: `proposal: done`, `specs: missing`, `nextRecommended: spec`, no dispatcher blocker

### Failure envelope

```json
{"schemaName":"gentle-ai.sdd-task-result-failure/v1","status":"blocked","code":"sdd_task_result_empty","phase":"sdd-spec","summary":"sdd-spec returned no valid task result. Do not retry or advance SDD; inspect the existing artifact state and surface the terminal failure to the user.","continuation":"gentle-ai sdd-status --cwd '<repo>' --json"}
```

Safe reason identifier: `sdd_task_result_empty`.

### Expected behavior

The phase should return the documented SDD Result Contract and persist its declared specification artifacts, or return a non-empty structured failure that preserves an actionable continuation.

### Actual behavior

The task provider returned no valid phase result. Gentle AI emitted the typed terminal transport failure and correctly prevented retry and downstream phase advancement.

### Minimal reproduction

1. Initialize an OpenSpec-backed SDD project.
2. Create a named change with a valid `proposal.md` and no specification artifacts.
3. Dispatch `sdd-spec` through an already-running OpenCode session.
4. Observe that the phase returns no valid task result and Gentle AI emits `sdd_task_result_empty`.
5. Run the supplied sanitized continuation once and verify that the proposal remains done while specifications remain missing.

### Preserved-state evidence

The consumer workflow remains stopped at `spec`; no application code, DNS state, source content, or downstream SDD artifact was mutated.

Alan-TheGentleman:
Closes #2677

Refs #2609 (same class for general workers; this PR fixes the SDD half — the general-worker envelope needs its twin and stays open).

## Summary

Eight reports across six reporters, three SDD phases, two platforms and two artifact modes, all showing `sdd_task_result_empty`. The decisive evidence is the controlled two-arm experiment in the thread: identical fixtures, only the child model differing, and the failing arm's provider rejected the request with HTTP 403 **before generation**. Our wrapper replaced that with a summary claiming the phase "returned no valid task result" — the child never ran.

**What the hook can and cannot see, established empirically** (installed `@opencode-ai/plugin` types + OpenCode 1.18.14 task-tool source):

- The provider error text is **not available** at `tool.execute.after`: it lives on the child session record, reachable only via the `client` the transport reduction removed by design. Worse, when the child job status is `error` the after-hook never fires at all. No `cause` field is added for a fact that does not exist at this line.
- The child's **model route** (`providerID/modelID`) IS available in `output.metadata` and was being discarded. It is exactly the discriminating variable of the two-arm experiment, so it now travels as `taskModel` in the envelope.

**Two truths, two summaries.** The empty case now states the child produced no output at all and names the likeliest causes (provider rejection before generation: authentication, region, or model access; interruption; or a phase that genuinely wrote nothing). The malformed case keeps its existing wording byte-identical.

**Hostile values are omitted, not truncated**: `taskModel` is validated per token (provider and model separately); an absolute path or a provider dump never enters the transcript, proven by test.

**Explicitly not done**: retries, recovery-from-disk (a 403 leaves no artifact to recover), new states or flags, `client` reintroduction. Reviewer transport untouched — the diff is confined to the SDD block. Schema stays `v1`: every consumer tolerates the additive field (verified pin by pin); docs and goldens pin only codes and contract, none needed regeneration.

**Exit-0 note for the thread**: the outer process exiting 0 on failure is OpenCode host behavior, unreachable from the plugin; that half belongs upstream.

## Test Plan

```bash
go test ./internal/assets/... ./internal/components/... -count=1   # ok
go test ./internal/cli/ -count=1                                   # ok (210s)
cd bench && go test ./... -count=1                                 # ok
gofmt -l . && go vet ./... && go build ./...                       # clean
```

New/updated cases in `TestSDDTaskResultFailuresAreTerminalAndScoped`: truthful empty summary with `taskModel`; key absent without metadata; hostile path and 400-char provider dump proven absent from the handoff; malformed wording pinned unchanged. Reviewer independently decoy-verified the hostile-value guard (weakened the token regex; the test failed naming the case).

Review budget: 132 additions + 11 deletions = 143 changed lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of failed task results, including clear distinction between empty and malformed output.
  * Added clearer reporting when a provider or model rejects a task.
  * Preserved available task routing details in failure summaries while omitting missing or unsafe metadata.

### DIFF
--- a/internal/assets/assets_test.go
+++ b/internal/assets/assets_test.go
@@ -519,6 +519,14 @@ func TestReviewResultArtifactsPluginContract(t *testing.T) {
 		`"sdd_task_result_malformed"`,
 		`failedSDDSessions`,
 		`extractionClass(cause, "sddClass")`,
+		// #2677: an empty result means the child produced no output at all
+		// (for example a provider rejection before generation), and the
+		// handoff must say so and carry the one causal fact the hook
+		// receives -- the child's provider/model route -- after validation.
+		`function taskRouteModel(`,
+		`produced no task output at all`,
+		`provider rejected the request before generation (authentication, region, or model access)`,
+		`taskRouteModel(metadata)`,
 		`export default ReviewResultArtifactsPlugin`,
 	} {
 		if !strings.Contains(source, want) {
--- a/internal/assets/opencode/plugins/review-result-artifacts.ts
+++ b/internal/assets/opencode/plugins/review-result-artifacts.ts
@@ -132,9 +132,55 @@ type SDDTaskFailureError = Error & { sddFailure: SDDTaskFailure }
 function shellQuote(value: string): string {
   return `'${value.replace(/'/g, "'\\''")}'`
 }
-function sddTaskFailure(phase: string, cwd: string, cause: unknown): SDDTaskFailureError {
+// SDD_TASK_ROUTE_TOKEN bounds one provider or model identifier before it may
+// enter the failure handoff. The hook cannot see the child session's own
+// provider failure -- a pre-inference rejection (for example an HTTP 403
+// region refusal, #2677) is recorded on the child session record, which this
+// plugin deliberately cannot query since the client argument was removed --
+// so the child's model route from the task result metadata is the one causal
+// fact available at this boundary. A value is carried only when it looks
+// like a plain route identifier; anything with separators, whitespace, or
+// path shapes is omitted entirely rather than truncated, so hostile or
+// accidental metadata (absolute paths, provider dumps) never reaches the
+// session transcript.
+const SDD_TASK_ROUTE_TOKEN = /^[A-Za-z0-9][A-Za-z0-9._:@-]{0,127}$/
+
+// taskRouteModel extracts the child task's provider/model route from the
+// task tool's result metadata ({parentSessionId, sessionId, model:
+// {providerID, modelID}}), or undefined when the metadata does not carry a
+// valid route. Absence is tolerated, never invented: the handoff simply
+// omits the field.
+function taskRouteModel(metadata: unknown): string | undefined {
+  if (!metadata || typeof metadata !== "object" || Array.isArray(metadata)) return undefined
+  const model = (metadata as Record<string, unknown>).model
+  if (!model || typeof model !== "object" || Array.isArray(model)) return undefined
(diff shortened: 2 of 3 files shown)


dnlrsls:
Thanks, the envelope improvement is useful, but our sanitized occurrence confirms an empty-result path that this PR does not appear to classify.

The `sdd-spec` child did run on `opencode-go/qwen3.8-max`. Two assistant turns completed with `finish: tool-calls`; the final turn ended with `finish: length` after 32,002 output tokens and emitted no textual result. The child recorded no provider error, executed read-only tools only, and never persisted the specification. The parent reduced that available terminal signal to `sdd_task_result_empty`.

Adding `taskModel` helps identify the route, but the same occurrence would still be indistinguishable from authentication, region, interruption, or a phase that genuinely wrote nothing. Could we either preserve a sanitized terminal finish classification such as `length` when that signal is available at the integration boundary, or track this case explicitly in a follow-up? A focused regression could model a child ending with `finish: length` and no result, then assert that the caller receives the causal classification while the existing proposal remains preserved and downstream phases stay blocked.

PR #2751, opened 2026-08-08
fix(review): prove Claude advisory transport to a terminal receipt

decode2:
## 🔗 Linked Issue

Fixes #2757

Refs #2692
Refs #2566

This PR proves the deterministic advisory transport and native delivery lifecycle. #2692 and #2566 remain open until a real Anthropic-backed Claude run proves the model judgment itself.

---

## 🏷️ PR Type

- [x] `type:bug` — Bug fix (non-breaking change that fixes an issue)
- [ ] `type:feature` — New feature (non-breaking change that adds functionality)
- [ ] `type:docs` — Documentation only
- [ ] `type:refactor` — Code refactoring (no functional changes)
- [ ] `type:chore` — Build, CI, or tooling changes
- [ ] `type:breaking-change` — Breaking change (fix or feature that changes existing behavior)

---

## 📝 Summary

- Proves the pinned Claude Code runtime can carry a scripted shell-free advisory result through native capture, terminal receipt, and one delivery gate.
- Extends the existing pinned Claude network-none lane: genuine Claude CLI, loopback Anthropic-compatible fixture, fake key, zero external network, and no account or paid API access.
- Adds a replayable issue-linked benchmark for the terminal delivery outcome.

This supersedes only the #2692 `tools: [Bash]` slice proposed in #2721. Claude must not regain shell access; #2721's unrelated fixes remain independent. The loopback fixture validates plumbing, not the judgment of a real Claude model.

---

## 📂 Changes

| File / Area | What Changed |
|-------------|-------------|
| `internal/advisoryreview/claude_adapter.go` | Adds the minimal Claude invocation adapter: empty scratch directory, isolated environment, zero tools, prompt via stdin, raw final bytes only. |
| `internal/advisoryreview/claude_adapter_test.go` | Verifies invocation shape, environment isolation, error envelopes, cancellation, and transport failures. |
| `e2e/organicruntime/claude_reviewer_e2e_test.go` | Drives negotiated STATUS/START/collect, real pinned Claude runtime, malformed/mismatch refusals, native capture, receipt finalization, and gate allow. |
| `e2e/Dockerfile.claude-network-none` | Includes the organic runtime package in the existing pinned network-none proof image. |
| `bench/journeys_advisory.go` | Adds `j76-claude-advisory-result-reaches-delivery`, linked to #2692. |
| `bench/journeys_sdd_test.go` | Updates the journey corpus count pin. |

---

## 🧪 Test Plan

**Focused Go tests**

```bash
[...]

PR #2759, opened 2026-08-08
fix(review): detect managed asset version skew

### LINKED ISSUE #2685: fix(install): nothing detects installed assets older than the running binary

Alan-TheGentleman:
## 🔗 Linked Issue

Closes #2685

---

## 🏷️ PR Type

---

## 📝 Summary

- Persist the exact binary identity that successfully wrote the complete managed asset set.
- Fail review lifecycle mutations and valid receipt-backed delivery closed when managed provenance is missing or mismatched, while preserving recovery and native corrupt/non-approval denials.
- Require `gentle-ai sync` after binary upgrades in install, usage, and release guidance.

---

## 📂 Changes

|-------------|-------------|
| `internal/state`, install, sync | Persist writer identity only after complete successful asset refresh |
| Review lifecycle boundaries | Gate start, capture, abandon, imports, and valid v3/compact/legacy receipt-backed delivery |
| Tests and harnesses | Prove bidirectional skew, partial failure, bypass closure, receipt ordering, and modern sandbox provenance |
| `README.md`, `docs/usage.md`, `docs/release-signing.md` | Require asset refresh after binary replacement or upgrade |

---

## 🧪 Test Plan

**Broad verification completed on the initial implementation:**

- [x] `go test ./... -count=1`
- [x] `go vet ./...`
- [x] `./scripts/deadcode-ratchet.sh`
- [x] Bench module tests and vet
- [x] CI-equivalent driven benchmark: `76 completed, 0 unsupported, 0 failed`
- [x] Source-coupled CI benchmark modes

**Focused verification after bounded corrections:**

- [x] Managed provenance gate, lifecycle-boundary, and partial-sync tests
- [x] v3 in-flight and approved-without-receipt ordering tests
- [x] Compact/legacy receipt ordering test, including corrupt, escalated, and valid approved authority
- [x] `git diff --check`

The broad suites were not redundantly rerun after each bounded correction; CI will rerun the complete candidate.

- [ ] Docker E2E passes in CI
- [x] Manually exercised the isolated managed-asset runtime and driven benchmark paths

---

## 🤖 Automated Checks

CI owns the final complete-candidate verification.

---

## ✅ Contributor Checklist

- [x] Maintainer approved `size:exception`: 419 changed lines, where the 19-line excess is the documentation required by #2685
- [x] Unit tests passed on the implementation and focused corrections passed
- [x] Go format and diff checks pass
- [ ] Docker E2E passes in CI
- [x] Benchmark validation completed
- [x] Updated required installation, upgrade, and release guidance
- [x] Commit follows Conventional Commits
- [x] Commit has no `Co-Authored-By` trailer

---

## 💬 Notes for Reviewers

The provenance guard applies only where managed reviewer assets can create, mutate, or authorize review authority. Help, mode, status, capabilities, sync, receiptless states, corrupt authority, and unrelated commands retain their existing behavior. Compact and legacy receipts are parsed and evaluated before provenance is checked, so stale provenance cannot hide their more specific corrupt or escalated denials.

This PR intentionally uses the maintainer-approved `size:exception`. The production-and-test work unit is exactly 400 changed lines; the remaining 19 lines complete the issue-required user and release documentation.

## Summary by CodeRabbit

- **New Features**
  - Added managed asset provenance verification for review and release operations.
  - Installation and synchronization now record managed asset identity.
  - Added safe locking and state merging for concurrent configuration updates.

- **Bug Fixes**
  - Operations requiring trusted assets now stop when assets are stale or mismatched and provide recovery guidance.
  - Synchronization more reliably preserves configuration and provenance.

- **Documentation**
  - Clarified that `gentle-ai sync` is required after replacing or upgrading the binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ go install github.com/gentleman-programming/gentle-ai/v2/cmd/gentle-ai@latest
 > [!WARNING]
 > Windows source builds and CI/runtime tests remain supported, but official Windows binary distribution and Scoop are temporarily unavailable. Windows installation and upgrades require Go 1.25.10+ and fail closed to source-install guidance; they never download an unsigned Gentle AI executable or execute a remote update script.
 
+> [!IMPORTANT]
+> After replacing or upgrading the `gentle-ai` binary, run `gentle-ai sync` to refresh its managed assets. See the [sync and upgrade reference](docs/usage.md#sync).
+
 ### Configure project context
 
 Once your agents are configured, open your AI agent in a project and run these two commands to register the project context:
--- a/bench/journeys_advisory.go
+++ b/bench/journeys_advisory.go
@@ -344,5 +344,22 @@ func advisoryJourneys() []Journey {
 				{Name: "finish the review", Requires: captureResultCapability, Composite: finishAdvisoryReview},
 			},
 		},
+		{
+			ID:     "j76-claude-advisory-result-reaches-delivery",
+			Title:  "Claude advisory result: provider-bound review reaches terminal delivery without reviewer tools",
+			Source: "https://github.com/Gentleman-Programming/gentle-ai/issues/2692",
+			Steps: []Step{
+				{Name: "fixture: repo", Fixture: baseRepo},
+				{Name: "fixture: stage auth code", Fixture: stageAuthCode},
+				{Name: "review start (negotiated v2, consent granted)", Requires: startContractCapability, Composite: startAdvisoryReview},
+				{Name: "bind the first pending lens", Requires: statusCapability, Composite: captureAdvisoryBinding},
+				{Name: "render the canonical Claude advisory prompt", Requires: advisoryPromptCapability,
+					Args: advisoryPromptArgs("claude-code"), After: rememberAdvisoryPrompt("issue-2692-claude-prompt")},
+				{Name: "fixture: write the provider-bound reviewer result", Fixture: writeAdvisoryResult("issue-2692-result", "")},
+				{Name: "admit the provider-bound reviewer result", Requires: advisoryValidateCapability,
+					Args: advisoryValidateArgs("issue-2692-result"), After: assertAdvisoryValidateAdmitted},
+				{Name: "capture, finalize, and allow delivery", Requires: captureResultCapability, Composite: finishAdvisoryReview},
+			},
+		},
 	}
 }
--- a/bench/journeys_sdd_test.go
+++ b/bench/journeys_sdd_test.go
@@ -34,8 +34,11 @@ func TestPortableSDDFailClosedAuthorityJourneysAreRegistered(t *testing.T) {
 			want[journey.ID] = true
 		}
 	}
-	if got := len(seen); got != 76 {
-		t.Errorf("core journey count = %d, want 76", got)
+	// 77 since j76-claude-advisory-result-reaches-delivery (#2692, #2566).
+	// Bump this deliberately when a journey is added, and name it here: the
+	// count exists so a journey cannot appear or vanish unnoticed.
+	if got := len(seen); got != 77 {
+		t.Errorf("core journey count = %d, want 77", got)
 	}
 	for id, found := range want {
 		if !found {
--- a/bench/runner.go
+++ b/bench/runner.go
@@ -73,27 +73,6 @@ func newSandbox(binary, root string) (*Sandbox, error) {
 	return sandbox, nil
 }
 
-func (s *Sandbox) stampManagedAssetWriter() error {
-	command := exec.Command(s.Binary, "review", "capabilities")
-	command.Env = s.env()
-	payload, err := command.CombinedOutput()
-	if err != nil {
-		if strings.Contains(string(payload), "unknown review command") {
-			return nil
-		}
-		return fmt.Errorf("read bench binary build identity: %w: %s", err, payload)
-	}
-	var capabilities struct {
-		Build struct {
-			ID string `json:"id"`
-		} `json:"build"`
-	}
-	if err := json.Unmarshal(payload, &capabilities); err != nil || capabilities.Build.ID == "" {
-		return fmt.Errorf("decode bench binary build identity: %s", payload)
-	}
-	return s.write(filepath.Join(s.Home, ".gentle-ai", "state.json"), fmt.Sprintf(`{"managed_asset_writer":%q}`, capabilities.Build.ID))
-}
-
 // env is a closed environment: only what the product legitimately needs.
 // PATH is inherited because the product shells out to git.
 func (s *Sandbox) env() []string {
@@ -520,11 +499,6 @@ func runJourney(binary string, journey Journey) JourneyResult {
 		return result
 	}
 	sandbox.NewLineageActivation = journey.NewLineageActivation
-	if err := sandbox.stampManagedAssetWriter(); err != nil {
-		result.Status = StatusFailed
-		result.FailureReason = err.Error()
-		return result
-	}
 	accumulator := newAccumulator()
 	probe := newCapabilityProbe(sandbox)
 	run := &journeyRun{sandbox: sandbox, probe: probe, accumulator: accumulator}
--- a/bench/runner.go
+++ b/bench/runner.go
@@ -73,6 +73,27 @@ func newSandbox(binary, root string) (*Sandbox, error) {
 	return sandbox, nil
 }
 
+func (s *Sandbox) stampManagedAssetWriter() error {
+	command := exec.Command(s.Binary, "review", "capabilities")
+	command.Env = s.env()
+	payload, err := command.CombinedOutput()
+	if err != nil {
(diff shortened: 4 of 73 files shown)


Alan-TheGentleman:
Final read validation done, plus a merge of `origin/main`. One real finding, one readability change.

**The finding is a circular dead end this PR introduced.** `authorizeManagedReviewerAssets` refuses every review entry point until sync records the writer, and the new sync path returns a bare error whenever the state file cannot be read. A corrupt `state.json` therefore left no way out: review said to run sync, and sync failed without naming anything runnable. Fixed by naming `gentle-ai install`, which rebuilds the file whole and stamps the writer at `run.go:238`, so the refusal now points at a command that actually resolves it.

**The readability change** flattens the operation gate in `resolveReviewOperationRoot` into a switch. The `else if` form combined an assignment, a nil check, and a second operation test in one condition, and the mutate arm's asset gate was hidden one call deeper. Behavior is identical: mutate and abandon get the asset gate, read-only operations do not.

Verified the gate coverage rather than assuming it: `review_artifact.go`'s non-opaque branch reaches the gate through `resolveReviewMutationRoot`, and `receiptBacked` is set only on the approved-with-loadable-receipt path, so an invalidated or escalated evaluation still reports its own reason instead of the asset refusal.

Both modules build, and `internal/cli` passes for the asset-provenance, sync, kill-switch, and review-mode paths. Merging once CI is green.

Alan-TheGentleman:
CI went red, and the diagnosis matters more than the two failures. Both are root-caused; one of them says the current comparison is the wrong one.

**Failure 1, a test bug.** `stampReviewTestAssets` called `t.Setenv` from inside `initReviewCLIRepo`, and parallel tests call that. Go panics with "test using t.Setenv or t.Chdir can not use t.Parallel", which is what killed `TestReviewFacadeCorrectionFlowResumesFromEachCompactIntermediateState` in 0.00s and took the whole `internal/cli` package down on both lanes. Fixed by deleting the helper and stamping the package home once in the TestMain that already exists at `protocol_probe_test.go`, which also covers the spawned binary because the child inherits `HOME` and `USERPROFILE`. A per-test helper cannot reach a child process.

**Failure 2 is the real finding.** The acceptance tests refuse with "managed reviewer assets are outdated". It reproduces on Linux once `GENTLE_AI_TEST_BINARY` is set, which is why it looked Windows-only: the Linux unit lane skips those tests for lack of a binary. The cause is that those tests mix in-process calls, which compute the test binary's build identity, with spawned real-binary calls, which compute the product binary's, against one `state.json`. Whichever stamps it, the other is refused.

That is not a test artifact. It is the same defect users would hit: because the gate compares a **build** identity, any rebuild with a different `vcs.revision` declares byte-identical assets outdated and demands a needless `gentle-ai sync`. Every dev build and every asset-free patch release trips it.

**The correct comparison is a digest of the embedded managed assets**, not of the binary that compiled them. Two binaries built from the same source then agree, which is precisely why the mixed test would pass under it. I have `internal/assets` digest plus the state-field rename drafted.

**Two things block finishing it, and both need your call.** Exposing that digest through `review capabilities` collides with the byte-exact conformance fixtures and `additionalProperties: false`, so it needs a protocol minor bump with a new schema and fixture pair, on the same surface whose last pin bump broke every negotiated START. And `bench` cannot compute the digest itself: it is module `github.com/gentleman-programming/gentle-ai/bench`, outside the `/v2` prefix, with no require or replace, so it cannot import `internal/assets` either. Out-of-process stamping has no cheap path today.

Not merging this as it stands. Holding for your decision on how the digest reaches out-of-process callers.

Alan-TheGentleman:
Correcting my previous comment on one point, and narrowing the blocker considerably.

**The mismatch is structural, not incidental.** A test binary and a product binary can never agree on a build identity: `go test` does not stamp VCS build settings, so `debug.ReadBuildInfo()` in-process has no `vcs.revision` while the built binary does, and CI adds a version via ldflags on top of that. Any gate keyed on build identity therefore fails permanently in any test that mixes in-process calls with spawned-binary calls. That is not a test smell to work around, it is the design telling us the wrong thing is being compared.

**The protocol bump is mostly avoidable.** The asset digest is a property of the source, not of the binary, so every in-repo consumer can compute it directly from `internal/assets` with no wire change at all: `review_binary_acceptance_test.go`, `gate_boundary_matrix_test.go`, and `e2e/organicruntime` all import it today. Only `bench` cannot, because its module path sits outside the `/v2` prefix and Go's internal rule blocks it regardless of any replace directive. So the open question is one consumer, not the contract.

**Test surface fixed and verified.** `go test ./internal/cli -count=1` passes in full on Linux (136s), including your in-flight install-state lock work. The parallel panic is gone, and the fix was not the one-time TestMain stamp I described before: the package home is shared, so a test that rewrites `state.json` without the writer refuses every review after it. Per-test stamping is restored, now read-modify-write so it preserves the install and sync fixtures, and with no `t.Setenv`, which is what made it panic under `t.Parallel`. TestMain keeps its stamp because spawned children inherit that env and no per-test helper can reach them.

What remains red is exactly the mixed-identity acceptance suite, which the digest change resolves. Left uncommitted in the worktree since you are working in it.

Alan-TheGentleman:
Finished, with the design changed to what the CI failure was actually pointing at.

**Provenance is now bound to asset content, not to the build.** The capabilities build identity carries `vcs.revision`, module version and ldflags, so it declared byte-identical assets outdated after any rebuild, and it made agreement impossible between a test binary, which Go does not stamp with VCS settings, and a released one. `internal/assets.ManagedDigest()` hashes the embedded assets with sorted paths, so two binaries built from the same source agree by construction.

**The refusal now fires only on a recorded digest that disagrees.** A home with no state file, or with no digest recorded, never installed managed assets, so there is nothing stale and nothing the refusal could tell the caller to repair. The previous fail-closed shape would have blocked every user who installed with `go install` and never ran `gentle-ai install`, while naming a sync that fixes nothing. That is the population most likely to hit it and least able to diagnose it.

That one condition is what deleted the scaffolding: the stamping helpers in `bench`, the organic e2e, the gate boundary matrix, and the CLI test fixtures all existed only to satisfy a gate that refused an uninstalled home. Net for the whole PR is 162 added against 161 deleted.

Also folded in and kept: install and sync now serialise their state writes under one lock and re-read inside it, which closes a genuine read-modify-write window where a long sync wrote back state it had read before doing its work. And the delivery gate is scoped to an allowed evaluation, so a denied receipt-backed gate reports its own reason instead of being masked by the asset refusal. That second one was a real bug in my earlier commit, caught by the test that now pins it.

Verification: `go vet ./...` clean, `gofmt` clean, `go test ./... -count=1` green at 66 packages with `GENTLE_AI_TEST_BINARY` set, which is the exact condition that was red on Windows. Decoy run: neutering the comparison produced 10 failures across every gated entry point, including the receipt-ordering cases and the post-apply masking test, then restored to green.

Ready to merge once CI confirms.

Alan-TheGentleman:
One unrelated commit rode along, and it is mine to own. `Unit Tests` was failing on the bench corpus ratchet, and it fails on `origin/main` too: `2a0533ae` added `j76-claude-advisory-result-reaches-delivery` without bumping the expected count, so the lane has been red since I merged #2751 this morning. Bumped 76 to 77 and named the journey in the constant so the next bump has to justify itself.

I missed it here because I verified the main module's suite and only built the bench module rather than testing it. bench is a separate Go module with its own tests, so `go test ./...` at the root never reaches it.

PR #2342, opened 2026-08-03
fix(sdd): emit PowerShell-safe UserPromptSubmit hook on Windows

ardelperal:
## Linked Issue
Closes #2124

## PR Type
- [x] `type:bug` — Bug fix (non-breaking change that fixes an issue)

## Summary
Claude Code on Windows executes hook commands through Windows PowerShell 5.1 (the default on Windows 11), which does not support bash-only syntax like `||` as a statement separator. The pre-fix literal written by `gentle-ai sync` failed at parse time, so the skill-registry refresh never executed on Windows hosts even though the failure was swallowed as non-blocking.

This PR makes the Claude `UserPromptSubmit` hook command literal platform-aware:

- Windows: a PowerShell 5.1-safe wrapper using `$env:CLAUDE_PROJECT_DIR` and explicit `exit 0` instead of `${VAR:-default}` and `|| true`.
- Linux/macOS: the existing POSIX literal, unchanged.

On Windows, a `settings.json` that still holds the pre-fix POSIX literal is migrated in place during the same sync call: the legacy entry is removed before the canonical PowerShell entry is appended, so no duplicate `UserPromptSubmit` entries are produced. The migration is a no-op on non-Windows hosts, where the canonical literal is byte-identical to the pre-fix one.

The change is scoped to `ensureClaudeSkillRegistryHook` only. `ensureCodexSkillRegistryHook` is left untouched because issue #2124 is about the Claude `UserPromptSubmit` contract specifically (Codex uses a different `commandWindows` shape that is out of scope for this PR).

## Changes
| File / Area | What Changed |
|-------------|--------------|
| `internal/components/sdd/inject.go` | `ensureClaudeSkillRegistryHook` selects the hook command literal based on `runtime.GOOS`; on Windows, an inline migration removes the pre-fix POSIX literal from `hooks.UserPromptSubmit` before appending the canonical PowerShell entry. |
| `internal/components/sdd/inject_test.go` | Two new tests: `TestEnsureClaudeSkillRegistryHookWritesPlatformAwareCommand` (verifies the written literal matches the GOOS-aware expectation) and `TestEnsureClaudeSkillRegistryHookReplacesLegacyPOSIXCommand` (Windows-only, verifies legacy POSIX is removed and replaced by the canonical PowerShell literal). |

## AI Assistance
- [x] **Material assistance used** — Complete all applicable declaration fields below.

[...]

PR #2807, opened 2026-08-08
fix(hooks): make skill-registry hooks PowerShell-safe on Windows

### LINKED ISSUE #2124: bug(hooks): UserPromptSubmit hook command uses bash `|| true`, fails under Windows PowerShell 5.1
### Pre-flight Checklist

### Bug Description
The `UserPromptSubmit` hook that `gentle-ai sync` writes into the user's Claude Code `settings.json` runs `gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "${CLAUDE_PROJECT_DIR:-$PWD}" || true`. This is bash/POSIX syntax. On Windows, Claude Code executes hook commands through Windows PowerShell, and PowerShell 5.1 (the version bundled with Windows 11 by default) does not support `||`/`&&` as statement separators — that was only added in PowerShell 7+. The hook fails with a parser error before `gentle-ai` ever runs, so `skill-registry refresh` never executes on affected Windows machines, even though the failure is swallowed as non-blocking.

### Steps to Reproduce
1. On Windows 11 with the default Windows PowerShell 5.1 (not PowerShell 7+), install/sync Gentle AI so it writes the `UserPromptSubmit` hook into `~/.claude/settings.json`.
2. Start a Claude Code session in a project with that hook configured.
3. Submit any prompt (triggers `UserPromptSubmit`).
4. Observe the hook's stderr.

### Expected Behavior
The hook command is valid in the shell Claude Code actually invokes it with on Windows, and `gentle-ai skill-registry refresh` runs (or fails for its own reasons, not a syntax error).

### Actual Behavior
The hook fails every time with a PowerShell parser error, and `gentle-ai skill-registry refresh` never executes:
```
Failed with non-blocking status code: En línea: 1 Carácter: 93
+ ... sh --quiet --no-gitignore --cwd "${CLAUDE_PROJECT_DIR:-$PWD}" || true
+                                                                   ~~
El token '||' no es un separador de instrucciones válido en esta versión.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : InvalidEndOfLine
```
exitCode: 1, stdout empty, ~500ms. Confirmed reproducible: running the same command directly under PowerShell 5.1 throws the identical parser error; running the underlying `gentle-ai skill-registry refresh` invocation directly under bash (Git Bash) succeeds with exit 0 and no output, confirming the refresh command itself is fine — only the `|| true` suffix (and possibly the `${VAR:-default}` bash parameter expansion) is the problem on Windows PowerShell.

### Environment
- Gentle AI version: 2.2.4
- Operating system: Windows 11
- AI agent / client: Claude Code 2.1.220
- Shell: Windows PowerShell 5.1 (5.1.26100.8875) — the default hook shell on Windows
- Affected area: hooks (UserPromptSubmit, `skill-registry refresh` sync)

### Root Cause
The hook command template that gentle-ai writes for `UserPromptSubmit` is authored in bash/POSIX syntax (`${VAR:-default}` parameter expansion, `|| true` as a statement separator). Windows PowerShell 5.1, which Claude Code uses to run hook commands on Windows, doesn't support either construct the same way — `||`/`&&` are PowerShell 7+ only. The template needs a PowerShell-safe equivalent when the target shell is Windows PowerShell, e.g.:
```powershell
if ($env:CLAUDE_PROJECT_DIR) { $d = $env:CLAUDE_PROJECT_DIR } else { $d = $PWD }; gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "$d"; exit 0
```
instead of the bash `<command> || true` form.

### Local workaround
Manually rewrote the hook command in `settings.json` to the native PowerShell form above; confirmed it now runs cleanly (exit 0) under PowerShell 5.1.

jjeg1979:
## 🔗 Linked Issue

Closes #2124

---

## 🏷️ PR Type

---

## 📝 Summary

The skill-registry hook that gentle-ai writes for Claude Code and Codex used bash syntax (`|| true` and `${CLAUDE_PROJECT_DIR:-$PWD}`) that Windows PowerShell 5.1 rejects with a parser error. On Windows without Git Bash, both agents run shell-form hooks through PowerShell, so the hook failed before gentle-ai ever ran, and `skill-registry refresh` never executed.

Each agent uses its own contract for a Windows-safe command:

- **Claude Code** command hooks accept a `shell: "bash" | "powershell"` field. On Windows we emit a PowerShell-safe command (using `$env:CLAUDE_PROJECT_DIR` and `exit 0`) tagged with `"shell": "powershell"`, so Claude Code runs it through PowerShell deterministically regardless of Git Bash presence. Other platforms keep the bash command.
- **Codex** hooks use a `commandWindows` field (a Windows-only command override). We keep the bash `command` and add a PowerShell-safe `commandWindows`.

An already-installed bash-only hook is migrated (replaced, not duplicated) on the next sync, so users who already have the old hook don't end up with two entries. `runtime.GOOS` is injected via a package-level var so the Windows and Unix shapes are unit-testable without build tags (mirroring the `runtimeGOOS` pattern in `internal/components/filemerge`).

---

## 📂 Changes

|-------------|-------------|
| `internal/components/sdd/inject.go` | Per-platform hook command generation for Claude (`shell` field) and Codex (`commandWindows`); new `pruneStaleSkillRegistryHooks` migrates/dedupes old hook variants; injectable `skillRegistryHookGOOS`; removed now-unused `claudeHookExists`/`claudeHookListContains`. |
| `internal/components/sdd/inject_test.go` | New tests for Windows PowerShell shape, Unix bash shape, and bash→PowerShell migration/idempotency; updated the existing Codex test for the `command` + `commandWindows` shape. |

---

## 🧪 Test Plan

**Unit Tests**

Ran the affected packages (the only two that reference these hooks):

```bash
go test ./internal/components/sdd/ ./internal/components/uninstall/   # ok
go build ./...                                                        # ok
go run ./internal/gofmtcheck                                          # ok (exit 0)
```

The full `go test ./...` exceeds a local 10-minute window on this machine (e2e-heavy), so I relied on CI for the complete workspace run. I verified the two packages that reference `installSkillRegistryAutomation` / the hook functions.

**Benchmark Validation**

N/A — this change does not touch review-lifecycle, gates, recovery, delivery, or the benchmark corpus/classifier. It only changes how the skill-registry hook command string is generated per platform.

- [x] Unit tests pass for affected packages (`go test ./internal/components/sdd/ ./internal/components/uninstall/`)
- [ ] E2E tests pass (`cd e2e && ./docker-test.sh`) — not run locally; deferring to CI
- [ ] Manually tested on Windows PowerShell 5.1 — I'm on macOS and can't run PS 5.1 end-to-end (see the question in the issue thread). Generation logic is covered by `runtime.GOOS`-injected unit tests.

---

## ✅ Contributor Checklist

- [x] PR stays within 400 changed lines (286 changed)
- [x] Unit tests pass for affected packages
- [ ] E2E tests pass — deferring to CI
- [x] Benchmark validation N/A (explained above)

---

## 💬 Notes for Reviewers

I raised two design questions in the issue thread (issuecomment-5225409015) before implementing and went with the explicit per-platform approach (`shell` for Claude, `commandWindows` for Codex). Happy to switch to a shell-neutral command if you'd prefer that.

Two things worth a close look:
1. The migration path in `pruneStaleSkillRegistryHooks` — it removes any prior gentle-ai skill-registry hook variant (matched by the `gentle-ai skill-registry refresh` marker) whose command differs from the current-platform command, so a user upgrading from the bash-only hook gets it replaced rather than duplicated.
2. I can't validate PowerShell 5.1 end-to-end on macOS. If you have a Windows environment, a manual check that the generated hook actually runs under PS 5.1 would be valuable before merge.

This change is in `internal/components/sdd`, not `internal/cli` / `internal/reviewtransaction` / `internal/sddstatus`, so the guard-population reviewer items don't apply.

## Summary by CodeRabbit

* **New Features**
  * Skill-registry hooks now support platform-specific Bash and PowerShell commands.
  * Windows environments use PowerShell commands where appropriate.
* **Bug Fixes**
  * Prevented duplicate hooks during installation and synchronization.
  * Automatically migrates outdated hook variants.
  * Removes obsolete managed hooks and empty matcher groups.
  * Improved consistent hook behavior across repeated synchronizations.

### DIFF
--- a/internal/components/sdd/inject.go
+++ b/internal/components/sdd/inject.go
@@ -6,6 +6,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/gentleman-programming/gentle-ai/v2/internal/agents"
@@ -1477,6 +1478,43 @@ func readMisnamedOpenCodeGentlemanSDDPrompt(settingsPath string) (string, error)
 	return prompt, nil
 }
 
+// skillRegistryHookGOOS is injected so unit tests can exercise the Windows and
+// non-Windows hook command shapes without build tags, mirroring the
+// runtimeGOOS pattern in internal/components/filemerge (issue #2124).
+var skillRegistryHookGOOS = func() string { return runtime.GOOS }
+
+// skillRegistryHookMarker uniquely identifies a gentle-ai skill-registry hook
+// regardless of the exact shell command variant. It is used for idempotency and
+// migration: an already-installed hook (bash or PowerShell) is detected and
+// replaced instead of being duplicated on the next sync (issue #2124).
+const skillRegistryHookMarker = "gentle-ai skill-registry refresh"
+
+// Bash command variants (macOS, Linux, and Windows with Git Bash). The trailing
+// `|| true` keeps the hook non-blocking when the refresh exits non-zero.
+const (
+	claudeSkillRegistryBashCommand = `gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "${CLAUDE_PROJECT_DIR:-$PWD}" || true`
+	codexSkillRegistryBashCommand  = `gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "$PWD" || true`
+)
+
+// PowerShell-safe command variants for Windows without Git Bash, where Claude
+// Code and Codex run the hook through PowerShell 5.1. PowerShell 5.1 rejects the
+// bash `|| true` separator and `${VAR:-default}` expansion, so we branch on the
+// project-dir env var explicitly and swallow non-zero exits with `exit 0`. The
+// `$env:` form is required; the bare `$CLAUDE_PROJECT_DIR` resolves to $null in
+// PowerShell.
+const (
+	claudeSkillRegistryPowerShellCommand = `if ($env:CLAUDE_PROJECT_DIR) { $d = $env:CLAUDE_PROJECT_DIR } else { $d = $PWD }; gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "$d"; exit 0`
+	codexSkillRegistryPowerShellCommand  = `gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "$PWD"; exit 0`
+)
+
+// isSkillRegistryHookCommand reports whether a hook command string is a
+// gentle-ai skill-registry hook in any shell variant. Used to migrate a
+// previously installed hook to the current platform's command without
+// duplicating it.
+func isSkillRegistryHookCommand(command string) bool {
+	return strings.Contains(command, skillRegistryHookMarker)
+}
+
 func installSkillRegistryAutomation(homeDir string, adapter agents.Adapter) (InjectionResult, error) {
 	if adapter.Agent() == model.AgentCodex {
 		hooksPath := filepath.Join(adapter.GlobalConfigDir(homeDir), "hooks.json")
@@ -1510,11 +1548,6 @@ func ensureCodexSkillRegistryHook(hooksPath string) (bool, error) {
 		return false, err
 	}
 
-	const command = `gentle-ai skill-registry refresh --quiet --no-gitignore --cwd "$PWD" || true`
-	if claudeHookExists(root, command) {
-		return false, nil
-	}
-
 	hooksRaw, hasHooks := root["hooks"]
 	hooksMap, _ := hooksRaw.(map[string]any)
 	if hasHooks && hooksMap == nil {
@@ -1529,14 +1562,24 @@ func ensureCodexSkillRegistryHook(hooksPath string) (bool, error) {
 	if hasSessionStart && sessionStart == nil {
 		return false, fmt.Errorf("Codex hooks %q has unsupported hooks.SessionStart shape: want array", hooksPath)
 	}
+
+	// Codex runs command hooks through PowerShell on Windows, where the bash
+	// `command` fails. Codex exposes a Windows-only `commandWindows` override, so
+	// we ship the bash command plus a PowerShell-safe override and let Codex pick
+	// the right one per platform (issue #2124).
+	sessionStart, hasCurrent, migrated := pruneStaleSkillRegistryHooks(sessionStart, codexSkillRegistryBashCommand)
+	if hasCurrent && !migrated {
+		return false, nil
+	}
 	sessionStart = append(sessionStart, map[string]any{
 		"matcher": "startup|resume|clear|compact",
 		"hooks": []any{
 			map[string]any{
-				"type":          "command",
-				"command":       command,
-				"timeout":       30,
-				"statusMessage": "Refreshing skill registry",
+				"type":           "command",
+				"command":        codexSkillRegistryBashCommand,
+				"commandWindows": codexSkillRegistryPowerShellCommand,
+				"timeout":        30,
+				"statusMessage":  "Refreshing skill registry",
 			},
 		},
 	})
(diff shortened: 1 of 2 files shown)


jjeg1979:
The one open risk on this PR was whether the two agent contract fields I rely on actually exist. I verified both against the official docs, and they do.

**Claude Code**, from the Command hook fields reference:

> `shell`: shell to use for this hook. Accepts `"bash"` or `"powershell"`. Defaults to `"bash"`, or to `"powershell"` on Windows when Git Bash isn't installed. Setting `"powershell"` runs the command via PowerShell on Windows. Does not require `CLAUDE_CODE_USE_POWERSHELL_TOOL` since hooks spawn PowerShell directly. Ignored when `args` is set.

**Codex**, from the hooks Config shape notes:

> `commandWindows` is an optional Windows-only command override. In TOML, use `command_windows` or `commandWindows`.

The managed-hooks example on that same Codex page ships `command` and `command_windows` side by side in a single handler, which is the exact shape this PR emits.

The Claude default also describes the problem better than my original framing did. Today the hook runs under Git Bash when it is installed and under PowerShell when it is not, so the current bash command works or breaks depending on the machine. Setting `shell` explicitly makes the behavior deterministic rather than only PowerShell-safe. Two details that should preempt review questions: this PR does not set `args`, so the "ignored when `args` is set" caveat does not apply, and no `CLAUDE_CODE_USE_POWERSHELL_TOOL` opt-in is required.

Sources: https://docs.claude.com/en/docs/claude-code/hooks and https://developers.openai.com/codex/hooks

One thing I cannot clear myself: no workflows have run on this PR. `gh pr checks` reports only CodeRabbit, so there is no Unit Tests, Windows Runtime, or E2E signal for a reviewer to judge. Could a maintainer approve the workflow run? I am happy to rebase or adjust once the matrix reports.

jjeg1979:
Small unblock request whenever you have a moment: the only failing check here is `Check PR Has type:* Label`, and the workflow output says to ask a maintainer since contributors cannot set labels. `type:bug` fits.

Everything else is green, including `Check PR Cognitive Load` at 366 lines.

ardelperal:
Heads up: I am proceeding on this issue under #2342 (opened 2026-08-03, ahead of this one) and force-pushed a reworked version. Per the policy that earlier-attached contributors retain priority when both PRs close the same issue, this one should close as duplicate once #2342 merges.

The reworked #2342 takes the inline-PowerShell approach instead of the Claude Code `shell: "powershell"` field you proposed here. Reason: the `shell` field requires the user's Claude Code to be recent enough to honor it, whereas the inline `powershell -NoProfile -Command '...'` wrapper runs on every Claude Code version that runs hooks. If you (or Alan) prefer the `shell` field approach for the cleaner JSON output, that is a one-commit swap on this branch and I am happy to coordinate.

What #2342 covers that overlaps with #2807:
- Windows-safe `UserPromptSubmit` command literal for Claude.
- Migration of pre-fix POSIX literal out of an existing `settings.json`.

What #2342 does NOT touch:
- Codex `SessionStart`. The Codex contract is `commandWindows`, not `shell`, and is out of scope for #2124 specifically. If you want the Codex fix folded in too, that is a separate PR.

No action required from your side. If you would rather #2807 proceed and #2342 close, please say so on the issue or in the maintainer thread and I will close mine.

jjeg1979:
You are right on both counts, and I checked rather than taking it on trust: #2342 was created 2026-08-03T13:40, mine 2026-08-08T18:58, both carrying `Closes #2124`. Priority is yours. Closing this now rather than parking it until #2342 merges, since a dead PR sitting in the queue costs reviewers attention either way.

Your compatibility argument also stands on its own merits. I used Claude Code's `shell: "powershell"` field, which only works if the user's Claude Code is recent enough to honor it. Your inline `powershell -NoProfile -Command` wrapper runs on every version that runs hooks at all. For a hook fix, where the whole point is working on the machine that is currently broken, the wider floor is the better trade. No swap needed on my side.

One piece worth not losing: this branch also carried the Codex `SessionStart` migration, adding a `wantCommandWindows` parameter to `pruneStaleSkillRegistryHooks` so a legacy bash-only Codex hook is treated as stale until it also has the Windows override. You are right that it is out of scope for #2124. I am keeping the branch so it can go up as its own PR against a Codex-specific issue if maintainers want it.

Thanks for the heads-up instead of letting both PRs sit. This is the second time I have implemented something that already had an earlier PR attached, so the lesson is mine to take: search open PR bodies for `Closes #N` before starting, not just the issue thread.

jjeg1979:
Closing as duplicate of #2342, which has priority by date. Details in the comment above.

Issue #2769, opened 2026-08-08
fix(sdd): two more runtime refusals name no runnable exit


Alan-TheGentleman:
### 📝 Bug Description

Sweep of the SDD runtime refusal surface after #1974 and #2530, both of which were the same defect: a refusal that either names an exit the state structurally refuses, or names none while a runnable one exists. Two more sites carry it. Anchors verified against `main` at `6d48fb6f`.

**A. `ErrRuntimeRescopeWidened` refuses a wider budget and names no way to get one.**

`runtime_ledger.go:1184-1195` refuses a widened `--max-changed-lines` or `--max-attempts` with "received X, the current objective allows Y", then points at `sdd-attempt status`. In that state (terminal failed or interrupted attempt, budget remaining, zero drift) status answers `next_action: begin`, which repeats the same objective at the same ceiling. Nothing tells the caller the route that does work: spend the objective's remaining attempts, and the run that exhausts them reaches `decision_required`, where `reset` is admitted and opens a fresh budget of any size.

This is the exact complement of #1974. That issue was the caller reaching for `reset` and being told nothing about `rescope`; this is the caller reaching for `rescope` and being told nothing about the exhaust-then-reset route. #2765 fixed one direction only.

**B. `ErrRuntimeObjectiveDone` refuses a same-work-unit begin and names no successor.**

`runtime_ledger.go:601-604` refuses `begin` against a completed objective with "SDD runtime objective is complete", then points at status, which answers `next_action: complete`. The code comment three lines above states the correct continuation itself: "the ordinary continuation is the successor objective, not a maintainer reset". `runtimeObjectiveAdvanceAdmissible` exists for exactly this, and `Complete` is only reachable from a passed, non-budget-exceeded attempt, so advance is admissible whenever the sole failing condition is the repeated work unit. Changing `--work-unit` is the whole fix, and the refusal never says so. `reset` is also admitted in this state.

**C. #2530 is the third instance and is already filed.** `compactBlockedExitText`'s maintainer-decision arm (`runtime_compact.go:440-443`) tells the caller to "rescope or reset", and `runtimeObjectiveRescopeStructurallyPermitted` returns false whenever `DecisionRequired` is set, so half that advice is structurally impossible in the one state that prints it.

### 🔄 Steps to Reproduce

[...]

PR #2835, opened 2026-08-08
test(bench): add a transition axis, and carry the typed cause into corrupt_authority

### LINKED ISSUE #2829: test(bench): no journey exercises a mid-lifecycle transition
### 📝 Bug Description

Every journey in the corpus runs with a **fixed configuration**: SDD on or off, RDD on or off, decided at the start and never changed. Nothing exercises a *transition* mid-lifecycle, and nothing chains two legitimate operations that are individually valid but invalid in sequence.

That is the gap #2830 fell through. It needs no damaged file and no disabled mode: `begin` → `finish failed` → `rescope` → `rescope` are four ordinary calls, the fourth wedges the ledger permanently, and zero journeys chain them. The corpus was green while a published build could brick a change.

`bench/axis_damaged_store.go` covers states reached by *damage*. What is missing is states reached by *sequence*.

### ✅ Expected Behavior

A transition axis that drives configuration and lifecycle changes as first-class journeys:

- two consecutive rescopes (the #2830 shape)
- reset after rescope
- RDD enabled, work started, RDD disabled mid-flight, delivery continues under ordinary policy
- RDD disabled, work started, RDD enabled mid-flight
- abandon while a remediation is open

Each asserts the ledger stays readable afterwards, which is the invariant #2830 broke and no journey checks.

### ❌ Actual Behavior

No journey changes mode or chains operations mid-lifecycle. A wedge reachable in four ordinary calls ships with a green corpus.

### Gentle AI Version

main @ `3d1e6735`

### 📋 Affected Area

Testing / CI

### LINKED ISSUE #2830: bug(sdd-attempt): verification rescope is rejected against terminal zero-drift finish
## Context

Automated provider-defect report from a blocked consumer workflow. Diagnostics are privacy-scrubbed: no project names, usernames, hostnames, absolute paths, raw argv, environment values, diffs, credentials, or source content are included.

## Observed behavior

On `gentle-ai 2.4.0-rc.1`, acquiring native SDD runtime-attempt authority for an independent verification continuation returns:

```text
state: blocked
reason: corrupt_authority
detail: the attempt ledger for this work unit cannot be read as valid authority
```

The advertised read-only status operation then fails with:

```text
replay SDD runtime revision sha256:858a8d7075c3d6657146162bf12df8e8ce42dda94b32cf9fa5a694c43c8104f5: objective rescope candidate does not match the terminal zero-drift finish
```

This is observed evidence only; the underlying root cause is not confirmed.

## Expected behavior

A verification continuation that follows the provider-owned attempt protocol should either acquire a bounded attempt, report the existing attempt as complete, or expose a documented native recovery/reset decision. A valid historical terminal record should not become unreadable merely because the next work unit narrows the evidence goal to independent verification.

## Actual behavior

Both acquisition and read-only status are blocked by the same replay invariant. No verifier can be launched through the required native attempt authority, and no supported continuation is returned.

## Minimal reproduction

1. Complete an SDD apply work unit and its review/delivery path, leaving a terminal runtime-attempt record in the Git common-dir.
2. Begin the independent verification continuation for the same SDD change with a distinct verification work-unit label and evidence goal.
3. Invoke native attempt acquisition with bounded attempts and changed-line limits.
4. Observe `blocked / corrupt_authority`.
5. Invoke read-only attempt status for the change.
6. Observe `objective rescope candidate does not match the terminal zero-drift finish`.

## Bounded attempts and outcomes

- Attempt acquisition: 1 invocation → blocked before actor launch.
- Read-only status: 1 invocation → replay error reproduced.
- Verifier/test harness launches: 0.
- Authority reset/repair operations: 0.
- Repository or source mutations: 0.

## Environment

- Gentle AI: `2.4.0-rc.1`
- OS/architecture: Linux x86_64
- Client: OpenCode
- Operation shape: `sdd-attempt acquire` for a verification continuation, followed by read-only `sdd-attempt status`

## Preserved-state evidence

The consumer candidate, worktree, branch, source files, and native authority were left unchanged. No verification actor, test command, reset, review lifecycle mutation, staging, commit, push, or archive operation ran after the failure.

## Duplicate lookup

Open and closed issues were searched using the exact failure phrases and broader SDD-attempt authority terms. No equivalent issue was found. #2702 concerns strict decoding of an older `begin_worktree` field after a version change; #2350 concerns keeping a passed apply attempt runnable for status-recommended verification. Neither reports this terminal zero-drift objective-rescope rejection.

Alan-TheGentleman:
Closes #2829
Refs #2830

## The gap

Every journey in the corpus ran a **fixed configuration**: SDD and RDD decided before the first command and never changed. `axis_damaged_store` covers states reached by corrupting files the CLI cannot write. Nothing covered states reached by **sequence**.

That is how #2830 shipped. `begin`, `finish failed`, `rescope`, `rescope` are four ordinary calls, no damaged file, no mode change, and the fourth leaves the attempt ledger permanently unreadable on a published build. A user found it, not the corpus.

## tr01 fails on `main`, deliberately

It reproduces #2830 through the CLI alone and reports the user's exact error string. **It is left failing so anyone running the bench sees precisely what is broken**, which is what a benchmark is for.

It is deliberately **not** asserted in CI. Pinning a journey as "must fail" would turn CI red the day someone fixes it, which is the bug-pinning antipattern. The CI step says to add tr01 to the assertion list in the PR that fixes #2830.

Note it does not assert the second rescope *succeeds*: whether consecutive rescopes are supported is still open on #2830. It asserts only that after any answer, `status` still works. That is the invariant that was missing.

## What tr02 and tr03 prove, narrowly

Two decoys narrowed tr02's claim and I kept the reasoning in the file rather than the strongest wording:

- An early version ran `reset` straight after the rescope. Injecting #2830's exact asymmetry into reset's replay changed **nothing**, because reset is refused at write time there and never reaches the publication path a wedge lives on. Decoration.
- Rebuilt to reach a reset that genuinely publishes, the same injected asymmetry **still** does not fire, because by then the last attempt belongs to the current objective. Reset does not have rescope's asymmetry.

So tr02 claims only what it demonstrates: #2769's advertised exhaust-to-widen route is walkable end to end. Its assertion does bite, proven when a misconfigured loop made every command "run" while the journey never reached the state it claimed.

tr03 is the maintainer-named case: turn review off mid-change, prove the ledger answers and the work continues. The kill switch is documented as always available; nothing tested it against work already in flight.

## Also here

Seven call sites discarded the error `store.load()` returned before reporting `corrupt_authority`. That is why #2830 could not be diagnosed from outside: an invalid HEAD, a broken chain link and a permissions refusal produced identical output. The cause now rides in `Detail`; the exit text keeps its exact bytes because callers route on it.

## Test Plan

- [x] Transition axis run against a real binary: tr01 `failed` with #2830's exact error, tr02 and tr03 `completed`.
- [x] Decoy on the cause threading: reverting it fails `TestUnreadableAuthorityBlockNamesItsCause`.
- [x] Two decoys on tr02, both reported above, one of which forced a redesign and one of which narrowed the claim.
- [x] CI assertion verified locally with the same `jq` expression.
- [x] `go test ./... -count=1` green at 66 packages; bench module green; `gofmt` and `go vet` clean.

## Contributor Checklist

- [x] Linked an approved issue
- [x] Exactly one `type:*` label

## Summary by CodeRabbit

* **New Features**
  * Added CLI lifecycle benchmark coverage for rescoping, resets, review-mode changes, delegated handoffs, and attempt exhaustion.
  * CI now validates key transition journeys and confirms each completes exactly once.

* **Bug Fixes**
  * Improved unreadable runtime ledger errors by preserving the underlying cause alongside recovery guidance.
  * Added clearer classification for remediation-successor and worktree-mismatch failures.
  * Added regression coverage for malformed runtime ledger scenarios.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,13 @@ jobs:
               [.journeys[] | select(.id == $id)] as $matches
               | ($matches | length == 1) and ($matches[0].status == "completed");
             completed_once("tr02-reset-after-rescope-keeps-the-ledger-readable") and
-            completed_once("tr03-review-disabled-mid-change-keeps-sdd-moving")
+            completed_once("tr03-review-disabled-mid-change-keeps-sdd-moving") and
+            completed_once("tr04-reset-then-rescope-keeps-the-ledger-readable") and
+            completed_once("tr05-review-re-enabled-mid-change-keeps-sdd-moving") and
+            completed_once("tr06-mode-flip-during-an-active-attempt") and
+            completed_once("tr07-reset-then-reset-keeps-the-ledger-readable") and
+            completed_once("tr08-scope-change-after-a-delegated-handoff") and
+            completed_once("tr09-mode-flip-while-a-review-lineage-is-open")
           ' "$RUNNER_TEMP/bench-transition.json"
           "$RUNNER_TEMP/gentle-ai-bench" run --binary "$RUNNER_TEMP/gentle-ai" --axis source-coupled --only j57-sdd-authority-drift-during-discovery-fails-closed --out "$RUNNER_TEMP/bench-source-coupled-untagged.json"
           jq -e '
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,15 +82,20 @@ jobs:
             completed_once("j75-intended-untracked-selection-executes-printed-start")
           ' "$RUNNER_TEMP/bench-portable.json"
           # Transition axis (#2829): sequences of ordinary operations, which is
-          # how #2830 shipped. tr01 reproduces that live defect and is
-          # deliberately NOT asserted here: pinning a journey as "must fail"
-          # would make CI go red the day someone fixes it. Add tr01 to this
-          # list in the PR that fixes #2830.
-          "$RUNNER_TEMP/gentle-ai-bench" run --binary "$RUNNER_TEMP/gentle-ai" --axis transition --out "$RUNNER_TEMP/bench-transition.json"
+          # how #2830 shipped. tr01 reproduces that live defect, so the runner
+          # exits non-zero on this axis until #2830 is fixed; the jq assertion
+          # below is the real gate, and it fails if the run produced no parsable
+          # report. tr01 is not pinned as "must fail" because that would make CI
+          # go red the day someone fixes it. Instead the first clause requires
+          # tr01 to be the ONLY journey not completed, so a twelfth journey
+          # regressing still turns this step red.
+          "$RUNNER_TEMP/gentle-ai-bench" run --binary "$RUNNER_TEMP/gentle-ai" --axis transition --out "$RUNNER_TEMP/bench-transition.json" || true
           jq -e '
             def completed_once($id):
               [.journeys[] | select(.id == $id)] as $matches
               | ($matches | length == 1) and ($matches[0].status == "completed");
+            ([.journeys[] | select(.status != "completed") | .id] | sort) ==
+              ["tr01-consecutive-rescope-keeps-the-ledger-readable"] and
             completed_once("tr02-reset-after-rescope-keeps-the-ledger-readable") and
             completed_once("tr03-review-disabled-mid-change-keeps-sdd-moving") and
             completed_once("tr04-reset-then-rescope-keeps-the-ledger-readable") and
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,19 @@ jobs:
             completed_once("j58-sdd-foreign-openspec-path-fails-closed") and
             completed_once("j75-intended-untracked-selection-executes-printed-start")
           ' "$RUNNER_TEMP/bench-portable.json"
+          # Transition axis (#2829): sequences of ordinary operations, which is
+          # how #2830 shipped. tr01 reproduces that live defect and is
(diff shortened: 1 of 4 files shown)


Alan-TheGentleman:
Applied `size:exception` at 601 lines, and the reasoning rather than just the label.

This is one axis file plus its CI assertion. Splitting it would mean shipping a transition axis with three journeys and then two follow-up PRs adding more to the same file, which reviews worse than reading the enumeration once: the value of the batch is that the pairs were derived from the verb space rather than picked, and that argument only holds when you can see all eleven together.

Roughly half the additions are comments, and they carry the two things a reader needs and cannot recover from the code: why each pair can block, and what two decoys proved about journeys that looked fine.

The final shape, eleven journeys across three surfaces:

**Ledger verb pairs** — `rescope`+`rescope` (tr01, the live defect), `rescope`+`reset` (tr02), `reset`+`rescope` (tr04), `reset`+`reset` (tr07), `handoff`+`rescope` (tr08).

**Mode flips** — off mid-change (tr03), on mid-change (tr05), flipped with an attempt open (tr06), flipped with a review lineage open (tr09).

**Cross-surface** — scope change after the review is bound (tr10), review abandoned while an attempt runs (tr11).

**Ten pass, one fails.** The ten passing ones are not filler: tr04, tr10 and tr11 passing is what scopes #2830 to consecutive rescopes specifically, rather than to generation advancement or to authority interaction generally. A green journey that narrows a defect earns its place.

PR #2923, opened 2026-08-10
feat(runtimecertification): add the runtime oracle evaluator (slice 1 of 2)

### LINKED ISSUE #1873: test(adapters): certify installed capabilities through each agent's native runtime
### Pre-flight Checklist

- [x] I have searched existing issues and pull requests and did not find a general native-runtime capability certification contract.
- [x] I understand that implementation PRs require the issue to receive `status:approved`.

### 🔍 Affected Area

Installation Pipeline

### 💡 Problem Statement

Gentle AI currently proves three different things, but presents them too close to one another:

1. an adapter declares a capability;
2. the installer renders and writes files;
3. the target client actually loads and can operate that capability.

The first two do not imply the third.

At audited commit `1b1a8676e028e5ebf91c160893468fa18e30abf5`:

- [`docs/agents.md`](https://github.com/Gentleman-Programming/gentle-ai/blob/1b1a8676e028e5ebf91c160893468fa18e30abf5/docs/agents.md#L7-L28) advertises Skills and MCP as supported across all listed agents.
- [`AgentCapabilityManifest`](https://github.com/Gentleman-Programming/gentle-ai/blob/1b1a8676e028e5ebf91c160893468fa18e30abf5/internal/agents/capabilitymanifest/manifest.go#L36-L59) explicitly contains static adapter claims and no runtime observations.
- [`doctor.Status`](https://github.com/Gentleman-Programming/gentle-ai/blob/1b1a8676e028e5ebf91c160893468fa18e30abf5/internal/doctor/doctor.go#L18-L25) can say pass/warn/fail, but has no vocabulary for the capability evidence boundary.

The distinction is not theoretical:

- #1805 found an install package whose embedded skill referenced a shared file that was not included in the install closure.
- #1868 found Claude MCP configuration written to a location/schema that Claude Code did not load.

Both paths can be file-level PASS while runtime behavior is absent. This makes the support matrix and diagnostics stronger than the evidence that currently backs them.

### 📦 Proposed Solution

Introduce a provider-neutral capability certification registry and native-runtime conformance suite.

#### 1. Evidence state machine

Represent each `(agent, capability, client version range, OS)` observation as:

`declared -> rendered -> written -> loaded -> operational`

These states are monotonic evidence levels, not synonyms:

- `declared`: adapter contract claims support;
- `rendered`: expected projection was produced;
- `written`: managed output exists at the resolved safe target;
- `loaded`: the pinned native client reports or observably loads it;
- `operational`: a deterministic, non-model-mediated probe completes.

`unknown`, `unverified`, and `degraded` must remain representable. File existence alone must never become runtime PASS.

#### 2. Capability-scoped support tiers

Support tier belongs to the full certification tuple, not to an agent-wide boolean:

- `core`: native load/operation evidence is release-blocking for the declared client/OS range;
- `compatible`: renderer/schema coverage plus periodic native smoke, with visible degradation when native evidence is missing;
- `experimental/community`: integration is available, but is never reported as verified without runtime evidence.

One agent may be Core for skills and Experimental for delegation.

#### 3. Deterministic native probes

- Run each case with isolated HOME/config roots and a pinned client version.
- MCP uses an inert local echo/handshake server and the client's native list/health/introspection surface.
- Skills, persona, routing, commands, and workflows use native parser/introspection output or an instrumented registry/loader trace.
- If a client exposes no trustworthy inspection surface, the result is `unverified`; a model response is not a substitute.
- Add negative controls for wrong path, wrong schema, dangling relative references, stale client process after sync, malformed existing config, and an unsafe root/symlink/reparse-point escape.
- `install` and `sync` must preserve unrelated user-owned bytes/keys while the probe observes only the managed projection.

#### 4. One machine-readable evidence contract

Each result should bind:

- candidate commit and immutable installer asset digest;
- agent/capability/tier;
- client version and OS/arch;
- fixture and oracle version/digest;
- sanitized environment/config-root identity;
- observed evidence state and native output digest;
- verdict `pass | fail | inconclusive | not_run`;
- limitations and capability-specific remediation.

Generated docs, the support matrix, and `doctor` must derive their runtime claims from this registry rather than independently converting adapter booleans into "Yes" or PASS.

#### 5. Calibrated release gate

This issue should not make every optional third-party client a blocker.

- Stable release is gated only on the clients/capabilities explicitly classified as Core.
- A capability without native evidence may remain available, but the docs and doctor must expose it as Compatible/Experimental or unverified.
- Start with the smallest Core matrix maintainers can execute reproducibly, then expand it without changing the evidence semantics.

### ✅ Acceptance Criteria

- [ ] A versioned schema represents the certification tuple, evidence level, verdict, limitations, and evidence digests.
- [ ] `loaded` and `operational` can only be produced by a native or instrumented runtime observation.
- [ ] `doctor` distinguishes `declared`, `written`, `loaded`, `operational`, `unverified`, and `degraded`.
- [ ] MCP has an offline deterministic handshake probe through the client's native surface.
- [ ] Every Core capability has a pinned client/OS test in an isolated HOME.
- [ ] Negative path/schema/reference/stale-process controls fail without modifying user-owned configuration.
- [ ] Unknown or unavailable introspection never appears as PASS.
- [ ] Docs/support claims are generated from, or checked against, the same certification registry.
- [ ] Evidence binds the exact commit, installer asset, client version, fixture, oracle, OS, and sanitized environment.
- [ ] The suite requires no external network access, paid model response, or model-quality judgment.

### 🔄 Alternatives Considered

- **Keep adapter booleans plus renderer unit tests.** Necessary for contract drift, but cannot detect a client ignoring a valid-looking file.
- **Use one successful model conversation as a smoke test.** Non-deterministic and conflates loader conformance with model/provider behavior.
- **Make all agents release-blocking immediately.** Operationally too broad; capability-scoped tiers preserve truthful support without freezing releases.
- **Treat #1160, #1805, or #1868 as the general contract.** They are valuable concrete cases, but respectively cover a specific integration/version, reference closure, and Claude MCP loading rather than the cross-agent evidence model.

### 📎 Additional Context

This is authorized open-source QA and resilience testing against local fixtures and installed client runtimes. It does not require access to third-party systems or user source code.

The architectural objective is not a larger matrix for its own sake. It is a truthful chain from adapter declaration to observable client behavior, so "supported" becomes falsifiable and reproducible.

jjeg1979:
## 🔗 Linked Issue

Part of #1873 (closed by slice 2 of this chain)

---

## 🏷️ PR Type

---

## 📝 Summary

Adds the provider-neutral runtime oracle evaluator: it decides whether oracle evidence certifies the current identity, without launching external providers or treating a rendered file as proof.

- Declared and written evidence stay strictly below certified evidence.
- A complete tuple plus non-empty input and evidence-boundary digests are required; a payload whose subject does not match the record is rejected.
- Expiry, revocation, and identity drift each resolve to their own state instead of one opaque failure.

This slice is the library and its unit tests. Nothing imports it yet.

## 🔗 Chain Context

This PR was 422 lines and failed the 400-line cognitive budget. Rather than ask for `size:exception` on a non-urgent feature, it is now split by dependency direction.

```
📍 slice 1  feat(runtimecertification): the evaluator      323 lines  ← this PR
   slice 2  feat(doctor): surface it as a doctor check      99 lines
```

- **Start:** no prior dependency, applies to `main`.
- **End:** a tested evaluator with no callers.
- **Follow-up:** slice 2 wires it into doctor and closes #1873. Branch is ready at `jjeg1979:feat/1873-runtimecertification-doctor`; I will open it once this merges. It is not open yet because GitHub cannot use a fork branch as a PR base, so opening it now against `main` would show a 422-line diff and re-break the same budget the split exists to respect.
- **Out of scope:** registering the check in the doctor run list. Both slices leave the code unreachable, which is why slice 2 carries baseline entries.

## 🧮 Deadcode baseline

Worth flagging because it looks inverted at first glance: **this slice changes no baseline entries, and slice 2 adds all twelve** — including the nine that belong to files in *this* PR.

`scripts/deadcode-ratchet.sh` measures reachability from `./cmd/gentle-ai`. A package nothing imports is never analyzed, so the evaluator alone sits outside that graph entirely and the ratchet reports nothing. Slice 2's import is what pulls both packages in, and neither has a caller yet. The growth is real and belongs to the slice that creates the edge, not the slice that adds the code.

## ✅ Validation

- [x] `go build ./...`
- [x] `go vet ./internal/runtimecertification/`
- [x] `go test ./internal/runtimecertification/` — pass
- [x] `gofmt -l internal/runtimecertification/` — clean
- [x] `scripts/deadcode-ratchet.sh` — `no new unreachable functions`

## 📌 Maintainer action

Still needs exactly one `type:*` label (`type:feature`); this account cannot set labels. That was the other CI failure alongside the budget, and it is the only one left.

## Summary by CodeRabbit

* **New Features**
  * Added runtime certification support for validating schemas, evidence, identities, digests, expiration, revocations, and oracle results.
  * Added certification evaluation states and policy handling.
  * Added OpenCode debug-agent certification data with command digest generation.

* **Bug Fixes**
  * Added validation for malformed certification data, version mismatches, missing digests, expired records, revoked evidence, and subject mismatches.

* **Tests**
  * Added coverage for successful and rejected certification scenarios, including evidence-level and oracle-status variations.

### DIFF
--- a/internal/runtimecertification/certification.go
+++ b/internal/runtimecertification/certification.go
@@ -1,0 +1,255 @@
+// Package runtimecertification evaluates runtime oracle evidence without
+// launching external providers or treating rendered files as proof.
+package runtimecertification
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"slices"
+	"strings"
+	"time"
+)
+
+type SchemaVersion string
+
+const SchemaV1 SchemaVersion = "gentle-ai.runtime-certification/v1"
+
+type State string
+
+const (
+	StateDeclared         State = "declared"
+	StateWritten          State = "written"
+	StateCertifiedCurrent State = "certified_current"
+	StateStale            State = "stale"
+	StateRevoked          State = "revoked"
+	StateInconclusive     State = "inconclusive"
+	StateFailed           State = "failed"
+)
+
+type EvidenceKind string
+
+const (
+	EvidenceDeclared EvidenceKind = "declared"
+	EvidenceWritten  EvidenceKind = "written"
+	EvidenceOracle   EvidenceKind = "oracle"
+)
+
+type OracleStatus string
+
+const (
+	OracleStatusPass         OracleStatus = "pass"
+	OracleStatusFail         OracleStatus = "fail"
+	OracleStatusInconclusive OracleStatus = "inconclusive"
+)
+
+const (
+	ClientOpenCode           = "opencode"
+	OracleOpenCodeDebugAgent = "opencode.debug-agent/v1"
+	CapabilityResolvedAgent  = "opencode.agent-config.resolved/v1"
+)
+
+type Tuple struct {
+	ClientName              string   `json:"client_name"`
+	ClientVersion           string   `json:"client_version"`
+	ClientVersionRange      string   `json:"client_version_range"`
+	BinarySHA256            string   `json:"binary_sha256"`
+	OS                      string   `json:"os"`
+	Arch                    string   `json:"arch"`
+	Capability              string   `json:"capability_id"`
+	Subject                 string   `json:"subject"`
+	OracleID                string   `json:"oracle_id"`
+	CommandDigest           string   `json:"command_digest"`
+	InputDigests            []string `json:"input_digests"`
+	EvidenceBoundaryDigests []string `json:"evidence_boundary_digests"`
+}
+
+type Record struct {
+	SchemaVersion SchemaVersion `json:"schema"`
+	EvidenceKind  EvidenceKind  `json:"evidence_kind"`
+	Tuple
+	Limitations  []string     `json:"limitations,omitempty"`
+	ExpiresAt    time.Time    `json:"expires_at"`
+	OracleResult OracleResult `json:"oracle_result,omitempty"`
+}
+
+type OracleResult struct {
+	Status  OracleStatus    `json:"status"`
+	Payload json.RawMessage `json:"payload,omitempty"`
+}
+
+type Revocation struct {
+	OracleID      string
+	ClientName    string
+	ClientVersion string
+	BinarySHA256  string
+}
+
+type Evaluation struct {
+	State  State
+	Detail string
+}
+
+type Policy struct {
+	Expected    Tuple
+	Now         time.Time
+	Revocations []Revocation
+}
+
+func Evaluate(record Record, policy Policy) Evaluation {
+	if record.SchemaVersion != SchemaV1 {
+		return Evaluation{State: StateFailed, Detail: "unsupported runtime certification schema"}
+	}
+	if record.EvidenceKind == EvidenceDeclared {
+		return Evaluation{State: StateDeclared, Detail: "declared evidence is not runtime certification"}
+	}
+	if record.EvidenceKind == EvidenceWritten {
+		return Evaluation{State: StateWritten, Detail: "written evidence is not runtime certification"}
+	}
+	if record.EvidenceKind != EvidenceOracle {
+		return Evaluation{State: StateFailed, Detail: "unknown evidence kind"}
+	}
+	if err := validateOracleEvidenceBoundary(record); err != nil {
+		return Evaluation{State: StateFailed, Detail: err.Error()}
+	}
+	if !matchesTuple(record, policy.Expected) {
+		return Evaluation{State: StateStale, Detail: "runtime certification tuple does not apply to the current identity"}
+	}
+	if isRevoked(record, policy.Revocations) {
+		return Evaluation{State: StateRevoked, Detail: "runtime certification tuple was revoked"}
+	}
+	if expired(record.ExpiresAt, policy.Now) {
+		return Evaluation{State: StateStale, Detail: "runtime certification expired for the current identity"}
+	}
+
+	switch record.OracleResult.Status {
+	case OracleStatusPass:
+		if err := validateOraclePayload(record.OracleID, record.Subject, record.OracleResult.Payload); err != nil {
+			return Evaluation{State: StateFailed, Detail: err.Error()}
+		}
+		return Evaluation{State: StateCertifiedCurrent, Detail: "runtime oracle passed for current identity"}
+	case OracleStatusInconclusive:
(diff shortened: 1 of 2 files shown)


jjeg1979:
Thanks for the latest CI signal. I can address the deadcode ratchet failure from this branch.

The two remaining non-code blockers need maintainer action:
- add exactly one type label, likely `type:feature`
- either approve `size:exception` or ask me to split/trim, since the PR is currently 410 changed lines against the 400-line budget

I’ll avoid guessing maintainer intent here. Once the preferred direction is clear, I’ll adjust the branch accordingly.

jjeg1979:
Split this instead of waiting on the `size:exception` call, since the answer only mattered while the PR stayed blocked and it had drifted from 410 to 422 lines in the meantime.

It is now 323 lines, so the cognitive budget check should pass on its own. The doctor slice is 99 lines and ready on `jjeg1979:feat/1873-runtimecertification-doctor`; I am holding it until this merges because a fork branch cannot be a PR base, and opening it against `main` now would show the full 422-line diff and defeat the split.

One thing worth a second opinion: the deadcode baseline grows entirely in slice 2, including the nine entries for files in this PR. That is because the ratchet measures reachability from `./cmd/gentle-ai` and never analyzes a package nothing imports, so the evaluator only enters the graph when the doctor slice imports it. Reasoning is in the PR description. Happy to move those entries if you would rather see them land with their own files.

Still needs a `type:feature` label when you get a moment.

jjeg1979:
Two mechanical blockers are keeping this parked, and both need a maintainer click.

**The workflows never ran.** Every run on this branch sits at `completed/action_required`, so the fork approval was never given after the 2026-08-14 push. CodeRabbit is the only signal here because it is an external status, not an Action. For comparison, #3421 ran 15 checks and this PR ran zero. Repo-wide there are currently 388 runs in `action_required`, so this is not specific to this PR.

**There is no type label**, so "Check PR Has type:* Label" cannot pass. The parent issue #1873 already carries `type:feature`, `status:approved`, `priority:high` and `up-for-grabs`, so the evidence for it already exists.

The change itself is still clean: `mergeable=true`, no conflicts despite being behind main, and `internal/runtimecertification/` does not exist on main yet, so nothing has landed that overlaps it. Slice 2 (the doctor, 99 lines) is ready on a separate branch once this merges.

Worth noting the second blocker is exactly what #3272 describes: the type is already in the PR checkbox, so the label could be derived instead of hand applied.

Issue #2540, opened 2026-08-05
fix(sdd): multi-repository change reports apply ready with only planning-repo edit authority


jivn:
### Pre-flight Checklist

- [x] I searched open and closed issues for multi-repository SDD, sibling repositories, `actionContext`, and `allowedEditRoots`.
- [x] This is not a duplicate of #1302: that proposal covers repo-local work-unit queueing and parallelism, while this occurrence blocks a sequential multi-repository change because no sibling repository receives edit authority.
- [x] No consumer repository, user, host, credential, path, source, diff, or environment value is included.

### Bug Description

A multi-repository SDD change can reach `applyState: ready` and `nextRecommended: apply` while `actionContext.allowedEditRoots` contains only the planning repository. The task plan targets separate sibling Git repositories, and the `sdd-apply` contract correctly requires the executor to stop when a needed edit is outside the authorized roots.

This creates a contradictory terminal route: native status authorizes apply, but no safe apply actor can execute the first work unit.

Observed evidence, not an inferred root cause:

- The change had complete proposal, two specs, design, and tasks.
- Native status reported 14 pending tasks, apply ready, and no blocked reasons.
- The first three repository-local work units targeted three sibling repositories in a strict sequential rollout.
- The returned edit authority named only the planning repository.
- No product apply was acquired or launched; all 14 tasks remained pending and consumer state was preserved.

### Steps to Reproduce

1. Create a disposable workspace containing one planning Git repository and two sibling service Git repositories.
2. Configure the planning repository as a multi-repository SDD ecosystem.
3. Create an OpenSpec change whose proposal, specs, design, and tasks are complete and whose first work units explicitly target the sibling service repositories.
4. Query native SDD status for the change with structured instructions enabled.
5. Observe that status reports apply ready and recommends apply.
6. Inspect `actionContext.allowedEditRoots`.
7. Observe that only the planning repository is authorized.
8. Start no writer: the documented apply guard would have to stop before editing either sibling repository.

Minimal sanitized layout:

```text
<workspace>/
  <planning-repo>/.git
  <service-a>/.git
  <service-b>/.git
```

Relevant sanitized status shape:

```json
{
  "schemaName": "gentle-ai.sdd-status",
  "schemaVersion": 1,
  "taskProgress": { "total": 14, "completed": 0, "pending": 14 },
  "applyState": "ready",
  "nextRecommended": "apply",
  "blockedReasons": [],
  "actionContext": {
[...]

PR #3032, opened 2026-08-11
fix(sdd): detect edit targets outside the roots inside one repository

### LINKED ISSUE #2891: fix(sdd): detect out-of-root sibling directories inside one Git repository
### Pre-flight Checklist

- [x] I searched open and closed issues for same-repository sibling directories, nested OpenSpec subprojects, `allowedEditRoots`, `edit_authority_missing`, and #2540 follow-up coverage.
- [x] This is not a duplicate of #2540 or #2547: those reports cover task targets that resolve to a different Git root, while this target remains inside the planning subproject's parent Git repository but outside the authorized nested workspace.
- [x] I understand that no implementation work should begin until a maintainer approves the issue.

> Automated provider-defect notice: this report was prepared from sanitized observations after bounded recovery attempts. It intentionally excludes consumer identifiers, private paths, raw payloads, source, diffs, and credentials.

### Bug Description

Native SDD status does not detect an explicit task target in a sibling directory when the planning workspace is a nested subproject and both directories are inside one Git repository. Apply remains ready even though the sibling directory is outside `allowedEditRoots`.

This report is limited to observed behavior and does not assert an internal root cause.

#### Relationship to #2540

#2540 is closed after its full edit-authority chain merged, and the installed published build exposes `sdd-attempt grant`. This report covers a narrower operation shape that remains observable after following that published recovery guidance: the additional target is outside the planning workspace but inside the same parent Git repository.

No `.gentle-ai-instance` marker or exact instance-bound grant invocation was emitted. Manually inventing the caller-owned instance would not constitute a valid relayed consent grant.

#### Operation shape

The consumer layout is one Git repository, not multiple repositories:

```text
<repo>/.git
<repo>/<subproject>/openspec/...
<repo>/<service-dir>/...
```

Native SDD planning uses `<repo>/<subproject>` as its cwd and workspace root. `allowedEditRoots` contains only `<repo>/<subproject>`. A pending unchecked task contains an explicit backticked target such as `../<service-dir>/file`.

### Steps to Reproduce

1. Create one Git repository with `<subproject>/openspec` and a sibling `<service-dir>`.
2. Complete an OpenSpec change under `<subproject>` whose next unchecked checkbox includes a backticked `../<service-dir>/file` target.
3. Query native SDD status with cwd `<subproject>`.
4. Observe apply ready, `allowedEditRoots=[<subproject>]`, and no consent envelope.

### Expected Behavior

Any explicit task target outside `allowedEditRoots`, including a sibling directory inside the same Git repository, should block apply and emit the typed per-change consent envelope (`gentle-ai.sdd-integration.consent/v1`) with reason `edit_authority_missing`. Alternatively, the action context should authorize the full intended repository scope.

### Actual Behavior

After making the path token explicit, status still reports apply ready with zero blocked reasons and no `blocked(edit_authority_missing)` consent envelope. The sibling target remains outside `allowedEditRoots`.

### Gentle AI Version

`2.4.0-rc.1`

### Operating System

Linux `x86_64`

### AI Agent / Client

OpenCode `1.18.15`, agent binding `opencode`

### Affected Area

Other: native SDD status edit-authority detection (`gentle-ai.sdd-status/v1`).

#### Bounded attempts

- One original status query and one continue before explicit paths were added.
- One planning-only task correction to add explicit resolvable paths.
- One status query after that correction.

#### Mutation outcome and preserved state

No grant, reset, acquire, apply actor, product or source edit, test, branch, commit, push, pull request mutation, or permission workaround occurred. The planning metadata correction remains preserved, and product/runtime state remains unchanged.

Alan-TheGentleman:
Reproduced on `main` before the change: one Git repository, planning workspace at `<repo>/subproject`, a task naming `` `../service/handler.go` ``.

```
applyState      : ready
blockedReasons  : []
allowedEditRoots: ["<repo>/subproject"]
```

Apply was admitted for a target outside every authorized edit root.

## Cause

`detectUnauthorizedEditRoots` was written for the cross-repository case and states the limit in its own comment: it flags a token only when it "resolves to a real Git repository different from the planning repository". That treats *outside the workspace* and *a different repository* as the same thing.

```go
target := gitRootOf(resolved)
if target == "" || target == planningGitRoot || withinAnyRoot(target, allowed) {
    continue                       // <- #2891 exits here
}
```

A sibling directory in a nested-subproject layout shares the planning Git root, so it was skipped. #2540's edit-authority chain does not cover it (its targets resolve to a different Git root) and neither does #3001's filesystem-root guard.

## Change

Detection now asks what the roots actually answer:

- target in **another** repository -> unit stays the Git root (unchanged);
- target in **the planning** repository -> unit is the containing directory, which is what an operator would grant and narrower than the whole repository.

After, same fixture:

```
applyState : blocked
reason     : blocked(edit_authority_missing): tasks.md targets paths outside the authorized edit roots:
             "<repo>/service"; edit tasks.md so every work unit stays inside the authorized edit roots,
             or grant this change edit authority for those roots
consent    : gentle-ai.sdd-integration.consent/v1
```

It plugs into the existing grant seam with no new mechanism: `TestGrantedSiblingDirectoryRestoresReadiness` shows an authorized root covering the sibling clears the block.

## False positives

The flat case is untouched by construction: when the workspace is the Git root, every in-repository path is already inside an authorized root. `TestSingleRepoTasksKeepStatusByteIdentical` still passes unchanged, and `TestNestedSubprojectTargetsInsideItsOwnWorkspaceStayReady` pins that a subproject editing its own files keeps readiness.

## A wording fix I did not plan, and the mistake behind it

The envelope said the roots were repository roots, which is false for the new kind. My first fix re-derived the kind by probing the filesystem, and the shipped fixture caught it: `/workspace/service-a` does not exist on disk, so the probe labelled two separate repositories as directories inside the planning repository. Task plans routinely name paths that do not exist yet, so the probe would mislabel exactly the roots it cannot see. The line now states the fact that holds for both and needs no probe.

Two pins moved deliberately and are documented in place: the consent fixture (regenerated from the builder itself, which is what its test requires) and its digest. No field, choice, answer token, or invocation changed.

`go test ./...` clean. Both new behaviors were RED first.

Closes #2891

### DIFF
--- a/contracts/sdd-integration/v1/fixtures/consent.fixture.json
+++ b/contracts/sdd-integration/v1/fixtures/consent.fixture.json
@@ -10,17 +10,17 @@
     "/workspace/service-b"
   ],
   "headline": "This change plans work outside its authorized edit roots.",
-  "reason": "the task plan targets repository roots that no edit authority covers, so apply stays blocked until a human decides.",
+  "reason": "the task plan targets paths that no edit authority covers, so apply stays blocked until a human decides.",
   "value": "Granting scopes edit authority to this change alone: the grant is recorded in the change's ledger, auditable, and dies with archive.",
   "evidence": [
-    "/workspace/service-a is a Git repository root outside the authorized edit roots",
-    "/workspace/service-b is a Git repository root outside the authorized edit roots"
+    "/workspace/service-a is outside the authorized edit roots",
+    "/workspace/service-b is outside the authorized edit roots"
   ],
   "choices": [
     {
       "answer": "granted",
       "label": "Grant this change edit authority over the named roots",
-      "effect": "This change's apply actor may edit the named repositories. The grant is per-change, audited (who, when, which roots), and dies with archive; nothing is granted to any other change.",
+      "effect": "This change's apply actor may edit the named roots. The grant is per-change, audited (who, when, which roots), and dies with archive; nothing is granted to any other change.",
       "invocation": "gentle-ai sdd-attempt grant --cwd \"/workspace/planning\" --change multi-repo-rollout --root \"/workspace/service-a\" --root \"/workspace/service-b\" --actor maintainer --reason edit-authority-consent-granted --request-id grant-ce874466367354a2 --change-instance sdd-b2f4c0a1d8e64953a7c15b9d3e0f2a68"
     },
     {
@@ -31,7 +31,7 @@
     }
   ],
   "off_path": {
-    "note": "To keep this change single-repository instead, edit its tasks.md so no work unit targets an unauthorized root, then re-enter through 'gentle-ai sdd-status multi-repo-rollout --cwd \"/workspace/planning\"'.",
+    "note": "To keep this change inside its current authority instead, edit its tasks.md so no work unit targets an unauthorized root, then re-enter through 'gentle-ai sdd-status multi-repo-rollout --cwd \"/workspace/planning\"'.",
     "command": "gentle-ai sdd-status multi-repo-rollout --cwd \"/workspace/planning\""
   }
 }
--- a/internal/sddstatus/consent_contract_test.go
+++ b/internal/sddstatus/consent_contract_test.go
@@ -31,7 +31,13 @@ var sddConsentDeclineInvocationShape = regexp.MustCompile(
 func TestSDDIntegrationConsentContractsArePinned(t *testing.T) {
 	root := filepath.Join("..", "..", "contracts", "sdd-integration", "v1")
 	want := map[string]string{
-		"fixtures/consent.fixture.json": "aecd0f4d3ae85527a33ebc708a7a49d7f9fda99b13ccabd1fbe9b89bbf2c41e9",
+		// #2891 widened detection to a second kind of unauthorized root: a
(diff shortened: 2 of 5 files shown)


Alan-TheGentleman:
Closing as redundant. @decode2's #2958 landed the same fix in `1ebb8e20` while this was in CI, and it is the better change.

I verified rather than assumed: the three tests from this branch pass unmodified against the merged implementation, and the reported layout now blocks with the same directory unit.

```
applyState : blocked
reason     : blocked(edit_authority_missing): tasks.md targets edit paths outside the authorized
             edit roots: "<repo>/service"; ...
consent    : gentle-ai.sdd-integration.consent/v1
```

Their coverage is also broader than mine: `TestDeletedAllowedRootDoesNotBroadenSiblingAuthorization` is a case I never considered, and `TestSDDSameParentRepositoryConsentGrantLoop` exercises the grant end to end rather than at the unit seam.

This is on me: #2958 was open since 2026-08-10, a day before I started, and I did not check for an existing PR against the issue before writing code. Sorry for the duplicate noise, @decode2.

Issue #2855, opened 2026-08-09
bug(sdd): sdd-apply repeatedly returns sdd_task_result_empty during remediation


teba-isea:
### Environment
- Gentle AI: `2.3.1`
- Client: OpenCode `1.18.15`
- Runtime: Node.js 24
- OS/architecture: Darwin arm64

### Operation shape
A bounded `sdd-apply` remediation actor was launched twice against the same preserved native objective after explicit user authorization.

### Observed evidence
Both bounded attempts returned `gentle-ai.sdd-task-result-failure/v1` with code `sdd_task_result_empty` and no valid task result. Native status showed zero changed lines and no evidence revision after each attempt.

### Expected behavior
The delegated apply actor returns a valid typed result, or an actionable provider failure that does not leave the active objective awaiting `finish`.

### Actual behavior
The provider returned an empty-result transport failure twice. No edits or evidence were recorded, while remediation remained blocked.

### Minimal reproduction
1. Start a bounded remediation objective after failed SDD verification.
2. Acquire the active objective token.
3. Launch the dedicated `sdd-apply` actor for a small typing fix.
4. Observe `sdd_task_result_empty`; execute the prescribed native status continuation.
5. With explicit authorization, repeat once against the same preserved objective.
6. Observe the same typed failure with zero changed lines and no evidence.

### Preserved-state evidence
- 45/45 implementation tasks remain complete.
- Remediation remains required.
- Verification and archive remain blocked.
- The active objective is preserved with zero changed lines and no evidence.

### Duplicate lookup
Open and closed issues were searched before creation. Existing issues #2677 and #2117 cover different SDD phases; #2609 covers a general worker. No equivalent repeated `sdd-apply` remediation report was found.

This report contains observed evidence only; no root cause is asserted. Privacy scan passed before GitHub operations.

yeerliin:
New occurrence of the same behavior. This evidence is privacy-scrubbed.

### Environment
- Gentle AI: 2.3.1-0.20260808092929-bf1d9de53b8c
- Client: OpenCode 1.18.15
- OS/architecture: Windows amd64

### Operation and observed outcome
A bounded, artifact-only sdd-apply correction was launched to update two stale SDD record fields after implementation and focused checks had already passed. The first launch coincided with a brief network outage. After connectivity was restored, the same bounded operation was launched again.

[...]

PR #3062, opened 2026-08-11
fix(opencode): don't latch the session when a failed SDD phase left real work behind

jarruiz:
## Problem

`sdd_task_result_empty` can be a **transport/timing failure, not a phase failure**. Observed on gentle-ai 2.3.0 + OpenCode 1.18.15 (issue #2855): the `sdd-apply` actor **executes the full work unit** — creates source + test files, runs TDD RED→GREEN, persists `apply-progress.md`, marks tasks complete — yet the task result never reaches the orchestrator (runtime log shows the subagent stream finishing and the session later recording `error=Aborted` ~6.5 min after the last stream).

The plugin unconditionally latched `failedSDDSessions` on any empty/malformed result. That blocks **every later SDD launch in the session with zero execution**, even though the work exists and is recoverable via the handoff's own continuation (`sdd-status`).

## Fix

Add `hasRealWork()`: before latching the session, check whether the failed phase actually left work behind:

1. Extract the change name from the task prompt (JSON binding form or plain `change:` form).
2. Prefer the prompt's `Project root:` over the plugin cwd, which can resolve to `/` in global/desktop sessions and makes both the native query and the filesystem fallback miss the real project.
3. Query the native dispatcher (`sdd-status --json`) for apply-progress / apply-state; fall back to a filesystem check for the `apply-progress.md` artifact.

- **Real work exists** → surface the transport failure but **do not latch**. The orchestrator runs the continuation, sees the progress, and relaunches a directed retry that merges apply-progress (the sdd-apply skill already supports merge-on-resume).
- **No work exists** → latch as before, protecting against blind relaunch over an empty change.

## Validation

- Unit tests: change-name extraction (JSON + plain), project-root preference, apply-progress detection, block decision — all pass.
- Integration (local OpenSpec change, strict TDD):
  - Normal success: unchanged behavior, valid result returned.
  - Work + forced-empty result: session **not** latched; a directed retry executed, merged apply-progress, and returned a valid result.
  - No-work + empty result: session latched (protection intact).
- `bun build` validates TS syntax; `go test ./internal/components/sdd/` passes.

## References

- Issue #2855 (occurrence comment documents the work-executed-but-result-lost case, distinct from the zero-execution occurrences).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after failed workflow phases.
  * Sessions can now be retried when partial progress remains, allowing continuation instead of being permanently blocked.

### DIFF
--- a/internal/assets/opencode/plugins/review-result-artifacts.ts
+++ b/internal/assets/opencode/plugins/review-result-artifacts.ts
@@ -241,6 +241,57 @@ function captureCwd(worktree: string | undefined, directory: string): string {
   return worktree || directory
 }
 
+// hasRealWork decides whether a failed SDD phase actually left work behind.
+// A phase can complete its files/tests/progress yet still return an empty task
+// result (transport/timing failure: the child ran, the result was lost). When
+// real work exists, latching the whole session is wrong: the orchestrator
+// should run the handoff's continuation (sdd-status), see the progress, and
+// relaunch a directed retry that merges the existing apply-progress. When no
+// work exists, the failure is genuine and the session must stay latched (no
+// blind relaunch over an empty change).
+async function hasRealWork(phase: string, cwd: string, prompt: unknown): Promise<boolean> {
+  // 1. Extract the change name from the task prompt, if present. Support both
+  //    the JSON binding form ("change":"<name>") and the plain form (change: <name>).
+  const text = typeof prompt === "string" ? prompt : ""
+  let change = ""
+  const jsonM = /"change"\s*:\s*"([A-Za-z0-9._-]+)"/i.exec(text)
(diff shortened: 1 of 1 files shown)


Alan-TheGentleman:
## Coordination: managed-plugin writer collision

#3062 writes `review-result-artifacts.ts`, the managed OpenCode plugin that #3138 D3 replaces transactionally with no coexistence or legacy fallback. Its session-latching behavior cannot be merged as another durable plugin path while that replacement is the approved owner.

Hold this PR rather than extending the plugin. If the reported recovery outcome remains required, re-scope it as a native Go-owned recovery concern in a <=400-line #3138 chain child. This comment makes no closure or review decision on #2855. `skill_resolution=paths-injected`

PR #3405, opened 2026-08-17
ci(windows): rebalance internal/cli shards by measured cost

### LINKED ISSUE #3404: ci(windows): cli-d-q shard exceeds the 1800s test timeout because shards are partitioned by letter, not cost
## Problem

The `cli-d-q` shard of the Windows Full Suite hits the 1800s `go test` timeout and panics, failing the whole run.

## Root cause

The shard matrix partitions `internal/cli` by test-name first letter (`a-c`, `d-q`, `r-a-m`, `r-n-z`, `s-z`). That tiling is blind to cost, and Go test names cluster by prefix while cost clusters with them: **423 of the 649 `TestR*` tests in `internal/cli` are `TestReview*`**, a single indivisible letter worth more than an entire shard's budget.

Measured `Run shard` step durations from run `32022500158`:

| shard | seconds |
|---|---|
| cli-d-q | **1818** (against a 1800s go-test timeout) |
| cli-r-a-m | 1692 |
| cli-s-z | 711 |
| cli-a-c | 295 |
| cli-r-n-z | 143 |

Two shards carry 77% of the package while two idle. `cli-d-q` is not slow because any one test is slow; the panic dump shows it mid-run on a 7s test. It simply ran out of budget.

## Not the cause

The `git clone … gentleman-guardian-angel` line that appears immediately before the panic is a red herring. It is a hardcoded fixture literal from a stubbed `runCommand` in `internal/cli/gga_install_fix_test.go`, printed to stderr by production error wrapping. No test performs a live clone: instrumenting `executeCommand` to panic on that argv and running the full `^Test[D-Q]` shard produces zero hits. The step runs without `-v`, so that single stub write was the only output in the run, which is why it looks adjacent to the panic.

## Expected behavior

Every shard completes inside its timeout, with headroom, and the coverage ratchet keeps proving that every test is claimed by exactly one shard.

## Notes

Raising the timeout is not the fix. It would leave the imbalance in place and simply move the ceiling.

Alan-TheGentleman:
Closes #3404.

## What changed

Shard selectors for `internal/cli` are now sized from measured cost instead of test-name first letter. Six shards replace five.

The `cli-d-q` shard was hitting the 1800s `go test` timeout and panicking. It was not slow because of any single test: the A-Z tiling is cost-blind, and `TestReview*` alone is 423 of the 649 `TestR*` tests in the package, one indivisible letter worth more than a shard's budget. Two shards carried 77% of the package while two idled.

## Coverage ratchet is now stronger

Free-form selectors cannot be checked by range algebra, so `TestWindowsFullSuiteShardsCoverEveryTestName` (`internal/assets/formatter_ordering_test.go`) now validates the matrix against each package's real `go test -list` inventory: every test claimed by exactly one shard, and no shard matching zero. That is strictly stronger than the range check it replaces, and it drops the hardcoded `TestR` partition pair the old check had to special-case.

Negative-tested: a dropped arm, an overlapping pair, and an empty shard each fail with a named test.

## Measurements

Linux seconds, back-to-back on one machine against the same test binary:

| before | s | after | s |
|---|---|---|---|
| cli-a-c | 13.9 | cli-a-m | 37.0 |
| **cli-d-q** | **91.0** | cli-n-o | **51.0** |
| cli-r-a-m | 83.1 | cli-p-q-s-z | 43.0 |
| cli-r-n-z | 10.6 | cli-review-a-m | 40.1 |
| cli-s-z | 29.9 | cli-review-n-z | 31.8 |
| | | cli-r-other | 15.4 |
| **total** | **228.5** | **total** | **218.3** |

All 1493 tests still run: 193+505+472+177+146 before, 378+228+238+235+187+227 after.

Windows tracks Linux at a steady ~20x across every shard in the matrix, and that model predicts the observed 1818s for `cli-d-q` from its 91.0s almost exactly. Projected worst shard after this change is ~1020s, or ~1220s at a pessimistic 24x: 57-68% of the 1800s timeout instead of 101%.

`reviewtransaction-a-c` (1087s) and `rest` (991s) are both comfortable and are left alone.

## Also

`internal/cli/gga_install_fix_test.go` now labels its stubbed failure `simulated install failure: …`. The old fixture text rendered a `git clone … gentleman-guardian-angel` line to stderr immediately before the panic, which reads exactly like a live network clone in CI. It is not one, and it never was: instrumenting `executeCommand` to panic on that argv and running the whole `^Test[D-Q]` shard produces zero hits. Only the log wording changed.

## Verification

| check | result |
|---|---|
| `gofmt -l internal/` | clean |
| `go vet ./...` | pass |
| `GOOS=windows go vet ./...` | pass |
| `go test ./internal/cli/ -count=1` | ok, 235.8s |
| `go test ./internal/assets/ -count=1` | ok, 5.4s |
| ratchet negative cases (gap / overlap / empty) | all three fail as designed |
| `./scripts/deadcode-ratchet.sh` | no new unreachable functions |

Windows Full Suite does not run on pull requests, so it is dispatched against this branch separately; that run is the real proof and is linked in a comment below.

## Summary by CodeRabbit

* **Tests**
  * Improved Windows CLI test sharding to use six cost-based partitions for more balanced execution times.
  * Added validation to ensure all tests are assigned to exactly one shard.
  * Improved install failure test diagnostics with clearer simulated-error output.

### DIFF
--- a/.github/workflows/windows-full-suite.yml
+++ b/.github/workflows/windows-full-suite.yml
@@ -43,9 +43,23 @@ jobs:
   #
   # Split by measured cost: internal/cli and internal/reviewtransaction are
   # ~44% and ~33% of total test time. internal/cli alone exceeded the 30m
-  # go-test timeout, so it takes five shards; TestR* dominates the package and
-  # is split again by its next letter. `rest` is derived from `go list` rather
-  # than listed, so a new package is covered the day it is added.
+  # go-test timeout, so it takes six shards. `rest` is derived from `go list`
+  # rather than listed, so a new package is covered the day it is added.
+  #
+  # The first cut of this split tiled A-Z by first letter, which is cost-blind:
+  # cli-d-q reached 1818s and cli-r-a-m 1692s against a 1800s go-test timeout
+  # while cli-r-n-z finished in 143s. Test names cluster by prefix, and cost
+  # clusters with them, so a first-letter tile cannot be balanced -- 423 of the
+  # 649 TestR* tests are TestReview*, a single indivisible letter. The shards
+  # below are sized from measured per-selector cost instead (Linux seconds,
+  # which Windows tracks at a steady ~20x across every shard in the matrix):
+  #
+  #   cli-a-m         33.5s   cli-review-a-m  41.1s
+  #   cli-n-o         46.1s   cli-review-n-z  34.9s
+  #   cli-p-q-s-z     48.6s   cli-r-other     14.3s
+  #
+  # That puts the worst shard near 1000s on Windows, roughly half the go-test
+  # timeout, instead of 1% over it.
   #
   # Intra-package t.Parallel() would be the deeper fix, but internal/cli has 89
   # os.Setenv/t.Setenv sites and t.Setenv panics in a parallel test, so that is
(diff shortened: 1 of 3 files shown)


Alan-TheGentleman:
Windows Full Suite dispatched against this branch: https://github.com/Gentleman-Programming/gentle-ai/actions/runs/32031664658

That workflow triggers only on push to `main`, the daily cron, and `workflow_dispatch`, so it produces no check on this PR. This run is the actual proof that the rebalanced shards complete inside the timeout, and it should be green before merge.

Issue #3440, opened 2026-08-18
review: the pi transport refusal never names the handshake that would fix it, and offers to disable review instead


Alan-TheGentleman:
## Problem

`gentle-ai review status --agent pi` from a plain shell refuses with a message that cannot be acted on, and recommends turning review off.

Reproduced on released `v2.4.0` in a throwaway repository:

```
code:  immutable_review_transport_unsupported
cause: the active runtime is not eligible for immutable receipt review;
       exit receipt-driven review with `gentle-ai review mode disable --scope clone --cwd <repo>`;
       supported immutable review runtimes: claude-code, opencode, codex
```

Exporting `GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1` makes the identical command work.

So the actual fix is one exported variable, and the message instead offers to disable receipt-driven development.

## Why the list can never name pi

`reviewTransportSupportedRuntimeIDs` (`internal/cli/review_transport_capability.go`) builds the list by calling `reviewImmutableRuntimeCapability` for each agent **in the current process**:

```go
for _, agent := range catalog.AllAgents() {
    if reviewImmutableRuntimeCapability(agent.ID).supportsImmutableReceiptReview() {
        supported = append(supported, string(agent.ID))
    }
}
```

And for `AgentPi` that function returns `Unsupported` when the handshake variable is absent (`:77`). The list is therefore computed under the exact condition that is failing, so `pi` can never appear in the message that would explain how to enable `pi`.

The eligibility rule itself is correct and deliberate: the relay's declared contract narrows the compiled boundary and must be a required conjunct. Only the diagnostic is wrong.

## Expected behavior

When the refused runtime is `pi` and the only missing condition is the handshake, the cause names `GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1` and does not offer `review mode disable` as the remedy. Disabling review is a valid exit from receipt-driven development; it is not the fix for a runtime that is one environment variable away from eligible.

## Impact

This is the shell-driven orchestrator path, which is the Pi use case: an agent driving `gentle-ai` through a terminal. A community report reached the conclusion that `v2.4.0` had dropped Pi support entirely, from exactly this message.

## Provenance

Split out of the community report Gentleman-Programming/gentle-pi#367 (defect B), whose analysis of this was accurate.

Alan-TheGentleman:
[...]

Issue #1997, opened 2026-07-29
fix(review): review status silently ignores --lineage, so the escalated abandon refusal names an unusable continuation


diegnghtmr:
## Pre-flight Checklist
- [x] I have searched [existing issues](https://github.com/Gentleman-Programming/gentle-ai/issues) and this is not a duplicate
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`

## 📝 Bug Description

v2.2.1 shipped "Refusals name a runnable continuation". The `review abandon --incomplete-inspection` refusal for an `escalated` lineage names one — and it cannot work, for three compounding reasons.

The printed command is not runnable as printed: it omits a flag the runtime then demands. The error that demands it names only `gentle-ai.review-integration/v1`, but `/v2` is accepted too, so the message misdirects. And once corrected with either contract, `review status --lineage` is a silent, unvalidated no-op: the output is byte-identical to the same command with no selector at all, a nonexistent lineage exits 0, and a string that is not lineage-shaped also exits 0.

So the refusal tells the operator "See where this review actually is with <command>", and that command can never answer the question — not because of the missing flag, but because the selector it depends on does nothing. Fixing only the flag would leave a runnable command that is still useless.

This is distinct from the closed #1540, which asked that native stop *advertising* an impossible abandon for escalated authority. Native does refuse correctly here. The defect is in the continuation hint that postdates that fix. It is also distinct from the open #1987, which reports `--lineage` ignored by `review start` (recovers a foreign lineage); this is `review status`, where the selector is ignored silently rather than resolving to the wrong thing.

## 🔄 Steps to Reproduce

Part A — the unusable continuation (needs an escalated lineage holding a fully captured lens plan):

1. Reach an `escalated` compact-v2 lineage whose selected plan captured every lens.
2. Run `gentle-ai review abandon --cwd <REPO> --lineage <LINEAGE> --expected-revision <REV> --actor x --reason x --incomplete-inspection=true --maintainer-authorization x`
3. Observe the refusal, which is correct, and copy the continuation command it prints.
4. Run that continuation verbatim. It exits 1.
5. Add `--contract gentle-ai.review-integration/v1` as the error demands. It exits 0 but never mentions the lineage or its `escalated` state.
6. Substitute `--contract gentle-ai.review-integration/v2`. It also exits 0, contradicting the error that named v1 as the requirement.

[...]

PR #3456, opened 2026-08-18
fix(review): preserve public contract identities in defect-report scrubber

### LINKED ISSUE #3443: review: the defect-report scrubber redacts the product's own contract identifiers, making its printed commands unrunnable
## Problem

The defect-report privacy scrubber redacts the product's own public contract identifiers, so guidance the product prints as an exact command cannot be run.

Reproduced on released `v2.4.0`, in a throwaway repository, from the ordinary `capture-result` preflight refusal:

```
gentle-ai review status --contract gentle-ai.review-integration<redacted> --next-transition
```

The message exists to tell the operator the exact invocation. As printed, that invocation is invalid: `gentle-ai.review-integration/v2` lost its version at the slash.

## Cause

Every refusal crosses `reviewScrubDefectReportField` on its way into the envelope's `cause` (`internal/cli/review_operation_contract.go:1132`). Two rules do the damage, measured against the real binary:

| input | output | rule |
| --- | --- | --- |
| `gentle-ai.review-integration/v2` | `gentle-ai.review-integration<redacted>` | `reviewDefectReportAbsolutePathRegexp` matches at **any** slash, not only a path root, despite its own comment |
| `GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1` | `<redacted>` | `reviewDefectReportEnvAssignmentRegexp`, the `KEY=VALUE` rule |

Neither value is private. Both are compile-time constants of this product, published in its own contracts and documentation.

## Why the scrubber is right in general

It exists so an automated defect report cannot carry a user's absolute paths, usernames, hostnames, or credentials into a public issue. That is correct and must stay. The bug is scope, not intent: a schema identity is not a filesystem path, and a documented contract constant is not a secret.

## Impact

- The negotiated entry point that this refusal names cannot be copied and run. Every operator who hits it has to already know the contract string, which is precisely what the message was trying to supply.
- It blocks #3440. That issue asks the pi transport refusal to name `GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1`; the value is destroyed on the way out, and an early draft produced `export <redacted>`, which is worse than the message it replaced. The variable name can be delivered today; its value cannot.

## Expected behavior

Public schema and contract identifiers the product itself emits survive scrubbing. Absolute paths, usernames, hostnames, credentials and genuine environment values continue to be redacted.

The narrow reading is that the path rule should anchor at a path root rather than matching any embedded slash, and that a `KEY=VALUE` whose value is a known product constant is not a secret. Whether that is the right implementation is open; the requirement is that the product's own instructions come out runnable.

## Note

A tripwire test is landing with the #3440 work (`TestPiRelayHandshakeGuidanceSurvivesTheFailureCausePrivacyGate`) that fails once the gate stops destroying the handshake, so this does not get forgotten.

Cobies:
## 🔗 Linked Issue

Closes #3443

## 🏷️ PR Type

## 📝 Summary

Preserves public contract and schema identities (such as `gentle-ai.review-integration/v2` and `gentle-pi.review-relay/v1`) and declared public contract environment variable assignments (`GENTLE_PI_REVIEW_RELAY_CONTRACT=gentle-pi.review-relay/v1`) from being incorrectly redacted by `reviewScrubDefectReportField`.

Ensures absolute paths (POSIX, Windows), emails, and sensitive environment assignments continue to be safely and strictly redacted.

Updates Pi handshake refusal guidance to provide the full actionable contract command.

## 📂 Changes

|-------------|-------------|
| `internal/cli/review_defect_report.go` | Anchored absolute path regex to path root delimiters and added allowlist for known public contract environment assignments |
| `internal/cli/review_transport_capability.go` | Updated Pi handshake guidance to name the full contract assignment value |
| `internal/cli/review_pi_handshake_refusal_test.go` | Updated Pi refusal cause and privacy gate test assertions to expect exact handshake string |
| `internal/cli/review_defect_report_test.go` | Added comprehensive test matrix verifying preservation of public contract IDs and redaction of secrets/paths |

## 🧪 Test Plan

**Unit Tests**
```bash
go test -v ./internal/cli -run "TestReviewDefectReport|TestPi"
```

**Go Format**
```bash
go run ./internal/gofmtcheck
```

- [x] Manually tested locally

## ✅ Contributor Checklist

- [x] PR stays within 400 changed lines, or I have requested/obtained maintainer-applied `size:exception` with rationale documented
- [ ] E2E tests pass (`cd e2e && ./docker-test.sh`)
- [ ] I have updated documentation if necessary

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic privacy scrubbing to redact secret values, email addresses, multiline content, and rooted Unix/Windows filesystem paths.
  * Preserved recognized public contract identifiers, commands, URLs, and approved environment settings in diagnostics.
  * Updated Pi handshake and relay refusal messages to include the required environment variable and exact relay contract value.
* **Tests**
  * Added coverage confirming sensitive information is redacted while valid public contract details remain visible.


Cobies:
Updated with full regression coverage addressing review feedback:

- POSIX paths following delimiters like `open:/path` and `{/path}` are cleanly redacted without affecting URL schemes (`http://`, `https://`, `file:///`).
- Environment variable assignments with trailing punctuation (such as `;`, `.`, `)`) isolate delimiters so allowlisted public contracts remain intact and executable while sensitive values stay redacted.

biggs-100:
Reviewed the full diff. The fix is solid, and the test table covers exactly the shapes that made this bug hard: path rules now anchor at a real path root, URL schemes are excluded by construction (the char after `://` is itself a slash, so the rule never fires), the allowlist compares the exact expected value so a user-set `GENTLE_PI_REVIEW_RELAY_CONTRACT=<something-else>` still redacts, and trailing punctuation survives for readability. Inverting the tripwire in `TestPiRelayHandshakeGuidanceSurvivesTheFailureCausePrivacyGate` is the right ending for that story.

Two forward-looking observations, from auditing this whole refusal class recently (#3443, #3440, #3557, and #1997 share the same root: printed guidance that cannot run as printed):

1. `reviewKnownPublicContractEnvAssignments` is a one-entry map maintained by hand. The next refusal that names a new public env contract gets silently destroyed again unless someone remembers to add it here. It may be worth deriving entries from the contract constants themselves, or leaving an explicit pointer at the guidance-construction sites toward this map.

2. Per-case tripwires prove today's strings survive the gate, but nothing fails CI when tomorrow's guidance carries an unallowlisted assignment or a placeholder-ridden command. #3595 proposes the missing half of this invariant: make the refusal-resolution ratchet enforce the same runnable-command grammar the benchmark already classifies. This PR is strong evidence for why that check wants machinery rather than memory.

Housekeeping note: CI is fully green except "Check PR Has type:* Label", which looks like it just needs the label applied.

Alan-TheGentleman:
Closing this in favor of #4206, which landed the same root fix today, with credit to this diagnosis of #3443. The reason is the path regex: anchoring it to a root delimiter stops redacting `~/.pi/agent/settings.json` and `file:///Users/x`, which the current regex redacts, and a public defect report cannot leak a home-rooted layout. #4206 keeps the broad path rule and protects only the product's own identifiers and the exact relay handshake, so the Pi guidance can spell the value out without that regression. The idea was right; thanks for pushing it.

PR #3460, opened 2026-08-18
fix(opencode): persist POSIX launcher profiles

### LINKED ISSUE #3043: feat(opencode): extend the SDD orchestrator with native background subagents
### Pre-flight Checklist

### 🔍 Affected Area

Other

OpenCode SDD orchestrator render/injection and OpenCode install/sync activation.

### 💡 Problem Statement

Gentle AI needs native OpenCode background subagents without creating a second SDD runtime architecture. Every CLI/runtime must follow one shared SDD orchestrator contract, while OpenCode receives a small provider-specific addendum for native task behavior and capability activation.

Users should see genuinely independent work run in the background when the installed OpenCode supports it. Dependency-sensitive, decision-sensitive, review-sensitive, or result-dependent work must remain foreground, with an explicit foreground fallback whenever background mode is unavailable or cannot be enabled safely.

Native background subagents shipped in OpenCode v1.14.51. The feature must not be described as beginning in v1.15.0.

#### Shared contract and OpenCode addendum

- Keep one canonical/shared SDD orchestrator contract for every CLI/runtime.
- During render/injection, give OpenCode that shared base plus one OpenCode-specific addendum. Do not maintain a divergent complete OpenCode orchestrator copy as the intended architecture.
- The addendum instructs the orchestrator to use native `task(..., background: true)` only for genuinely independent work; keep dependencies, user decisions, formal review/RDD actors, and result-dependent actions foreground; rely on completion notifications rather than sleep or polling; never duplicate work; and never run parallel writers in one worktree/branch unless explicit isolation is established.

#### Incremental and lossless migration

Migration is incremental and lossless. The issue-closing work first establishes one canonical composition/rendering seam and verifies byte-for-byte output preservation for existing runtimes. Default and named-profile OpenCode rendering must use that seam plus an additive OpenCode background-policy addendum injected exactly once. Kilocode must not inherit OpenCode-native background policy merely because it historically shared an asset path. Common fragments may be extracted later only with verified semantic/output parity. No migration slice may delete an existing runtime contract without an equivalent tested compatibility path, and the issue-closing PR does not require a big-bang rewrite of every CLI/runtime. The final issue-closing work still includes Gentle AI CLI/env activation, safe foreground fallback, tests/docs, and the required issue-linked `bench/` flow.

#### Install and sync control plane

Gentle AI owns the user-facing control plane:

- Both install and sync accept `--opencode-background-subagents=auto|on|off`.
- `GENTLE_AI_OPENCODE_BACKGROUND_SUBAGENTS=auto|on|off` provides the Gentle AI environment control.
- Precedence is CLI option > Gentle AI environment variable > prior Gentle AI-managed state > default.
- The default is `auto`: interactive installation offers activation; non-interactive install/sync remains foreground unless the CLI option, Gentle AI environment variable, or prior managed state explicitly enables it.
- `on` persistently configures future supported OpenCode CLI/TUI launches and enables the OpenCode background addendum.
- `off` makes Gentle AI dispatch no background work and removes only Gentle AI-managed activation. It must never erase or override external user-managed OpenCode environment; unrelated configuration remains untouched.
- Managed activation and rollback are idempotent and must report exactly what changed.

#### Upstream runtime activation and capability

- Effective OpenCode activation is `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`, or `OPENCODE_EXPERIMENTAL=true` only when the specific variable is absent. An explicit `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=false` wins, is an opt-out, and must not be silently overridden.
- There is no `opencode.json`/JSONC field or automatic `.env` loader for this feature. Do not invent either.
- The OpenCode server and embedded TUI worker must inherit the effective environment before startup. An attach-client environment cannot change an existing server; activation requires a restart.
- Correct specific-flag override semantics are available by v1.15.11. This is the minimum safe semantic baseline unless runtime capability proves equivalent behavior.
- `/experimental/capabilities` with `backgroundSubagents` exists from v1.17.9 and reports the effective server capability. Runtime tool schema/capability is authoritative; version is only a fallback.
- Unsupported or unknown runtimes, and unmanaged attached servers, fall back to foreground execution with actionable guidance. Never claim background mode is active when it is not.

#### Runtime safety and V1 lifecycle scope

Native task background jobs are process-local, non-durable, have no global concurrency limit, and provide no filesystem/worktree isolation. Gentle AI must impose a bounded concurrency policy, avoid duplicate work, never poll, keep formal review and dependent work foreground, and never run parallel writers in one worktree. Restart cancels/loses active background job state; the implementation must not claim recovery.

Automatic managed activation targets only OpenCode CLI/TUI launch paths that Gentle AI can safely manage. `serve`, `attach`, Desktop, and unmanaged daemon topologies are capability-detected and reported, but are not restarted or mutated unless Gentle AI demonstrably owns that lifecycle; otherwise they use the foreground fallback.

#### Explicit boundary

This is adapter-level installation and prompt composition. It is not another scheduler, queue, claim system, worktree authority, recovery ledger, or OpenCode V2 writer bridge, and it must not depend on [#1302](https://github.com/Gentleman-Programming/gentle-ai/issues/1302). #1302 remains a separate future/complementary bounded parallel-apply authority, not this issue's implementation.

Keep [#3026](https://github.com/Gentleman-Programming/gentle-ai/issues/3026) separate for managed non-SDD reviewer/worker roles. Do not restore legacy `background-agents.ts`; use native OpenCode task subagents.

#### Acceptance Criteria

- [ ] The issue-closing work establishes one canonical composition/rendering seam and proves byte-for-byte output preservation for existing runtimes. Default and named-profile OpenCode rendering uses that seam plus an additive OpenCode background-policy addendum injected exactly once; Kilocode does not inherit that policy merely because it historically shared an asset path.\n- [ ] The migration is incremental and lossless: this issue-closing PR does not require a big-bang rewrite of every CLI/runtime; no runtime introduces a divergent full orchestrator copy, no migration slice deletes an existing runtime contract without an equivalent tested compatibility path, and later common-fragment extraction requires verified semantic/output parity.
- [ ] Tests cover the complete CLI/env truth table for `auto|on|off`, the precedence CLI > Gentle AI environment > prior managed state > default, interactive activation offer, non-interactive foreground default, persistent `on`, idempotent managed activation, and `off` rollback of only Gentle AI-managed state.
- [ ] Tests cover existing `OPENCODE_EXPERIMENTAL=true`, explicit `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=false` winning over the umbrella flag, the specific/umbrella activation rule, absence of an `opencode.json`/JSONC field and automatic `.env` loader, pre-start server/TUI inheritance, and restart messaging.
- [ ] Tests cover the v1.15.11 semantic baseline, `/experimental/capabilities` schema gating from v1.17.9, authoritative runtime tool schema/capability detection, version fallback, and unsupported/unknown/unmanaged attached-server foreground fallback with actionable guidance.
- [ ] Generated behavior uses background tasks only for genuinely independent work; tests prove bounded concurrency, no duplicate dispatch, no polling, foreground handling for formal review and dependent work, and no parallel writers in one worktree.
- [ ] Tests and documentation make process-local/non-durable job behavior, restart cancellation/loss, lack of global concurrency/isolation, and the absence of recovery claims observable.
- [ ] Managed activation is limited to owned CLI/TUI launch paths; `serve`, `attach`, Desktop, and unmanaged daemon cases are detected/reported but not restarted or mutated without lifecycle ownership.
- [ ] The same PR adds a public/runtime benchmark flow under `bench/` that reproduces the issue through the public/runtime boundary and makes activation, capability gating, foreground fallback, safe dispatch, completion handling, and restart-loss semantics observable. It must include:

  ```text
  Journey.Source: https://github.com/Gentleman-Programming/gentle-ai/issues/3043
  ```

### 🔄 Alternatives Considered

- **Maintain a complete OpenCode-specific orchestrator asset.** Rejected because it would diverge from the canonical shared SDD contract.
- **Write a background-subagent setting into `opencode.json` or load it automatically from `.env`.** Rejected because neither is an upstream feature contract.
- **Add a scheduler, queue, worktree/claim authority, recovery ledger, or V2 writer bridge here.** Rejected as out of scope; #1302 remains a separate future/complementary bounded parallel-apply authority.
- **Run every task in the background or poll for completion.** Rejected because dependencies, decisions, review/RDD work, and result-dependent actions require foreground sequencing, while completion must be notification-driven.
- **Always run in the foreground.** Retained as the explicit safe fallback, not the supported-background path.

### 📎 Additional Context

- `internal/components/sdd/inject.go:sddOrchestratorAsset()` currently selects complete agent-specific assets: OpenCode maps to `opencode/sdd-orchestrator.md`, while the generic fallback maps to `generic/sdd-orchestrator.md`. The intended architecture is shared-base composition with an OpenCode addendum, not another full copy.
- Native OpenCode task behavior requires the effective process environment before the server/TUI starts; attaching with a changed client environment cannot alter an existing server.
- Capability detection and runtime tool/schema behavior take precedence over version-only assumptions. Unsupported or unmanaged topologies must remain foreground with actionable restart or upgrade guidance.

### LINKED ISSUE #3452: fix(opencode): persist POSIX launcher profiles safely
## Pre-flight Checklist

- [x] I have searched existing issues and this is not a duplicate.
- [x] I understand that PRs will be rejected if the linked issue does not have `status:approved`.

## Bug Description

Prerequisite slice 2 of the three-child/four-PR stack for parent #3043. POSIX zsh/bash launcher profiles need atomic managed-block persistence and transaction safety.

Parent: #3043
Original report: https://github.com/Gentleman-Programming/gentle-ai/issues/3043#issuecomment-5323530390

## Steps to Reproduce

1. Activate managed OpenCode background mode for a zsh or bash shell.
2. Use profiles with LF, CRLF, malformed managed markers, or a concurrent replacement.
3. Toggle activation on and off, including a later activation failure requiring rollback.

## Expected Behavior

The exact managed block is idempotent, preserves line endings and safe modes, rejects malformed ownership, and rolls back only files still owned by this transaction.

## Actual Behavior

The prerequisite slice delivers durable POSIX profile activation and its filesystem transaction protections.

## Environment

- Gentle AI Version: development tree based on `301fb2ad7f3f3bda71f516d6e2848ef3fa6fe9bb`
- Operating System: POSIX (macOS and Linux)
- AI Agent / Client: OpenCode
- Shell: zsh and bash
- Affected Area: Installation Pipeline

## Logs / Error Output

No external logs required; deterministic `t.TempDir()` regressions cover line endings, modes, markers, idempotency, and rollback.

## Additional Context

This is an approved carrier issue for PR 2 of 4. It depends on PR 1 and does not complete #3043.

decode2:
## Linked Issue

Closes #3452

---

## PR Type

- [ ] `type:feature` - New feature
- [ ] `type:docs` - Documentation only
- [ ] `type:refactor` - Code refactoring
- [ ] `type:chore` - Build, CI, or tooling changes
- [ ] `type:breaking-change` - Breaking change

---

## Summary

- Persists only marked POSIX login-profile PATH blocks and removes them reversibly.
- Preserves user content, permissions, CRLF bytes, malformed markers, and concurrent edits.

---

## Changes

|-------------|-------------|
| `internal/opencode/background.go` | POSIX profile transaction, snapshots, rollback, and effective-resolution reporting. |
| `internal/opencode/background_test.go` | Profile persistence, stale-write, CRLF, shell, and idempotency regressions. |
| `internal/components/uninstall/*` | Removes only the managed profile block. |

---

## AI Assistance

- [ ] **None** - No material AI assistance was used.
- [x] **Material assistance used** - Complete all applicable declaration fields below.

**Tool/model (if known):** OpenCode, openai/gpt-5.6-terra

**Material scope:** Implemented the approved POSIX profile persistence slice and tests under explicit maintainer direction.

**Verification performed:** Focused profile/uninstall tests, package tests, race tests, gofmt, diff checks, and frozen-candidate validation.

---

## Test Plan

**Unit Tests**
```bash
go test ./internal/opencode ./internal/components/uninstall -count=1
go test -race ./internal/opencode ./internal/components/uninstall -count=1
```

**Go Format**
```bash
gofmt -w internal/opencode/background.go internal/opencode/background_test.go internal/components/uninstall/service.go internal/components/uninstall/service_test.go
```

**E2E Tests**
Not run. Filesystem transaction coverage is in focused and package Go tests.

**Benchmark Validation**
N/A. The end-to-end journey is PR 4.

- [x] Unit tests pass (focused package suite)
- [x] Go format passes
- [ ] E2E tests pass
- [x] Manually tested locally

---

## Automated Checks

| Check PR Cognitive Load | Pending | 411 incremental lines, maintainer-approved size:exception. |
| Check Issue Reference | Pending | Closes approved issue #3452. |
| Check Issue Has `status:approved` | Pending | Issue was inspected before publication. |
| Check PR Has `type:*` Label | Pending | Exactly `type:bug` will be applied. |
| Unit Tests | Pending | Focused package and race tests passed locally. |
| Go Format | Pending | gofmt passed locally. |
| E2E Tests | Not run | Not applicable to this transaction boundary. |

---

## Contributor Checklist

- [x] PR stays within 400 changed lines, or I have requested/obtained maintainer-applied `size:exception` with rationale documented
- [x] Unit tests pass (focused slice package tests)
- [x] Go format passes
- [ ] E2E tests pass (not run)
- [x] Benchmark validation completed, or this change is not applicable to the benchmark (explained above)
- [x] I have updated documentation if necessary (not needed until PR 4)
- [x] I selected exactly one AI-assistance option and completed the declaration

---

## Chain Context

This is a prerequisite slice. It closes #3452 and does not complete #3043. Maintainer-approved `size:exception`: 411 incremental additions plus deletions is required to keep the complete POSIX profile transaction and its regressions together.

```text
main
  |- PR1 ownership
  |- PR2 profiles [current, draft]
  |- PR3 readiness (draft)
  `- PR4 lifecycle (draft, closes #3043)
```

All PRs target `main`. Merge and review order is PR1, PR2, PR3, PR4. Later PR diffs include predecessor commits because they also target `main`; review only the stated incremental boundary. Review mode: disabled/unmanaged.

Predecessor: PR1 (https://github.com/Gentleman-Programming/gentle-ai/pull/3459)

Successor: PR3 (https://github.com/Gentleman-Programming/gentle-ai/pull/3461)

Rollback boundary: revert only managed profile selection, block mutation, profile rollback, and profile-block uninstall cleanup.

---

## Notes for Reviewers

Review the exact marked-block contract and concurrent-edit preservation. No qualifying governance guard was added.

## Summary by CodeRabbit

- **New Features**
  - Improved OpenCode background activation detection across supported shells and platforms.
  - Added clearer activation statuses, including pending and effective, with actionable explanations.
  - Doctor checks now verify that the managed OpenCode launcher is active and correctly resolved.

- **Bug Fixes**
  - Prevented background execution from being enabled when activation is not yet effective.
  - Sync and dry-run results now accurately reflect activation readiness.
  - Uninstall removes only managed launchers and profile entries, preserving user-owned changes and settings.


Alan-TheGentleman (changes requested):
The profile persistence itself is careful work and I want to say that first: **there is no partial-write path**. Every profile write goes through `filemerge.WriteFileAtomic` — temp in the same directory, chmod, fsync, rename, read back, fsync parent. A crash or ENOSPC leaves the original bytes intact. Mode is preserved from the pre-image, a file that did not exist gets `0644` and is removed on rollback, CRLF and a missing trailing newline are handled, and repeated activation produces one block. I confirmed all of that live, including a byte-exact restore on deactivate. That is the hard part and you got it right.

The blocker is not the writing. It is the signal.

## D1 — CRITICAL. The effective-activation probe can never be false on POSIX

`internal/opencode/background.go:723-731`:

```go
} else if err := p.options.AddToUserPath(BinDir(p.homeDir)); err != nil { ... }
if p.goos != "windows" {
    p.effective = pathResolvesTo(os.Getenv("PATH"), POSIXLauncherPath(p.homeDir), p.goos)
    if p.effective { p.activationReason = "" }
}
```

`system.AddToUserPath` → `AddToUserPathWithResult` (`internal/system/path.go:44-49`) → `addToProcessPath` (`:203-213`), which is `os.Setenv("PATH", dir+":"+currentPath)`. That is the exact in-process mutation #3043 is about. The probe then reads `os.Getenv("PATH")` one line later, so the managed dir is guaranteed present and the launcher it looks for was just written by the same `Apply`. `p.effective` is unconditionally `true` after every successful POSIX activation, and `activationReason` is cleared.

Live, same shell, same second:

```
$ gentle-ai sync
OpenCode background runtime ready: true
OpenCode background activation status: ready

$ gentle-ai doctor
[xx] opencode:managed_activation  OpenCode background policy is on, but bare
     opencode does not resolve to managed launcher .../.gentle-ai/bin/opencode
Status: unhealthy
```

Unit-level, with the real `AddToUserPath`: `prepare: Effective=false` → `apply: Effective=true`.

**Why the suite does not catch it.** Every activation test passes `AddToUserPath: func(string) error { return nil }` (`testActivationOptions` in `background_test.go`). The stub never touches the process PATH, so the probe returns false and `TestPOSIXActivationPersistsAndRemovesOnlyOwnedLoginProfileBlock` asserts "pending" for the wrong reason. Nothing anywhere pins the post-`Apply` value.

**Damage.** `runtimeReady` and `BackgroundPolicyEnabled` go true (`run.go:245`, `sync.go:1638`), the foreground warning is suppressed (`run.go:240`), and sync prints ready. And `stagePlan()` computed `backgroundPolicy` from the **pre**-Apply value (`run.go:794`), so the agent config is written *without* the background policy while the result claims it is on.

That is #3043 restated with a green light on top, which is worse than `main` — `main` never claimed to have solved it.

**Fix**: capture PATH *before* `AddToUserPath` and probe against that, or better, share doctor's resolution instead of a second implementation.

## D2 — HIGH. `pathResolvesTo` ignores PATH precedence

`background.go:909-917` returns true if the managed dir appears anywhere in PATH, without checking whether an earlier entry also holds an `opencode`.

Live with `PATH=$HOME/bin:/usr/bin:/bin:$HOME/.gentle-ai/bin`: `command -v opencode` resolves to `$HOME/bin/opencode`, sync reports ready, doctor fails.

This is not hypothetical. `.zprofile` prepends the managed dir, then `.zshrc` runs `eval "$(brew shellenv)"`, which prepends `/opt/homebrew/bin` **after** it. That is the documented Homebrew setup on Apple Silicon zsh, which is the platform this issue was reported from.

## D4 — HIGH. A symlinked `~/.zprofile` aborts the entire install

`readLauncherSnapshot` (`:984`) is reused for profiles and routes through `ManagedLauncherOwnership`, which hard-errors on any non-regular file (`:1011-1013`). `addProfileChange` propagates it and `PrepareActivation` returns the error, which is fatal:

```
Error: prepare OpenCode background activation: inspect managed OpenCode launcher
  ".../.zprofile": refusing non-regular OpenCode launcher collision
```

Nothing installs. The symlink and its target are correctly untouched, but chezmoi, yadm, dotbot and stow all symlink dotfiles, and the message calls the user's login profile a "managed OpenCode launcher". This should skip activation with a reason, not abort the install.

## D5 — MEDIUM. A read-only profile is rewritten anyway

Atomic replace-by-rename only needs the directory's write bit. Live: the block was appended to a `0444` `.zprofile` and the mode preserved afterward. Not corruption, but `0444` is an explicit "do not modify" and it is silently overridden.

## D6 — MEDIUM. Blocks accrete and orphan, and uninstall says nothing

`hasProfileBlock` (`:948-950`) requires a byte-exact match of the whole block including the `binDir` export line. Live: install, move home, `sync` → **two** blocks. `uninstall --all --yes` removes only the current-`binDir` one and leaves the stale block permanently, with nothing in "Manual cleanup required".

A hand-edited block survives uninstall untouched, which is the safe direction, but also silently. `docs/platforms.md` saying deactivation removes "only Gentle AI's marked shell-profile block" is true and reads as complete.

Forward hazard: any future change to the block text strands every existing block and appends a new one.

## D8 — MEDIUM. The target misses the most common Linux setup

`.zprofile` and `.bash_profile` are login-shell files. Linux terminal emulators start non-login interactive shells, which read `.zshrc`/`.bashrc` and never source `.zprofile`. `loginProfile` (`:843-854`) also refuses bash unless `.bash_profile` already exists — Debian and Ubuntu ship only `~/.profile` — and refuses every other shell, fish included, which is the maintainer's own shell. Those users get an early return that writes nothing (`:695-698`) and a doctor that reports unhealthy forever.

Scoping to login shells may be a deliberate first step. If so, say it in the PR and the docs, because the titles currently read as a complete fix for #3043.

## D10 — LOW

`internal/components/uninstall/service.go:411-416` unconditionally snapshots both `~/.zprofile` and `~/.bash_profile` into the backup store, including a `.bash_profile` Gentle AI never touched. Login profiles routinely hold tokens.

---

D1 is the one that has to change before this can land. The rest are worth fixing but would not block on their own.

decode2:
Addressed the activation-probe review on head `34345fbe`: activation and doctor now share one fail-closed resolver over the pre-mutation PATH snapshot, profile writes roll back even when a write lands before returning an error, and install correctly reports pending until a fresh shell resolves the managed launcher.

Focused, race, benchmark, portability, and deadcode checks pass locally. Current GitHub checks are queued. @Alan-TheGentleman, could you re-review the updated head?

decode2:
The validated #3459 foundation is now an exact ancestor of #3460 on head `0e60d09c`.

The four historical overlaps were resolved semantically, preserving current-main benchmark/app contracts, #3459 launcher ownership and rollback behavior, and #3460 profile persistence, pre-mutation PATH snapshot, first-executable activation check, profile safety, convergence, and rollback coverage. Focused app/OpenCode/uninstall tests, race tests, full bench tests, gofmt, and diff checks all pass.

D10 remains the explicitly LOW, non-blocking concern about backing up unmanaged profile files; I did not broaden this chain repair into a new backup-selection/data-handling change. Hosted checks are running. The PR remains draft while CI and fresh maintainer review complete, and this head is now the required parent for #3461.

decode2:
All 16 hosted checks are now green and GitHub reports `CLEAN/MERGEABLE`, so this PR has been marked ready for review.

Merge order remains strict: #3459 first, then this PR, followed by #3461 and #3462.

PR #3520, opened 2026-08-20
fix(review): distinguish unachievable lens attempts from unattempted slots

### LINKED ISSUE #3442: review: a host cannot report a lens slot as unachievable, so STATUS re-offers it and the two layers contradict each other
## Problem

A host that discovers a lens slot cannot be completed has no way to say so. The only thing it can do is fail, and negotiated STATUS then re-offers the identical slot, because from Go's side the slot is simply outstanding.

`reviewMissingCaptureTransition` (`internal/cli/review_next_transition.go:548-567`) derives the offer from captured artifacts alone:

```go
captured := make(map[int]bool, len(artifacts))
for _, artifact := range artifacts {
    captured[artifact.SelectedOrder] = true
}
for order, lens := range selectedLenses {
    if !captured[order] {
        inputs = append(inputs, reviewCaptureInput(...))
    }
}
```

That is correct as far as it goes, and it correctly re-offers only outstanding slots. The gap is that "outstanding" and "achievable" are the same thing here, and they are not the same thing in the field.

## The case that surfaced it

Gentleman-Programming/gentle-pi#367. A reviewer subprocess exceeded the host relay's bound on a 1.58 MB lens prompt. The relay reported a transport failure, STATUS re-offered the exact same slot, and the sanctioned continuation was to relaunch it unchanged, which re-spent reviewer model tokens to hit the same wall.

gentle-pi has now fixed its side (Gentleman-Programming/gentle-pi#368): the bound scales with prompt bytes, the failure carries elapsed and limit, and its own `next_action` opens with "Do not relaunch this slot unchanged". But that guidance lives entirely in the host's envelope. Nothing reaches Go, so STATUS keeps offering the slot as though nothing had happened, and the two continuations contradict each other.

## Expected behavior

A host can report that a bound slot was attempted and is not achievable in the current conditions, with its evidence, and the negotiated transition reflects that rather than re-offering the slot as if untried.

What "reflects that" means is the design question. It could be a stop with an actionable reason code, an offer annotated with the prior attempt, or a distinct collection state. What it should not be is what happens today: two layers giving the operator opposite instructions about the same slot.

## Scope note

This is not about making Go trust a host's claim. A host asserting "unachievable" must not be able to skip a lens or fabricate a result. The narrow ask is that an attempted-and-failed slot is distinguishable from a never-attempted one.

## Related

- #3384 stop transitions emit a bare reason code, so consumers invent their own menus. Same family: the transition vocabulary cannot express what the consumer needs to act on.
- Per-slot admission already works correctly: a lens admitted before the failure stays admitted, and only the outstanding slots are re-offered. Verified at `:548-567`. That half is not the problem.

Cobies:
Distinguish attempted-and-failed/unachievable lens slots from never-attempted slots in review transitions so STATUS does not re-offer failed slots.

Closes #3442

## Summary by CodeRabbit

* **Bug Fixes**
  * Review finalization now proceeds only when all selected review lenses have admitted artifacts.
  * Review routing distinguishes untried, admitted, unverifiable, and unachievable attempts.
  * Unachievable attempts are retried up to three times before producing a terminal stop.
  * Reviewer and admission failures are retained for diagnostics without being treated as completed results.
  * Unverifiable artifacts continue to use the `captured_artifacts_unverifiable` status.

* **Documentation**
  * Added guidance for handling the `unachievable_reviewer_attempt` stop reason.


decode2 (changes requested):
Requesting changes at head `2159483e8ed50fabd8f5704e00def39d3efbad3e`.

`ArtifactAdmissionUnachievable` currently has no production producer, validated admission, or durable discovery path. The focused test passes only because it manually injects a state that the host/provider flow cannot create. Provider refusal currently returns without recording the bound attempt, so STATUS can re-offer the lens without retry accounting.

Please add validated provider evidence bound to lineage, target, lens, and selected order; persist and discover the attempted slot without admitting it as a reviewer result; and add retry accounting. STATUS needs a distinct truthful operational reason and continuation for an unachievable attempt, rather than `captured_artifacts_unverifiable`, which incorrectly suggests captured-authority corruption. Keep finalize and receipt issuance blocked and add a black-box host/provider regression. The PR also lacks the required `type:*` label.

decode2 (changes requested):
Thanks for adding the production path for unachievable attempts. Two correctness gaps remain:

- `review_artifact.go` discards attempt-persistence errors. An empty reviewer response is then rejected by `compact_reviewer_capture.go`, leaving no durable attempt and allowing STATUS to reoffer the slot.
- `CaptureUnachievableReviewerAttempt` accepts `req.StoreDir` independently of the authoritative store directory.

Please make every failed attempt durable, propagate persistence failures, and bind attempt storage to the authoritative store. The failing refusal-ratchet unit test also needs resolution.

PR #3694, opened 2026-08-24
fix(app): surface corrupt state read errors in persistAssignments

mpazminoecl:
## Summary

`persistAssignments` silently swallowed read errors from `state.Read` when `state.json` existed but was unreadable (e.g. invalid JSON): it returned `nil`, so the caller's new model assignments were dropped without any warning to the user.

## Fix

- The fail-safe invariant is **preserved**: corrupt state is never overwritten.
- Read failures other than `os.ErrNotExist` are now wrapped (`read existing state: %w`) and returned, so the sync flow fails with a clear `persist model assignments: ...` message instead of losing assignments silently.
- `TestPersistAssignmentsSkipsCorruptState` now also asserts that the error is propagated (it previously ignored the return value).

## Verification

- `go build ./internal/app/` ✅
- `go test ./internal/app/ -count=1` — full package suite passes ✅
- `gofmt -l` clean on both touched files ✅

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of invalid or unreadable saved state files.
  - Prevents existing state from being overwritten when the saved data cannot be read.
  - Reports an error so failures are visible instead of being silently ignored.

- **Tests**
  - Added validation to ensure corrupt state data produces an appropriate error.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -912,9 +912,11 @@ func persistAssignments(homeDir string, selection model.Selection) error {
 	current, err := state.Read(homeDir)
 	if err != nil {
 		// State file may not exist yet (e.g. pre-state users). Other read
-		// failures, such as invalid JSON, must not overwrite existing state.
+		// failures, such as invalid JSON, must not overwrite existing
+		// state — and must not be swallowed either, otherwise the caller's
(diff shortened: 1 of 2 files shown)


mpazminoecl:
Closing and reopening from my personal account.

PR #4125, opened 2026-09-04
test(e2e): keep OpenCode 4R bootstrap egress off the reviewer barrier

### LINKED ISSUE #4121: bug(e2e): OpenCode 4R journey races a live npm fetch against a hardcoded 30s barrier
### Pre-flight Checklist

### 📝 Bug Description

`TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently` (`e2e/organicruntime/organic_runtime_test.go:977`) is flaky because it races OpenCode's **live npm registry bootstrap** against a **hardcoded 30-second barrier**.

The test config declares its provider with an unpinned npm package:

```go
"provider": {"loopback": {"npm": "@ai-sdk/openai-compatible", ...}}
```

and the transport plugin imports `@opencode-ai/plugin`. Meanwhile `organicEnvironment` (`:2988`) points `HOME` at a `t.TempDir()` and sets no `npm_config_cache` / `BUN_INSTALL`. Since the npm cache lives under `$HOME`, **every run starts cold** and OpenCode must fetch from the network before it can issue its first model call. The barrier at `:1228` gives it 30 seconds.

Root cause confirmed by X-raying the hung process 20s into a failure: it held **15 ESTABLISHED TLS connections to external hosts and zero to the loopback provider** — 14 to `104.16.{0,1,3}.34` (all in the A-record set of `registry.npmjs.org`) and 1 to `172.66.173.149` (`api.opencode.ai`). CPU was 0.4% with the main thread parked in `kevent64`: idle network wait, not a spin.

Two secondary defects make this far harder to diagnose than it should be:

1. **The failure message names the wrong component.** It reports `OpenCode did not schedule all four foreground reviewer requests`, but instrumentation shows `arrived=0` / `titleAnswered=false` — OpenCode never issued a single request, so the 4R scheduling path never executed. The message sends readers hunting a concurrency bug that does not exist.
2. **The timeout branch emits no diagnostics.** It is a bare `t.Fatal` that reports no `arrived`, `arrivals`, `maxInFlight`, `handlerFailure`, or subprocess output — unlike its sibling `case result := <-run:` branch, which prints both error and output. From CI logs alone the failure is undiagnosable.

Note the contrast with the sibling test at `:888`: it also writes the same plugin and launches OpenCode, but passes in ~7-8s because it runs behind a **loopback denying proxy**, which fails external fetches fast instead of hanging on them. The mitigation already exists in this suite; it is simply not applied to this test.

### 🔄 Steps to Reproduce

1. Install the pinned runtime: `npm install --global opencode-ai@1.18.10` (matches `internal/versions/versions.go:17`).
2. Run the test in isolation, repeatedly:
   `GENTLE_AI_OPENCODE_RUNTIME_E2E=1 go test ./e2e/organicruntime -run '^TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently$' -count=1`
3. Repeat 10+ times. Saturating all CPU cores during the run makes failures markedly more likely.

Measured locally: **7 FAIL / 6 PASS over 13 runs**, strictly bimodal — PASS ~12s, FAIL pinned at the 30s wall. Three consecutive passes on an idle machine; failure on the first run after saturating all cores.

### ✅ Expected Behavior

The test should assert 4R reviewer scheduling deterministically, without depending on live npm registry latency. Its outcome should not vary with network speed or machine load, and when it does fail it should report enough state to identify which stage failed.

### ❌ Actual Behavior

OpenCode spends the entire 30-second budget fetching npm dependencies over the network, never contacts the loopback provider, and the test fails claiming a 4R scheduling failure that never occurred. The subprocess stays alive throughout — `CombinedOutput` returns zero bytes — so the `case result := <-run:` branch never fires.

Observed in CI on PR #4102 (a documentation-only change that cannot reach this code path: the harness builds a temp repo and temp `HOME`, and `docs/` is excluded from the `go:embed` tree, as noted in `internal/cli/review_next_transition_docs_test.go:83`). The same job passed on four earlier heads of that branch and failed on three consecutive attempts at the fifth, whose only delta is 20 deleted lines of `README.md`.

### Gentle AI Version

gga v2.10.1

### Operating System

macOS

### AI Agent / Client

Claude Code

### 📋 Affected Area

Other

### 💡 Logs / Error Output

```shell
# CI (ubuntu-latest), PR #4102:
organic_runtime_test.go:1228: OpenCode did not schedule all four foreground reviewer requests
--- FAIL: TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently (31.13s)

# Local run with the timeout branch temporarily instrumented:
DIAG arrived=0 inFlight=0 maxInFlight=0 settled=0 arrivals=map[] \
     titleAnswered=false tasksIssued=false handlerFailure=""
OPENCODE OUTPUT:
(empty)
ERR: signal: killed

# lsof on the hung process, 20s in (local IP redacted):
PID   STAT %CPU %MEM ELAPSED
31310 S     0.4  5.9   00:20

TCP <local>:57262->104.16.1.34:443    (ESTABLISHED)   # registry.npmjs.org
TCP <local>:57259->172.66.173.149:443 (ESTABLISHED)   # api.opencode.ai
TCP <local>:57294->104.16.1.34:443    (ESTABLISHED)   # registry.npmjs.org
... 15 external connections total, 0 to 127.0.0.1

# sample(1) main thread:
2619 kevent64 (in libsystem_kernel.dylib)   # idle network wait
```

### Additional Context

Suggested fixes, in order of value:

1. **Remove the live network dependency.** Either pre-warm/pin the npm dependencies for this test, or run it behind the same loopback denying proxy the sibling test at `:888` already uses.
2. **Fix the misattributing message.** It should not name 4R scheduling when no request was ever issued.
3. **Make the timeout branch diagnosable.** Dump `arrived`, `arrivals`, `maxInFlight`, `handlerFailure`, and the subprocess output, matching what the `<-run` branch already does.

The test file is unchanged since `cd4ab336` (2026-08-31), so this affects `main` today and is not specific to any one PR. On 2026-09-03 a sibling live-network test, `TestOpenCodeRuntimeIsPinnedForTheLiveProviderTransport`, also failed once on an unrelated branch — consistent with the same fragility class.

Denver2828:
## Linked Issue

Closes #4121

---

## PR Type

---

## Summary

`TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently` starts OpenCode 1.18.10 against a fresh HOME. Before the first model request, OpenCode bootstraps `@opencode-ai/plugin@1.18.10` into every config directory (the temp HOME's `~/.config/opencode` and `OPENCODE_CONFIG_DIR`) through npm Arborist and refreshes models.dev; because the config directory carries an external plugin, plugin loading waits for that install. When the egress outlasts the 30 s barrier the test fails without ever reaching the loopback provider, the message blames 4R scheduling, and nothing is printed because the subprocess is still alive.

This PR keeps the accepted scope from the issue thread:

- reuses the loopback-denying proxy from the pinned transport test (new `loopbackDenyingProxyEnvironment` helper) so the bootstrap egress is refused at once, and sets `OPENCODE_DISABLE_MODELS_FETCH=1`; the transport plugin's `@opencode-ai/plugin` import is type-only, so the refused install has no effect on the session;
- keeps the 30 s barrier and the four-reviewer assertions unchanged;
- on the deadline, cancels the subprocess (through the shared `organicCommandContext`, which bounds the wait) and reports the barrier state (`tasksIssued`, `arrived`, `arrivals`, `maxInFlight`, `settled`, `handlerFailure`, denied egress count) with the collected output under a stage-neutral message; the early-exit branch reports the same state.

Measured on two Linux hosts with opencode 1.18.10: base fails at about 31 s while opencode holds dozens of concurrent TLS sockets to registry.npmjs.org (28 to 57 across runs on two hosts) plus one to api.opencode.ai, and none to models.dev; with this change the test passes in a few seconds (3.9 s on one host, 6.6 to 7.0 s on another), well inside the 30 s barrier, with zero external connections. Forcing a stall (never-answering proxy) now yields the new message with `process=signal: killed` and `tasksIssued=false arrived=0/4`.

---

## Changes

|-------------|-------------|
| `e2e/organicruntime/organic_runtime_test.go` | `loopbackDenyingProxyEnvironment` helper; 4R test runs OpenCode behind the loopback-denying proxy with `OPENCODE_DISABLE_MODELS_FETCH=1`; timeout branch cancels the subprocess and reports barrier state plus output with a stage-neutral message |

---

## AI Assistance

**Tool/model (if known):** Claude Code (Claude, model claude-fable-5-1)

**Material scope:** investigation of the OpenCode 1.18.10 startup path (Arborist install and models.dev refresh), the test change in `organic_runtime_test.go`, and the local verification runs.

**Verification performed:** commands in the Test Plan, executed on Linux (WSL2) with the pinned opencode 1.18.10; RED on `adf47dac` under natural conditions (five runs on the author host plus one on the audit host, all FAIL at about 31 s), GREEN on this branch on both hosts (all PASS in a few seconds with no external sockets), and a forced-stall run showing the new diagnostics.

---

## Test Plan

**Unit Tests**
```bash
go test ./e2e/organicruntime -count=1
```

**Go Format**
```bash
go run ./internal/gofmtcheck
```

**Targeted E2E (Linux, pinned opencode 1.18.10)**
```bash
GENTLE_AI_OPENCODE_RUNTIME_E2E=1 go test ./e2e/organicruntime -run 'TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently$' -count=1 -v
```

**Benchmark Validation:** N/A — e2e harness change only; no review-lifecycle, gate, recovery, delivery, or benchmark code touched.

- [x] Unit tests pass (`go test ./e2e/organicruntime -count=1`: ok, 23 PASS / 6 env-gated SKIP)
- [ ] E2E tests pass (`cd e2e && ./docker-test.sh`) — not run locally (no Docker on the verification host); relying on CI
- [x] Manually tested locally: targeted test PASS in 3.9 s on this branch versus FAIL at 30.85 s on `adf47dac`; `TestOpenCodeRuntimeIsPinnedForTheLiveProviderTransport` could not run locally (no strace) and is unchanged

---

## Notes for Reviewers

The proxy makes the two Arborist installs fail immediately; OpenCode logs `NpmInstallFailedError (cause: rF: 403 Forbidden - GET https://registry.npmjs.org/@opencode-ai%2fplugin)` twice per run (once per config directory) in its own log under the temp HOME, not in the output the test collects; the run then proceeds because the plugin never needs the package at runtime. The sibling transport test keeps its inline proxy block; it can adopt the helper in a follow-up if desired.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end coverage for concurrent bound-reviewer workflows.
  * Added validation that runtime operations behave correctly when outbound network access is restricted.
  * Enhanced failure reporting for timed-out workflows, including captured process output and reviewer activity details.
  * Prevented unnecessary model metadata refreshes during the test scenario.

### DIFF
--- a/e2e/organicruntime/organic_runtime_test.go
+++ b/e2e/organicruntime/organic_runtime_test.go
@@ -631,6 +631,22 @@ func (proxy *loopbackDenyingProxy) deniedRequests() int {
 	return proxy.requests
 }
 
+// loopbackDenyingProxyEnvironment routes every proxied protocol of a child
+// process through the loopback-denying proxy while loopback stays direct, so
+// an external request is refused instead of reaching the host network.
+func loopbackDenyingProxyEnvironment(proxy *loopbackDenyingProxy) []string {
+	return []string{
+		"HTTP_PROXY=" + proxy.URL,
+		"HTTPS_PROXY=" + proxy.URL,
+		"ALL_PROXY=" + proxy.URL,
+		"http_proxy=" + proxy.URL,
+		"https_proxy=" + proxy.URL,
+		"all_proxy=" + proxy.URL,
+		"NO_PROXY=127.0.0.1,localhost,::1",
+		"no_proxy=127.0.0.1,localhost,::1",
+	}
+}
+
 func denyingProxyTargetIsExternal(request *http.Request) bool {
 	host := request.URL.Hostname()
 	if host == "" {
@@ -1169,6 +1185,8 @@ func TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently(t *testing.T) {
 		writeOpenCodeChatText(writer, "review task group complete")
 	}))
 	defer server.Close()
+	proxy := newLoopbackDenyingProxy(t)
+	defer proxy.Close()
 
 	configDirectory := t.TempDir()
 	pluginDirectory := filepath.Join(configDirectory, "plugins")
@@ -1203,13 +1221,21 @@ func TestOpenCodeRuntimeRunsFourBoundReviewersConcurrently(t *testing.T) {
 	}
 	ctx, cancel := context.WithTimeout(t.Context(), organicAgentTimeout)
 	defer cancel()
-	command := exec.CommandContext(ctx, binary, "run", "--format", "json", "--dir", harness.repo.worktree, "--model", "loopback/loopback", "start the grouped Go-bound reviewer tasks")
+	command := organicCommandContext(ctx, binary, "run", "--format", "json", "--dir", harness.repo.worktree, "--model", "loopback/loopback", "start the grouped Go-bound reviewer tasks")
 	command.Dir = harness.repo.worktree
 	command.Env = append(harness.environment(),
 		"OPENCODE_CONFIG_DIR="+configDirectory,
 		"OPENCODE_CONFIG_CONTENT="+string(config),
 		"PATH="+filepath.Dir(organicBinary)+string(os.PathListSeparator)+os.Getenv("PATH"),
+		// OpenCode startup bootstraps @opencode-ai/plugin into every config
+		// directory through npm Arborist and refreshes models.dev before it
+		// issues the first model request. Against the fresh HOME, that egress can
+		// outlast the barrier, so the loopback proxy refuses it and the models.dev
+		// refresh is switched off. The transport plugin's import of
+		// @opencode-ai/plugin is type-only, so the refused install costs nothing.
+		"OPENCODE_DISABLE_MODELS_FETCH=1",
 	)
+	command.Env = append(command.Env, loopbackDenyingProxyEnvironment(proxy)...)
 	type runResult struct {
 		output []byte
 		err    error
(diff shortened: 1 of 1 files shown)


egdev6:
Reopening this work as #4861, carrying your change with your authorship intact. This pull request was closed on 2026-09-20, and its fork is gone, so it cannot be reopened. The diff is unmodified: proxy-based egress denial, models.dev fetch off, the thirty-second barrier and the four-reviewer assertions kept, cancellation and output collection on timeout with the barrier state, and a stage-neutral message.

PR #4438, opened 2026-09-10
fix(review): anchor managed-assets continuation to the invoking executable

### LINKED ISSUE #4434: bug(review): managed-assets continuation can invoke a different Gentle AI binary
### Pre-flight Checklist

### 📝 Bug Description

The negotiated `managed_assets_outdated` continuation can invoke a different Gentle AI binary from the one that produced the refusal.

`gentle-pi` intentionally runs its integrity-pinned package-local Gentle AI binary by absolute path. When a newer global `gentle-ai` is first on `PATH`, the provider-owned continuation `gentle-ai sync --agent pi` runs that global binary instead. The global binary records its own embedded managed-asset digest, so the package-local binary remains stale and returns the same continuation indefinitely.

The digest comparison is behaving correctly. The non-converging behavior comes from `managedAssetsContinuation` hard-coding an unqualified executable name while describing the command as the exact runnable recovery.

### 🔄 Steps to Reproduce

1. Install or run `gentle-pi` with its package-local Gentle AI v2.7.0 binary.
2. Put a different Gentle AI build on `PATH`; in the reproduced environment this was `v2.7.1-0.20260910030252-7d3eec9c5c24`.
3. Run `gentle-ai sync --agent pi`, which records the global binary's managed-asset digest.
4. Invoke negotiated review STATUS with the package-local binary:
   `review status --contract gentle-ai.review-integration/v2 --cwd <repo> --projection workspace --agent pi --next-transition`.
5. Observe `next_transition.kind=stop`, `reason_code=managed_assets_outdated`, and continuation `gentle-ai sync --agent pi`.
6. Run that continuation normally through `PATH`, then repeat step 4.

### ✅ Expected Behavior

The provider-owned continuation should invoke the same Gentle AI executable that detected the stale assets, or otherwise carry enough executable identity to guarantee convergence. After running the offered continuation, the same package-local STATUS should advance to `fresh_target_ready`.

### ❌ Actual Behavior

The unqualified continuation resolves through `PATH` to the different global binary. That binary considers its own digest current, while the package-local binary still reports `managed_assets_outdated`. Repeating the offered continuation cannot clear the refusal.

### Gentle AI Version

Package-local provider: v2.7.0; global/PATH sync binary: v2.7.1-0.20260910030252-7d3eec9c5c24

### Operating System

Windows

### AI Agent / Client

Other — Pi through `gentle-pi`

### 📋 Affected Area

CLI (commands, flags)

### 💡 Logs / Error Output

```shell
Package-local .gentle-ai/v2.7.0 STATUS:
  applicability=unrelated
  next_transition.kind=stop
  reason_code=managed_assets_outdated
  continuation.command="gentle-ai sync --agent pi"
  stale_assets=["sha256:fc3394c6d24ae4a8e27c595aa6202ff59c6f860f241c5b3317fdc36b0c172650"]

Global go/bin v2.7.1 pseudo-version STATUS for the same target/projection:
  applicability=unrelated
  next_transition.kind=execute
  reason_code=fresh_target_ready

~/.gentle-ai/state.json:
  installed_binary_version=2.7.1-0.20260910030252-7d3eec9c5c24
  managed_asset_digest=sha256:fc3394c6d24ae4a8e27c595aa6202ff59c6f860f241c5b3317fdc36b0c172650
```

### Additional Context

`internal/cli/review_asset_provenance.go` correctly defines stale as a diagnosed installed/expected digest mismatch. `internal/cli/review_operation_contract.go` constructs the continuation independently of the invoking executable.

`gentle-pi/lib/native-review-cli.ts` deliberately resolves an absolute package-local binary and relays the provider command verbatim, so rewriting the continuation in Pi would violate the provider boundary.

Related issues cover adjacent symptoms but not this executable-identity invariant: #4379, #4271, #3857, #3848, #3299, #4170, and #3049.

Suggested regression: `TestManagedAssetsContinuationUsesInvokingExecutable`, proving that a STATUS/START refusal produced by a package-local binary offers a continuation that targets that same executable even when `PATH` resolves `gentle-ai` to another version.

FranklinF25:
## 🔗 Linked Issue

Closes #4434

---

## 🏷️ PR Type

---

## 📝 Summary

The `managed_assets_outdated` continuation rendered an unqualified `gentle-ai sync --agent <id>` command while describing it as the exact runnable recovery. When a different Gentle AI binary is first on `PATH`, running that continuation executes the `PATH` binary instead, which records its own embedded managed-asset digest, so the binary that produced the refusal keeps refusing: repeating the exact advertised recovery could never converge.

`managedAssetsContinuation` now anchors the command to the binary that diagnosed the skew, resolved through `os.Executable()`:

- The executable token is quoted (with `\"` escapes) only when the path contains whitespace or a double quote, so the command stays literally runnable in POSIX shells, cmd.exe, and PowerShell alike.
- When the executable cannot be resolved, the historical bare `gentle-ai sync` form is kept rather than guessing a path that cannot be proven.
- After running the offered continuation, the same invoking binary's STATUS advances to `fresh_target_ready` — the convergence invariant the issue names.

---

## 📂 Changes

| `internal/cli/review_operation_contract.go` | `managedAssetsContinuation` anchors to the invoking executable; new `reviewManagedAssetsExecutablePath` var, `managedAssetsExecutableToken`, `managedAssetsContinuationExecutable` helpers; failure `Validate()` now enforces the widened command shape via `validManagedAssetsContinuationCommand` |
| `internal/cli/review_asset_provenance_test.go` | New regression `TestManagedAssetsContinuationUsesInvokingExecutable` (quoting table + end-to-end STATUS proof against the published schema); the two envelope assertions now derive the expected command from the invoking executable |
| `contracts/review-integration/v2/schemas/failure.schema.json` | `managed_assets_continuation.command` pattern widened to admit the bare `gentle-ai` fallback or an invoking-executable path (quoted when it contains whitespace); every previously valid envelope still validates |
| `docs/review-integration.md` | `managed_assets_outdated` row now states the continuation is anchored to the executable that reported the stale assets |

---

## 🤖 AI Assistance

Select exactly one option. Do not check both options.

**Tool/model (if known):** Pi coding agent (el Gentleman harness) driving GLM-5.3

**Material scope:** Root-cause analysis from the issue report, implementation of the executable-anchored continuation, schema pattern widening, regression and envelope test updates, docs row update.

**Verification performed:** RED→GREEN demonstration of the regression test; focused local runs of the managed-assets and full `Review`-filtered suites in `internal/cli` (93s) and `internal/reviewtransaction` (100s); `go run ./internal/gofmtcheck`; `go build ./...`; schema regex case matrix (back-compat + new forms + rejection of malformed commands). All final verification completed locally: E2E Docker suite passed on all three platforms (ubuntu, arch, fedora — 3/3, 79 pass / 0 fail); remaining authority (fork CI) still runs on the PR.

---

## 🧪 Test Plan

**Unit Tests**
```bash
go test ./...
```

**Go Format**
```bash
go run ./internal/gofmtcheck
```

**E2E Tests** (Docker required)
```bash
cd e2e && ./docker-test.sh
```

**Benchmark Validation**

N/A — no benchmark implementation, corpus, classifier, or claim changed; this fix only alters the executable token rendered inside the existing managed-assets continuation contract.

- [x] E2E tests pass (`cd e2e && ./docker-test.sh`) — ran locally: 3/3 platforms PASSED (ubuntu, arch, fedora)
- [x] Manually tested locally (RED→GREEN: before the fix the STATUS stop offered `gentle-ai sync --agent opencode`; after it, the command is rooted at the binary that produced the refusal)

---

## ✅ Contributor Checklist

- [x] PR stays within 400 changed lines, or I have requested/obtained maintainer-applied `size:exception` with rationale documented (181 changed lines)
- [ ] I have added the appropriate `type:*` label to this PR — external contributors cannot apply labels; requesting `type:bug` from a maintainer
- [x] Unit tests pass (`go test ./...`) — focused local suites green, full suite via CI
- [x] E2E tests pass (`cd e2e && ./docker-test.sh`)
- [x] Benchmark validation completed, or this change is not applicable to the benchmark (explain why in the Test Plan).
- [x] I have updated documentation if necessary
- [x] I selected exactly one AI-assistance option and, if material assistance was used, completed all applicable declaration fields

---

## Pending repository workflow actions

Actions that require maintainer authority, recorded per the current API state:

- [ ] `type:bug` label — external contributors cannot apply PR labels
- [ ] Fork workflow approval — CI and PR Validation runs sit in `action_required` awaiting a maintainer (runs 34488929384 / 34488929531)
- [ ] `size:exception` — the PR is 486 changed lines against the 400 budget. Rationale: the delta over budget is review-driven growth, not scope creep — the fix itself is ~180 lines; the rest is the regression depth five CodeRabbit rounds and two reproduced merge-blocking compatibility gaps (review feedback) demanded: real-shell execution of the printed continuation, subprocess stand-in dispatch, quoting round-trips, and the bench j93 anchored-token validation. Splitting the regression out of the fix would leave the issue's suggested regression (`TestManagedAssetsContinuationUsesInvokingExecutable`) unmerged with the fix it proves.

---

## 💬 Notes for Reviewers

Could a maintainer add the `type:bug` label so the label check can pass? (External contributors cannot apply labels.)

For production Go changes in `internal/cli`, `internal/reviewtransaction`, or `internal/sddstatus`:

- [x] Identify any qualifying security, integrity, admission, repair, or governance guard and challenge its legitimate input population against real-world evidence. — The one touched guard is the `managed_assets_continuation` shape check inside failure `Validate()` (a `refusal:by-design world-action` declaration). Its population is exactly the two refusal producers (`reviewManagedAssetsStopTransition` for STATUS, `authorizeReviewStart` for START); both were re-exercised end-to-end by the regression test. No other guard's input population changed.
- [x] Confirm its `guard:population` direction and claim are adjacent and accurate, and that `.guard-population-baseline.txt` changed only when the guard contract intentionally changed. — The declaration count is unchanged and `GuardPopulation` tests pass; `.guard-population-baseline.txt` is untouched because the guard's contract (exactly one executable-anchored sync command) did not change in kind, only in admitted executable form.
- [x] Do not treat a passing declaration/registry check as proof that no qualifying guard was omitted or that the population claim is semantically complete. — The schema-exercising test (`TestManagedAssetsContinuationSchemaRuleIsExercised`) additionally validates real produced envelopes against the published schemas in both directions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Recovery commands for stale managed assets now use the executable that reported the issue, avoiding conflicts with other installed versions.
  - Executable paths containing spaces are handled correctly across supported shells, including Windows Command Prompt and PowerShell.
  - Continuation commands now receive stricter validation, including executable identity, optional agent settings, multiline input, and malformed status transitions.

- **Documentation**
  - Updated recovery command guidance to reflect executable-specific synchronization commands and shell-specific quoting requirements.


FranklinF25:
Thanks @coderabbitai — both findings read, verified against the code, and addressed in b4eced86. One fixed, one kept with reasoning.

## Finding 1 (shell-dependent quoting) — fixed

The continuation now uses one exact platform-specific encoding, chosen by the invoking binary's `runtime.GOOS`:

- POSIX platforms quote with **single quotes**, the only form no POSIX shell expands (`$` and backticks expand even inside double quotes). An embedded single quote uses the standard close/escape/reopen idiom (`'\''`).
- Windows keeps **double quotes**: cmd.exe does not treat single quotes as command syntax, and both cmd.exe and PowerShell accept double quotes.
- A path over the conservative safe class `[A-Za-z0-9/._+=@:,-]+` stays bare. An allowlist, not a blacklist: safety never depends on enumerating every shell's metacharacter set.

I took the second option you offered ("implement and declare one exact platform-specific shell encoding") rather than the structured `executable` + `arguments` representation: publishing a new contract shape changes what `gentle-pi` relays verbatim today, and that coordination belongs to maintainers as a deliberate follow-up, not inside a bugfix.

I also made the published schema itself enforce the safe form: the `managed_assets_continuation` `command` pattern now shares the same safe class as its bare-token grammar, so `/opt/$HOME/gentle-ai sync` unquoted is schema-invalid rather than merely renderer-avoided. `Validate()` enforces the identical regex in-process, and the regression table covers every quoting form plus that negative case (`/opt/o'brien/gentle-ai` renders as `'/opt/o'\''brien/gentle-ai'`).

## Finding 2 (bare fallback recreates PATH ambiguity) — kept, deliberately

1. The published contract mandates a continuation for `managed_assets_outdated`: failure.schema.json's if/then requires it in both directions, and `Validate()` refuses its absence. Emitting a non-runnable recovery is a coordinated schema + validator + producer change.
2. The alternative resolution ladder (`exec.LookPath(os.Args[0])`) would resolve through PATH, which can return the *other* binary. A false identity claim is strictly worse than an honest unqualified name: the failure mode you describe requires `os.Executable()` to fail, which does not happen on any supported platform (Linux `/proc/self/exe`, Windows `GetModuleFileName`, macOS `_NSGetExecutablePath`). The fallback is a documented degradation to pre-fix behavior in a state normal execution cannot reach.
3. If maintainers prefer a verified-identity or non-runnable-recovery contract, I'm glad to do that as a follow-up. It touches the schema, `Validate()`, and every consumer, so it shouldn't ride silently inside this bugfix.

## Verification after the change

Focused managed-assets suite, the full `Review`-filtered `internal/cli` suite, `go run ./internal/gofmtcheck`, `go build ./...`, and the Docker E2E matrix (ubuntu, arch, fedora: 3/3) all pass locally on b4eced86.

FranklinF25:
Follow-up finding (integration regression for convergence) addressed in d3fee668.

The end-to-end section of `TestManagedAssetsContinuationUsesInvokingExecutable` now executes the advertised continuation — the sync this very binary performs for `gentle-ai sync --agent opencode`, via `RunSyncWithSelection` against the same home — and asserts that the next STATUS from the same binary returns `execute` / `fresh_target_ready` instead of the refusal. That is the exact convergence invariant the issue names, previously covered only by the `recordManagedAssetDigest` shortcut in the older STATUS test.

Verified locally: managed-assets cluster, the full `Review`-filtered `internal/cli` suite, `gofmtcheck`, and `go build ./...` all green on d3fee668 (production code unchanged since the last full Docker E2E run on b4eced86).

FranklinF25:
Round 3 finding (execute the emitted continuation) addressed in bd46c3f4.

The convergence step no longer calls `RunSyncWithSelection` in-process. The regression now executes the emitted command line end to end:

- The rendered command is split into argv with an independent splitter that honors both quoting forms the renderer may emit (POSIX single quotes, Windows double quotes) by unquoting, not by shelling out.
- The executable token must equal the invoking (test) binary, so a continuation aimed at any other executable fails the regression before it runs.
- The process is re-executed as a subprocess under a `TestMain` stand-in guard that routes the CLI arguments through the same verb dispatch and flag parsing `app.RunArgs` performs for `sync` (the exact `cli.RunSync(args[1:])` hop; `internal/app` cannot be imported from this package's tests without a cycle, so the guard replicates that one routing line and rejects anything else).
- The captured test-home environment is inherited, so the sync records this binary's own digest, and the next STATUS from the same binary must return `execute` / `fresh_target_ready`.

This covers the full chain the finding names: executable startup, CLI dispatch, argument parsing, and convergence. A broken executable-anchored continuation can no longer pass: a bare `gentle-ai` token fails the argv identity check, and a wrong-binary token would record the wrong digest and leave STATUS refused.

Verified locally on bd46c3f4: managed-assets cluster, the full `Review`-filtered `internal/cli` suite, `gofmtcheck`, `go vet ./internal/cli/`, and `go build ./...`. Production code is unchanged since the full Docker E2E run (ubuntu, arch, fedora: 3/3) on b4eced86.

FranklinF25:
Splitter finding addressed in 00e306a2 — verified valid before fixing.

The splitter indeed mishandled the splice: for `/opt/o'brien/gentle-ai` the renderer emits `'/opt/o'\''brien/gentle-ai'`, and the old decoder wrote the splice's backslash literally and lost the apostrophe, producing argv[0] = `/opt/o\brien/gentle-ai`.

The splitter now decodes both escape forms:

- Outside any quote, a backslash escapes the next character — exactly the splice the renderer emits between single-quoted spans for an embedded apostrophe.
- Inside double quotes, a backslash escapes only a double quote or another backslash (the Windows form the renderer emits); anything else stays literal, matching POSIX double-quote semantics.

Per the finding's request, a round-trip table now asserts renderer-then-splitter identity for every quoting form — safe bare path, `$`-expansion path, embedded-apostrophe path, and Windows path with spaces — so the argv for the apostrophe fixture provably contains the original executable path. Verified locally: the regression (now 9 subtests), the managed-assets cluster, the full `Review`-filtered `internal/cli` suite, `gofmtcheck`, `go vet`, and `go build ./...` all green.

FranklinF25:
UNC backslash finding addressed in 9a0391f6 — verified valid before fixing.

The v2 splitter decoded `\\` inside double quotes, so a UNC lead (`\\server\share\gentle-ai.exe`) collapsed to a single separator and argv[0] stopped naming the executable.

The splitter now decodes only backslash-quote sequences in both contexts:

- Outside any quote, a backslash escapes the next quote character only — the POSIX splice for an embedded apostrophe. A backslash before anything else stays literal.
- Inside double quotes, a backslash escapes only a double quote — the Windows form the renderer emits. Path separators and doubled UNC leads survive intact.

The round-trip table grows accordingly: a POSIX path containing backslashes, a quoted UNC path with spaces, and a quoted UNC path without spaces, each asserting renderer-then-splitter argv[0] identity. The regression now runs 13 subtests; managed-assets cluster, the full `Review`-filtered `internal/cli` suite, `gofmtcheck`, `go vet`, and `go build ./...` all green locally.

microonline:
Thanks for taking this. I reproduced two merge-blocking compatibility gaps on the current PR head (`9a0391f`):

1. The new Windows regression test fails as committed:
   ```text
   go test ./internal/cli -run '^TestManagedAssetsContinuationUsesInvokingExecutable$' -count=1
   STATUS continuation command = "\"C:\\...\\cli.test.exe\" sync ..."
   want "C:\\...\\cli.test.exe sync ..."
   ```
   Production quotes every ordinary Windows path because backslashes fall outside the bare allowlist, while the independent expectation only quotes paths containing whitespace or shell metacharacters. More importantly, PowerShell's parser rejects `"C:\Program Files\Gentle AI\gentle-ai.exe" sync --agent pi`; it requires the call operator, `& "..." sync ...`. The `exec.Command` test bypasses shell parsing, so it cannot prove the documented "literally runnable" command on PowerShell.

2. The existing driven journey also fails:
   ```text
   cd bench && go run . run --binary <pr-binary> \
     --only j93-stale-managed-assets-start-is-not-unknown
   ```
   `j93` rejects the absolute executable because `staleManagedAssetsStatusIsNotUnknown` still requires the `gentle-ai sync` prefix, and `printedCommandArguments` requires the first token to equal `gentle-ai`.

Could you make the Windows continuation runnable in the operator's shell and update `j93` to execute and validate the anchored executable token? That would preserve the fix's convergence claim without bypassing the two real consumer boundaries.

FranklinF25:
Both reproduced gaps are closed in a1e11cca (plus 4e47e444 cleaning up after myself — details below). Thank you for reproducing them against the head instead of eyeballing the diff; both failure modes were real.

## Gap 1: Windows expectation divergence + PowerShell claim

The independent expectation quoted only whitespace/metacharacter paths while production quotes every path outside the safe bare class, so backslash-heavy Windows paths diverged exactly as you showed. The helper now mirrors the allowlist semantics with its own byte scan (`bare` iff every byte is in the safe class), so `C:\...\cli.test.exe` is expected quoted on Windows and bare stays bare.

On PowerShell: you are right, and the claim was wrong. There is no quoting form that is paste-and-run in both cmd.exe and PowerShell (the call operator `&` that PowerShell needs is a command separator in cmd.exe), so the honest fix was to narrow the contract rather than fake a universal one:

- The renderer comment now states the quoted Windows form is the cmd.exe contract, with the PowerShell call-operator caveat spelled out; bare tokens are unaffected in every shell.
- The docs row carries the same caveat.
- The regression no longer *decodes* the printed command and calls it proven: on POSIX it runs the exact printed line through a real `sh -c`, so paste-and-run is proven where a shell can prove it. On Windows the argv dispatch still exercises startup, CLI dispatch, and flag parsing, with the boundary documented in the test — that is the honest limit of what an in-process test can claim there.

## Gap 2: driven j93

j93's prefix assertion and `printedCommandArguments`' first-token rule are replaced by `anchoredContinuationArguments`: the printed continuation is split with the same splitter, its verb must be `sync`, and its first token must equal the driven binary itself (symlink-tolerant via `filepath.EvalSymlinks`, since the runner resolves `--binary` absolutely). The sync arguments then execute through that same driven binary, which turns the anchoring into a driven-corpus proof rather than a rendered-string one.

Verified exactly as you ran it: `go run . run --binary <pr-binary> --only j93-stale-managed-assets-start-is-not-unknown` → `completed`, 3 commands (STATUS stop → anchored sync → STATUS `fresh_target_ready`), 0 blocks. Also green: `go test ./...` in the bench module, the managed-assets cluster, the full `Review`-filtered `internal/cli` suite, `gofmtcheck`, and `go build ./...`.

## One accident, disclosed

My j93 verification wrote its single-journey output over the tracked full-corpus `bench/results.json` (that is the 12k-line delete in a1e11cca). 4e47e444 restores the 64-journey corpus byte-identical to base — the net diff on that file is zero.

The PR now sits at 486 changed lines against the 400 budget (review-driven growth, mostly regression depth); I have requested `size:exception` with that rationale in the Pending repository workflow actions section for a maintainer to weigh.

dnlrsls:
Good direction. I independently reproduced #4434 and exercised the same behavior through STATUS and driven j93. Three boundary cases are worth tightening before merge:

- Reject `\r` and `\n` in any executable identity before quoting. The current quoted grammar can otherwise admit a multiline command.
- The unresolved `os.Executable()` fallback concern is real. A bounded fallback can accept `os.Args[0]` only when it is absolute and line-break-free. If no verified identity remains, the continuation should not describe bare `gentle-ai` as exact.
- Use the same strict continuation validator for both FAILURE and STATUS so their accepted command shapes cannot drift.

I verified these cases locally with Windows spaced paths, POSIX quoting, malformed-command rejection, and the j93 continuation execution. Happy to share the focused cases if useful.

dnlrsls:
@coderabbitai resume

dnlrsls:
Addressed the multiline-parser concern in `31829191`: `anchoredContinuationArguments` now rejects CR/LF before tokenization, with focused LF and CR regression cases.

I am keeping the driven benchmark strict about the executable token rather than accepting bare `gentle-ai`. Journey `j93` specifically guards the #4434 invariant that the emitted continuation names the diagnosing binary; accepting a bare PATH-resolved token there would allow the original regression to pass. The last-resort bare fallback remains intentional in production and is already covered by injected unit cases where neither `os.Executable()` nor an absolute `argv[0]` can provide an identity.

microonline:
Rechecked `31829191` on Windows. Both reported gaps are resolved: the focused continuation regression passes, and driven `j93` completes with 1 completed / 0 failed, proving stale-assets refusal → anchored sync → `fresh_target_ready`. `gofmtcheck` also passes. Thanks for addressing both findings.

dnlrsls:
@coderabbitai resume

dnlrsls:
@coderabbitai full review

decode2:
This PR merged while our independent review of head 6c151c93 was completing. Recording the outstanding findings for post-merge follow-up rather than submitting a changes-requested review on a closed PR:

- The Windows display command quotes the executable but omits the PowerShell invocation operator. cmd.exe also expands percent-delimited environment variables inside quoted paths. Tests using exec.Command bypass these shell semantics, so they do not demonstrate that the emitted command is literally runnable in those shells.
- managedAssetsExecutablePath still falls back to bare gentle-ai when executable identity cannot be obtained. That can reintroduce the wrong-PATH-binary recovery loop from #4434; documentation must not claim unconditional executable identity.
- Benchmark command parsing applies POSIX double-quote escape rules to Windows UNC paths, so it can reject an otherwise valid anchored path. Please cover UNC and actual shell interpretation, not only direct process invocation.

The same-binary recovery direction remains correct. Focused CLI and benchmark tests passed with the patch on audited main 22159d20, but no native Windows shell execution or driven j93 was performed. Structured executable/arguments may be an option, not an approved contract migration: evaluate compatibility before changing the published schema. These are follow-up findings, not a request to revert the product direction.

Issue #1689, opened 2026-07-22
fix(opencode): enrich reviewer output with provider-owned capture subject


dnlrsls:
## Pre-flight Checklist
- [x] I have searched existing issues and this is not a duplicate.
- [x] I understand PRs require `status:approved`.

## Bug Description
A negotiated bounded review exposes provider-owned `artifact_subject`, exact candidate diff, and changed-path manifest through START/status. The selected OpenCode reviewer correctly returns the strict native model-owned shape `{findings,evidence}`. The managed `review-result-artifacts.ts` hook forwards that JSON unchanged to `review capture-result`, but native admission requires the capture-only envelope to also contain provider-derived top-level `subject_hash` and `inspection`. Capture therefore fails with `reviewer artifact admission incomplete: reviewer result omitted the provider-owned artifact subject`, preserves the raw result, and leaves authority in `reviewing` with no admitted lens result.

This is a product/integration regression or incomplete acceptance from merged PR #1550, which introduced provider-owned artifact admission and claimed managed OpenCode support. It is distinct from #1610 (authority revision propagation), #1411 (capture incident preservation/recovery), and #1454 (immutable diff/manifest delivery). The reviewer must not emit or construct provider-owned metadata; the adapter owns enrichment.

## Steps to Reproduce
1. Install/sync a Gentle AI beta containing provider-owned reviewer artifact admission and use OpenCode.
2. Start or resume a negotiated review whose native next transition is `collect/reviewer_results_required`; verify it provides `artifact_subject`, `candidate_diff`, and `changed_path_manifest`.
3. Launch the selected bound reviewer through the managed hook.
4. Let the reviewer return valid strict JSON with only `findings` and `evidence`, as required by its agent contract.
5. Observe the after-hook extract that JSON and forward it unchanged to `review capture-result`.
6. Observe capture fail with the admission-incomplete error and preserve the raw reviewer result while authority remains unchanged.

## Expected Behavior
The managed adapter should build a capture-only envelope by combining reviewer-owned `findings`/`evidence` with provider-owned `subject_hash` from `artifact_subject.subject_hash` and `inspection: {status: "completed", paths: [...]}` derived from the exact ordered changed-path manifest. Models must continue returning only `findings`/`evidence`. The same deterministic enrichment path should support replaying a preserved raw result through full native verification without rerunning the lens or opening another review budget.

## Actual Behavior
[...]

PR #4755, opened 2026-09-18
fix(review): bound reviewer context per runtime before START creates authority

### LINKED ISSUE #4680: bug(review): START accepts a candidate whose ~4 MiB lens context the Claude reviewer rejects as "Prompt is too long"; every capture fails deterministically (2.9.1)
## Automated provider-defect report

Observed evidence only; no root cause is asserted.

### Environment

- Gentle AI: `2.9.1` (stable channel)
- Client: Claude Code, in-process capture (`review capture-result ... --agent=claude-code`, no `--input`)
- OS/architecture: Windows x64
- Review contract: `gentle-ai.review-integration/v2`

### Operation shape

- Negotiated workspace review, `--committed-only=true`, explicit base ref (the merge base with the integration branch).
- Candidate: 84 changed paths, ~76,900 changed lines, risk `high`, canonical four-lens plan, 200-line correction budget.
- Almost all of the size is plain-text documentation/JSON planning files (the largest single path is ~5,800 lines); about 13 paths / ~870 lines are code.
- One consented START (`granted`), which created the lineage and returned `collect / reviewer_results_required` with four `review.capture-result` slots.

### Bounded attempts and outcomes

1. Four exact capture invocations, launched concurrently: all four exited 1 with `gentle-ai.review-integration.failure/v2`, `phase: preflight`, `code: invalid_request`, `mutation_outcome: not_started`, `retry_safe: true`, cause `claude reviewer transport failed: exit status 1: Prompt is too long`.
2. Bound STATUS re-offered the same four slots (unchanged transition).
3. One relaunch of the four exact invocations: identical failure on all four (deterministic).
4. Read-only `gentle-ai review lens-context` for one lens: the lens context is **4,130,976 bytes**, i.e. just under the 4 MiB byte budget, so no `lens_context_budget_exceeded` stop fires at START or STATUS.

### Expected behavior

A candidate that START accepts should be one whose lens contexts the selected reviewer transport can actually accept, or START/STATUS should refuse it up front with `lens_context_budget_exceeded` (or an equivalent typed stop naming a resolvable action). A lens context of ~4 MiB of text is far beyond what the Claude reviewer prompt accepts.

### Actual behavior

START persists authority; every capture fails deterministically with a `retry_safe: true` failure, and STATUS keeps re-offering the same slots, so the lineage stays in `reviewing` with no exit other than declaring slots unachievable or disabling review.

### Minimal reproduction

1. Commit a change containing a few thousand lines of code plus tens of thousands of lines of plain-text files, so the per-lens context lands between the reviewer prompt limit and 4 MiB.
2. Run the selectorless STATUS, then the returned START with `--consent granted`.
3. Run the four returned `review.capture-result` invocations with `--agent=claude-code`.

### Safe identifiers

- target: `sha256:59f98fc175508a24…906b3620`
- authority revision: `sha256:d9b3153048e7bf88…18507ea70`
- lineage: `review-7e90126edb25b54e`

### Preserved state

The lineage remains open in `reviewing`; no correction, receipt, or acknowledgement was produced. Related but distinct: #3389 (a single path over 4 MiB) and #4275 (reviewer model selection).

barbatdev:
## 🔗 Linked Issue

Closes #4680

Follow-ups opened from this work, both listed under Known limits below:

- #4777 — `capture-result --input` admits a result without consulting the budget guard
- #4778 — `invalidate` rebuilds current-snapshot evidence, so a drifted pristine lineage cannot be invalidated

---

## 🏷️ PR Type

- [x] `type:bug`: Bug fix (non-breaking change that fixes an issue)

---

## 📝 Summary

START was admitting candidates that no reviewer could actually hold. It checked the candidate against the 4 MiB Git ceiling, but a runtime holds far less than that, so a big candidate froze authority and then failed every single capture with `Prompt is too long`. By that point authority existed, so the lineage had no way out.

START now proves, before it persists anything, that the candidate's frozen evidence fits a 200 KiB per-runtime budget for every role, not just the lens. Generated paths travel as metadata summaries instead of content hunks.

That proof is an admission bound, not a lifecycle bound, and the difference matters. It measures what is frozen and knowable at START. It cannot measure what a reviewer produces later: findings are unbounded free text, and corrections are capped in lines but not in bytes. Bounding a worst case there would refuse candidates that fit, which is the same reasoning the code already records for refuter claims.

So the correction stage is closed the other way. An adversarial pass showed the targeted validator can exceed the budget once corrected evidence and findings exist, after lens results already made the lineage non-pristine, which left authority that could neither complete nor be invalidated. That refusal is now classified as a typed stop, `correction_context_budget_exceeded`, and it names the exit that already existed: `review abandon` accepts every non-terminal state, including `correction_required` and `validating`. Only the deterministic typed refusal stops a lineage, so an inconclusive probe never strands a healthy candidate.

---

## 📂 Changes

Measured against the merge base (`15ea98ed`) at the current tip: **44 files, 3,158 changed lines** — 1,894 test, 959 production, 295 the ODD feature document, 10 docs/assets/goldens.

| File / Area | Lines | What Changed |
|-------------|-------|-------------|
| `cli/review_lens_context.go` | 235 | Runtime-scoped budget, generated paths become immutable metadata summaries, START probe also proves the non-lens role envelopes fit |
| `cli/review_correction_context.go` | 150 | New. Read-only tri-state probe that assembles the real validator request in the correction branch of STATUS and emits the typed `correction_context_budget_exceeded` stop |
| `cli/review_provider_roles.go` | 126 | Role evidence matches what a lens gets, serialized role prompts bounded by the same cap, new `reviewProviderRoleEnvelopeFloor` |
| `reviewtransaction/snapshot.go` | 85 | Stamp and validate the generated-path interpretation discriminator |
| `reviewtransaction/frozen_candidate_context.go` | 46 | Carry the blob identities and numstat a summary needs |
| `reviewerprovider/capture_runtimes.go` | 28 | Per-runtime context budget declaration |
| `reviewerprovider/contract.go` | 24 | Validator briefing stops promising content it never receives (mostly briefing text) |
| `cli/review_next_transition.go`, `review_narration.go`, `review_facade.go` | 40 | Route and narrate the new stop, keep it inside Pi's facade-only contract |
| `cli/review_provider_correction.go` + 6 more | 26 | Corrective retries bounded by the same cap, persisted interpretation propagated instead of re-derived |
| `cli/review_operation_contract.go`, `review_status_contract.go` | 109 | Post-review: `ReviewManagedAssetsContinuation` becomes `ReviewStopContinuation` now that two stops carry one, each validated by its own validator |
| `odd/tasks/review-runtime-context-budget.md` | 295 | ODD feature document, not production code |

---

## ⚠️ Known limits

These are real, NOT closed by this PR, and now tracked:

- **#4777** — `review capture-result --input` never consults the budget guard (its only production call site is STATUS), so a hand-built result still admits on an over-budget candidate and makes the lineage non-pristine.
- **#4778** — `review invalidate` rebuilds current-snapshot evidence, so a pristine lineage whose worktree drifted cannot be invalidated.

Both matter specifically for `review recover`, which mints successor authority from a new snapshot and never reaches the START guard. An over-budget recovered lineage is still creatable on this build. That is not the dead end this PR closes: the recovered lineage arrives with no admitted role results, stays pristine, classifies as the same typed stop, and arrives with `review invalidate` available as its exit. It is an arrival guarantee, not an invariant — #4777 and #4778 are exactly the two paths that can take the exit away afterwards, and the doc comment and the test name (`TestRecoveredOverBudgetLineageStopsTypedAndArrivesWithItsExitIntact`) both state it that way on purpose.

Where this PR states a guarantee, it states the conditions too.

---

## 🤖 AI Assistance

- [x] **Material assistance used**

**Tool/model (if known):** Claude Opus 5, via Claude Code.

**Material scope:** implementation, tests and commit messages across all seventeen work-unit commits.

**Verification performed:** I ran every test cited below myself, and observed RED before each fix and GREEN after. Being blunt about the cost: every commit is AI-assisted, and several repair defects that adversarial review found in earlier commits of this same branch — some of those in commits that were themselves fixes. The assistance also produced one finding that was flat out wrong, and I implemented it before checking the premise and had to revert it. Two review runs on the same model reached opposite conclusions on one question and I settled it by reading the code, not by picking a winner.

---

## 🧪 Test Plan

Local runs, with `env -u OPENCODE_CONFIG_DIR` and a canonical `TMPDIR` so an inherited config stays out of it:

- [x] `go test ./internal/reviewtransaction` PASS
- [x] `go test ./internal/reviewerprovider` PASS
- [x] `go test ./internal/cli` — one failure, `TestEngramPathGuidanceDefault`, which is environment-dependent, reproduces on untouched `main`, and passes in CI
- [x] Go format passes (`go run ./internal/gofmtcheck`), `go vet` clean
- [x] Every behavioral claim below was reproduced by execution before its fix
- [x] After merging `main` (`dda02c89`) I re-ran components, sdd, providercontractbundle, assets and the cli battery (budget, correction, transitions, stops) — green

**On the current tip, the authoritative evidence is CI**, which is fully green over the merge commit: Unit Tests, E2E on ubuntu/fedora/arch, Darwin and Windows runtimes, Organic Runtime E2E on both, OpenCode V2 released SDK contracts, Go Format and every repo policy check. The local timings previously quoted here were from an earlier tip and have been dropped rather than restated for commits they were not run on.

**Benchmark validation: not run.** This touches the review lifecycle so the guide applies on its face. My read is that it does not move measured friction, since it adds a pre-authority refusal and changes how evidence is represented, without adding a journey, gate, recovery or delivery path. That is a read, not a measurement. If you want a before/after run, say so and I will do it before this merges.

**Known and not caused by this change:**
- `TestRuntimeCatalogDiscoveryIgnoresStaleProjectResults` (`internal/tui/model_test.go`) fails on `main` with nothing applied. No commit here touches `internal/tui`.
- An intermittent `review store lock could not be acquired: not a directory` (ENOTDIR) in store tests, 2 of 5 full runs. Pre-existing, unexplained, deliberately left out of this PR.

---

## 💬 Notes for Reviewers

Read it commit by commit. Seventeen work units plus one `main` merge; each carries its reasoning in the message.

Several commits repair defects introduced by earlier commits on this branch, two of them fixing a commit that was itself a fix:

- `keep legacy authorities valid against live evidence`: `snapshotsEqual` compares the interpretation discriminator, but `ValidateLiveSnapshot` and `rebuildCurrentSnapshotEvidence` never carried the stored value across the way two other call sites already did. A legacy authority got rejected against an unchanged repository, reporting identical expected and observed identities.
- `summarize generated paths for refuter and validator too`: START admits by proving the lens block assembles, which only works while every role gets the same evidence. Summarizing in the lens loop alone broke that and moved the unexecutable lineage to after authority creation.
- `stop promising the validator content it is not handed`, introduced by the commit above. The briefing still said its evidence carries "the complete frozen tree-to-tree patch for every path" and is "not a summary of them, so a verdict reached from it is a verified verdict". For a correction touching a lockfile or a golden that is simply false, and the role would sign a verified verdict over bytes it never saw.
- `measure the role envelope START admits a candidate under`: the lens block carries patch bytes raw, while a refuter or validator prompt carries them through `json.Marshal`, which doubles every quote, backslash, newline and tab. A quote-dense candidate passed the raw probe and dead-ended at role materialization with authority already frozen.
- `charge the frozen policy in the role envelope floor`, introduced by the commit above. A large `--policy` froze authority on evidence that fits and then failed every validator capture.

The last three commits are the adversarial pass on `recover` and the correction stage: pin the recover behavior by execution instead of prose, correct the guarantee to the arrival it actually proves (opening #4777 and #4778), then classify the correction-stage refusal and name its exit.

### Post-review corrections

Two commits sit on top of the `main` merge, both from CodeRabbit's pass on the reviewed tip. Neither changes the shape of the change or the admission/lifecycle distinction above.

- `070f82ed` — **a corrected over-budget lineage could not be released on Pi.** The `correction_context_budget_exceeded` row in Pi's shipped ledger told the maintainer to run "the release command the stop's `continuation` names", but `reviewStopTransition` sets only Kind and ReasonCode, so that field was nil and omitted; only the managed-assets stop populated one. Pi's contract never mentions `abandon`, and `validPiFacadeLifecycle` forbids the raw route, so Pi was told to release authority with no way to do it. I introduced this in `93c4c607` while making the row satisfy the facade-only contract. The stop now carries the continuation, following the managed-assets precedent, and it carries the flagless `review abandon --cwd <repo>` on purpose: the full release needs an actor and a maintainer binding no producer may invent, so a command with placeholders would not run as printed. When eligibility says the authority is not releasable it emits no continuation rather than a command that would be refused. The test executes the named command and asserts it refuses with the binding template.
  Worth stating: CodeRabbit's proposed patch here was wrong. Writing `gentle-ai review abandon` into Pi's ledger fails `validPiFacadeLifecycle` and reintroduces the build failure `93c4c607` closed.

- `55a1a072` — **the block terminator was written outside the budget.** `reviewLensContextBlock` charged the headers and every section but wrote the terminator after the accounting: a fixed 29-byte overshoot of the 200 KiB runtime cap. Not a dead end, and worth being precise about why — the START admission probe and the materialization run through this same function, so they always agreed and nothing was admitted that later failed. The defect was that the comment promised to bound the whole delivered block and the arithmetic did not. The terminator is now reserved before the first section, its length derived from the constant.

A third CodeRabbit finding, `capture-result --input` admitting without the guard, is #4777 above: already a declared known limit, not fixed here.

### Where I would push back

Two spots I would press if I were reviewing this:

1. Carrying a persisted discriminator forward could in principle mask a real repository change. I believe it cannot, because `GeneratedPathInterpretation` is a constant stamped on every rebuild and never derived from candidate content, so it carries no information about the repo. If you read it differently, that is the line to press.
2. The envelope floor still charges refuter claims zero. That omission is on purpose, since claims do not exist until a lens has run and guessing them would refuse candidates that fit. But it is the same shape as the policy defect above, and the only thing keeping it from being the same bug is that claims genuinely are not knowable at admission. Worth a second opinion.

There is also an audit note in the comments covering every path that materializes runtime input after START, with file and line.

### Size

3,158 changed lines, over the 400 budget, asking for `size:exception`. 959 are production, 1,894 are behavioral tests kept with the units they verify, and 295 are the ODD feature document. I looked at slicing this with `/chained-pr` and dropped it: fork-first with no cross-fork base means slices land one at a time, so the first would merge before anyone read the last.

### Guard population

`.guard-population-baseline.txt` unchanged, no guard contract changed here.

## Summary by CodeRabbit

* **New Features**
  * Added runtime-aware limits for review context, preventing oversized review requests from being partially processed.
  * Generated files now appear as metadata summaries, while authored files continue to include full patches.
  * Added clearer recovery guidance when correction evidence exceeds the context budget, including an executable review-abandon continuation when eligible.

* **Bug Fixes**
  * Preserved historical review snapshots and generated-file handling across validation, recovery, and resnapshot workflows.
  * Improved reviewer instructions for inspecting omitted generated-file content and reporting unavailable checks.


barbatdev:
## Audit note: every path that materializes runtime input after START

Ran this on the branch to check whether any path other than the lens can get around the runtime budget, or bring the dead end back under a different role. Putting it here because the file level detail does not fit in the body, and otherwise a reviewer has to rebuild it by hand.

**Short version:** the bound holds on all four paths now. It did not hold before this audit, and the defect it found was mine.

| Path | Bound |
|---|---|
| lens | `reviewLensContextBlock`, `review_lens_context.go:603` |
| refuter | `reviewProviderMaterializeEvidence` + `reviewProviderRolePrompt`, `review_provider_roles.go:346, 411` |
| validator | same two |
| corrective retry | `review_provider_correction.go:167` |

There is no fifth path. Every site that builds an `Invocation` (`review_provider_roles.go:170`, `:317`, `review_provider_correction.go:170`, `review_provider.go:76`) takes bytes from one of those constructors, and nothing else in the tree calls `NewInvocation`. `review lens-context`, `capture-result --materialize`, `capture-refuter`, `capture-validation` and both host relays all funnel into them. The OpenCode relay only prepends a JSON header to bytes that already came out of `Invocation.Prompt()`.

The runtime identity is not a way around it either. `RuntimeContextBudget` (`capture_runtimes.go:26`) returns `ApprovedRuntimeContextBudget` on both branches, so an unknown or empty identity gets the same cap and never a bigger one. All three `reviewProviderCaptureRetry` callers pass `state.RuntimeAgent`, the identity START froze.

### The defect it found, which was mine

The envelope floor charged the frozen policy zero, while the real validator prompt carries it twice: appended raw to the instruction, then again JSON escaped inside the marshalled request. The policy gets read and frozen at START and its file has no size bound of its own. So a large `--policy` froze authority on candidate evidence that fits, then failed every validator capture deterministically, with a lens result already persisted so `compactPristineReviewing` was false and `review invalidate` was gone.

Same dead end this issue is about, through the validator, caused by something that was knowable at admission time. My own comment justified the omission by saying the policy does not exist until a lens has run, which is not true. Fixed in `bf969778`, and I rewrote the comment too, because a wrong rationale tells the next reader not to look.

### One finding I implemented and then reverted

The audit also flagged START treating an undecided probe as a candidate that fits. I changed it before checking the premise and it was wrong. `c6e6a1e4` (#1689) makes a candidate whose diff exceeds the Git ceiling startable on purpose, because START carries frozen tree references instead of an eager diff, so a change bigger than any inline limit is still addressable. My change took that away and broke two tests that encode the design. Reverted, and the code now has a comment citing #1689 so nobody closes that path the way I did.

### One surface worth naming

`gentle-ai review inspect-candidate`, which the validator is told it may run, is bounded by the 4 MiB Git ceiling (`frozen_candidate_context.go:418`) and not by the 200 KiB cap. It is the only place where more bytes than the cap can reach a runtime.

I do not think it breaks anything: it takes no part in START admission, freezes no authority, and an oversized response degrades into the typed `inspection.status: unavailable` check instead of dead ending. That is a read, not a proof, and the surface had not been named anywhere.

### One finding still open

`review recover` mints a REVIEWING authority with a new snapshot and new lenses without any budget check. The START guard has exactly one call site, `review_facade.go:2193`. Its `Start{...}` also omits `RuntimeAgent`, so every recovered lineage carries the empty identity. Harmless today because the budget fails closed, but `capture_runtimes.go:16-18` reserves the right to specialize per runtime. Separate surface, deserves its own issue, and given that one unverified finding already cost a revert here, somebody should confirm the premise before writing it up.

barbatdev:
This is the known `capture-result --input` bypass, already tracked as #4777 and documented under Known limits in the body.

The finding is valid and it stays out of this change on purpose. This PR closes the START admission path and the deterministic correction-stage stop. Raw result admission is a different mutation boundary, with its own contract and its own regression surface, and widening this PR to cover it would not make either one easier to review.

Worth confirming so the thread does not have to re-derive it: #4777 covers the exact behavior, not just the bypass. It states that a recovered over-budget lineage is acceptable today only because it arrives pristine with `review invalidate` available as its exit, and that an admitted `--input` result flips `compactPristineReviewing` and leaves the lineage with no non-destructive exit. The arrival guarantee losing its last exit is tracked, not implied.

No change here.

Alan-TheGentleman:
Thanks for this one — it closes the class #4680 describes, and it stays necessary after #4784.

I merged `main` into your branch myself (`f17c764a`, maintainer edits are enabled on the fork) so the two land cleanly together:

- #4784 made the pi runtime's refuter and targeted validator host-mediated and advanced the negotiated status schema to `status/v9`. Your branch had no dependency on the code that PR removed; the only conflict was the import block of `review_provider_role_materialize_test.go` (kept `context`, which your runtime-budget test needs; dropped `fmt`, which nothing uses any more). No production line of yours changed.
- `status-v9.schema.json` carries no `continuation` rule, so `correction_context_budget_exceeded` passes the published-schema conformance tests as-is.

Local run on the merge, with `TMPDIR` under `$HOME` (the `internal/reviewtransaction` suite refuses the macOS `/var/folders` default on untouched `main` too): `go test ./...` green, `go run ./internal/gofmtcheck` clean, `cd bench && go test ./...` green.

One note for the known-limits section: after #4784, `capture-refuter --input` and `capture-validation --input` are admission surfaces of the same shape as `capture-result --input`, so #4777 now covers three forms. I left that as a scope comment on #4777 rather than widening this PR.

Merging once CI is green on the merge commit.

